PR #5, opened 2020-01-29
fix program_is_loaded() so it will not close fd 0

chaudron:
program_is_loaded() would close file descriptor 0, this patch avoids
it. Closing FD 0 in my scenario was causing a "No such file or
directory" error when trying to bpf_object__load() an object
which has the .attach_prog_fd attribute set.

### DIFF
--- a/lib/util/util.c
+++ b/lib/util/util.c
@@ -73,7 +73,7 @@ static int __get_xdp_prog_info(int ifindex, struct bpf_prog_info *info,
 {
 	__u32 prog_id, info_len = sizeof(*info);
 	struct xdp_link_info xinfo = {};
-	int prog_fd, err = 0;
+	int prog_fd = -1, err = 0;
 
 	err = bpf_get_link_xdp_info(ifindex, &xinfo, sizeof(xinfo), 0);
 	if (err)
@@ -115,6 +115,9 @@ static int __get_xdp_prog_info(int ifindex, struct bpf_prog_info *info,
 	}
 
 out:
+	if (prog_fd >= 0)
+		close(prog_fd);
+
 	return err;
 }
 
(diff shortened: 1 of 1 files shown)


chaudron:
This causes me a day of experimenting, as the following code would fail, and took me a while (comparing to earlier experiments) it would fail. The bellow code would work (trimmed down) if the get_loaded_program() was commented out, else it would fail with "No such file or
directory"

```
	if (get_loaded_program(&cfg_dumpopt.iface, &mode, &info) != 0) {
		pr_warn("ERROR: Specified interface must have an XDP program loaded!\n");
		return false;
	}

	//open_opts.attach_prog_fd = bpf_prog_get_fd_by_id(info.id); */
	open_opts.attach_prog_fd = bpf_prog_get_fd_by_id(57);
	trace_obj = open_bpf_file("xdpdump_bpf.o", &open_opts);

	error = bpf_object__load(trace_obj);
	if (error) {
			libbpf_strerror(error, error_msg, sizeof(error_msg));
			pr_warn("ERROR: Can't load eBPF object: %s(%d)\n",
				error_msg, error);
		}
```

tohojo:
Wow, not sure *what* was going on there. I **hope** that particular bit of logic only ended up that way because I've been moving the code around so much...

Anyway, thanks for the fix, and sorry you had to spend a whole day debugging this :/

chaudron:
> Anyway, thanks for the fix, and sorry you had to spend a whole day debugging this :/

No problem was some good time spend going over the libbpf/kernel code :)

tohojo:

>> Anyway, thanks for the fix, and sorry you had to spend a whole day debugging this :/
>
> No problem was some good time spend going over the libbpf/kernel code
> :)

Ah, great! :)

PR #6, opened 2020-02-14
Allow create of BPF binary blob to be include as a .c file

chaudron:


tohojo:
Hmm, that's interesting. Why do this instead of loading from the file system like xdp-filter is doing?

chaudron:
> Hmm, that's interesting. Why do this instead of loading from the file system like xdp-filter is doing?

For a tool like xdpdump, it does not make sense to me to ship a separate .obj file. We could add an option to load an external one, but it should not require one by default.

tohojo:

>> Hmm, that's interesting. Why do this instead of loading from the file system like xdp-filter is doing?
>
> For a tool like xdpdump, it does not make sense to me to ship a
> separate .obj file. We could add an option to load an external one,
> but it should not require one by default.

Why? It's loading a pre-compiled BPF program; why is it better to embed
it inside the binary than load it from the file system?

chaudron:
> Why? It's loading a pre-compiled BPF program; why is it better to embed it inside the binary than load it from the file system?

Look at it from a distribution perspective, where are you going to put the bpf program, how is your app going to look for it if it's not in the expected place. For some tools, like tracing, you might not want users to change it.

tohojo:

>> Why? It's loading a pre-compiled BPF program; why is it better to
>> embed it inside the binary than load it from the file system?
>
> Look at it from a distribution perspective, where are you going to put
> the bpf program, how is your app going to look for it if it's not in
> the expected place.

Oh, believe me, I have thought about it:
https://bugzilla.redhat.com/show_bug.cgi?id=1781646 :)

Having the files visible in the file system makes it easier to discover
which programs are installed on the system. And if/when we need to sign
them as part of the package building process, that is also way simpler
if they are files in the file system. So I'd much rather you just used
the same logic for loading them (see open_bpf_file() in util.c) instead
of doing this embedding trick...

> For some tools, like tracing, you might not want users to change it.

If people go around changing things in /usr/lib they either know what
they are doing, or they deserve the breakage they get ;)

chaudron:
> Oh, believe me, I have thought about it: https://bugzilla.redhat.com/show_bug.cgi?id=1781646 
Thanks, missed this discussion :) 

I was already using open_bpf_file() before changing it to a blob, so will revert back :)

I'll also close this merge request :)

tohojo:
Great, thanks :)

PR #8, opened 2020-02-18
headers/linux: update bpf.h and add bpf_trace_helpers.h

chaudron:
Update bpf.h to the v5.6-rc2 revision. Also added the
bpf_trace_helpers.h include file as it helps defining trace BPF
functions.


tohojo:
Hmm, bpf_trace_helpers.h is not a Linux UAPI header. So it probably shouldn't go into headers/linux but rather headers/bpf? Then it can be include as  #include <bpf/bpf_trace_helpers.h> similar to the libbpf helpers?

Also, could you please split this into two commits, one updating bpf.h and one adding bpf_trace_helpers.h? :)

chaudron:
Done, split it into two commits and moved it

PR #9, opened 2020-03-06
fix general use of struct enum_val arrays

chaudron:
These array need to be NULL terminated or else segfaults will happen.


tohojo:
Oops, good catch, thanks! :)

PR #11, opened 2020-03-16
xdp-dump: add xdpdump to xdp-tools

chaudron:


xdp-dump/xdpdump.c
```diff
@@ -0,0 +1,628 @@
...
+	.snaplen = DEFAULT_SNAP_LEN,
+	.rx_capture = PROG_IN,
+};
+struct dumpopt cfg_dumpopt;
+
+struct enum_val func_direction_options[] = {
+	{"fentry", PROG_IN},
+	{"fexit", PROG_OUT},
+	{"fentryfexit", PROG_INOUT},
+	{"fexitfentry", PROG_INOUT},
```
tohojo: How about just making the direction an OPT_FLAGS entry? That way people can just specify fentry,fexit as separate options, and there'll be no need to duplicate them?

xdp-dump/xdpdump.c
```diff
@@ -0,0 +1,628 @@
...
+
+/*****************************************************************************
+ * get_epoch_to_uptime_delta()
+ *****************************************************************************/
+static uint64_t get_epoch_to_uptime_delta(void)
+{
+	/* This function will calculate the rough delta between uptime
+	 * seconds and the epoch time. This is not a precise delta, but it's
+	 * good enough to get a general offset. The delta between packets is
+	 * still based on the timestamps from the trace infrastructure.
```
tohojo: Why is it not precise? Just because it takes a while before this is read until the first packet arrives, or?

xdp-dump/xdpdump.c
```diff
@@ -0,0 +1,628 @@
...
+	clock_gettime(CLOCK_MONOTONIC, &ts);
+	uptime = ts.tv_sec * 1000000000ULL + ts.tv_nsec;
+
+	return epoch - uptime;
+}
+
+/*****************************************************************************
+ * capture_on_legacy_interface()
+ *****************************************************************************/
+static bool capture_on_legacy_interface(struct dumpopt *cfg)
```
tohojo: I can see why it may be useful to have tool not fail when there's no XDP program loaded, but I wonder if it's not better to just load one? :)

xdp-dump/xdpdump.c
```diff
@@ -0,0 +1,628 @@
...
+			pr_warn("ERROR: Perf buffer polling failed: %s(%d)",
+				strerror(err), err);
+			goto error_exit;
+		}
+	}
+
+	fprintf(stderr, "\n%"PRIu64" packets captured\n",
+		perf_ctx.captured_packets);
+	fprintf(stderr, "%"PRIu64" packets dropped by perf ring\n",
+		perf_ctx.missed_events / 2);
```
tohojo: Why /2?

xdp-dump/xdpdump.h
```diff
@@ -0,0 +1,78 @@
...
+
+#ifndef container_of
+#define container_of(ptr, type, member) ({                    \
+	const typeof( ((type *)0)->member ) *__mptr = (ptr);  \
+	(type *)( (char *)__mptr - offsetof(type,member) );})
+#endif
+
+#ifndef min
+#define min(x, y) ((x) < (y) ? (x) : (y))
+#endif
```
tohojo: Maybe we should move all these into utils.h?

xdp-dump/xdpdump.h
```diff
@@ -0,0 +1,78 @@
...
+
+struct pkt_trace_metadata {
+	__u32 ifindex;
+	__u32 rx_queue;
+	__u16 pkt_len;
+	__u16 cap_len;
+	__u16 flags;
+} __packed;
+
+#ifdef _LINUX_PERF_EVENT_H
```
tohojo: This is to exclude the definitions when included from the BPF file? Feels a bit like a hack; wouldn't it be better to just split the bits that are shared between userspace and BPF into a separate file?

chaudron: I would prefer to keep it all in one file. Is there a special define always set for BPF programs, like Linux does for some of it's include files?

xdp-dump/xdpdump.c
chaudron: Will change this to use OPT_FLAGS (note to self, update documentation :)

xdp-dump/xdpdump.c
chaudron: More the delay between the two system calls, I assume they happen at the same time.

xdp-dump/xdpdump.c
chaudron: Thought about this, but it might alter the behaviour of the system. So I decided to just fall back to normal operation and print a warning. The goal for this was that a user could use the same tool and work as expected.

xdp-dump/xdpdump.c
chaudron: Good catch, no idea why the /2 is still here, guess some leftover debugging :(

xdp-dump/xdpdump.h
chaudron: Is see min() is already there, but I needed it in the bpf program (not anymore it looks like) so that's why it was here. The others I'll move there.

xdp-dump/xdpdump.c
tohojo: Yeah, sure, that's what I meant. I was just wondering if there was some other reason why this would be inaccurate, or if it was just the delay between operations; could you clarify the comment so this is clear, please?

xdp-dump/xdpdump.c
tohojo: Yeah, fair point, loading an XDP program can have side effects... Hmm, I wonder if we should just bail then? My fear is that we'll end up re-implementing a half-baked tcpdump if people start using this feature, basically... :)

xdp-dump/xdpdump.h
tohojo: Right, cool

xdp-dump/xdpdump.h
tohojo: Hmm, don't think so. I guess you could match on the compiler, but not sure that's a better solution? What happens if you just leave out the ifdef (i.e., why can't these just stay defined as part of the BPF program)?

xdp-dump/xdpdump.c
tohojo: Great! Should simplify some of the matching code as well...

xdp-dump/Makefile
```diff
@@ -0,0 +1,10 @@
+# SPDX-License-Identifier: (GPL-2.0 OR BSD-2-Clause)
+
+XDP_TARGETS  := xdpdump_bpf
+USER_TARGETS := xdpdump
+
+LIB_DIR   = ../lib
+CFLAGS    = -Werror
```
tohojo: BTW, this overrides the default cflags (-O2 -g) set in defines.mk - maybe it's better to add -Werror to those?

xdp-dump/xdpdump.h
chaudron: It will error out as perf_event_header is not defined, and if I include any of the kernel includes all hell breaks loose ;)

xdp-dump/Makefile
chaudron: Will try to do that, hope no other errors exist ;)

xdp-dump/xdpdump.c
chaudron: Had this discussion with Jiri also, he feels its better to warn and continue as people can do what they want to do using a single tool.

xdp-dump/Makefile
tohojo: Shouldn't be, but if you find any please fix them ;)

xdp-dump/xdpdump.c
tohojo: Right, well, as long as "what they want to do" does not end up requiring a full re-implementation of tcpdump...

I guess we can keep it, but just be warned that I plan to push back on any feature creep in this area :)

xdp-dump/xdpdump.h
tohojo: I see. Hmm, I still think checking for a particular header is not ideal. I couldn't find a way to check if we're using the BPF target in the LLVM documentation, but I guess you could just check if __clang__ is defined instead?

tohojo: That was supposed to be `__clang__` - not bold :)

xdp-dump/Makefile
chaudron: Done will take care of this in the next commit

xdp-dump/xdpdump.h
chaudron: There is a \_\_bpf\_\_ especially for this, so used it...

tohojo: Oh, neat! That was what I was looking for, obviously, but didn't find it. So great detective work! ;)

xdp-dump/xdpdump.c
chaudron: I agree it should only allow for the same functionality as the XDP trace, no additional items in this area!

xdp-dump/xdpdump.c
chaudron: Will update the comment to:
`This function will calculate the rough delta between uptime seconds and the epoch time. This is not a precise delta as there is a delay between calling the two functions below (and time() being in seconds), but it's good enough to get a general offset. The delta between packets is still based on the timestamps from the trace infrastructure.`

chaudron:
Update branch with fixes

xdp-dump/xdpdump.c
```diff
@@ -0,0 +1,628 @@
...
+
+/*****************************************************************************
+ * get_epoch_to_uptime_delta()
+ *****************************************************************************/
+static uint64_t get_epoch_to_uptime_delta(void)
+{
+	/* This function will calculate the rough delta between uptime
+	 * seconds and the epoch time. This is not a precise delta as there is
+	 * a delay between calling the two functions below (and time() being in
+         * seconds), but it's good enough to get a general offset. The delta
```
tohojo: Weird indentation

xdp-dump/xdpdump.c
```diff
@@ -0,0 +1,628 @@
...
+	}
+
+	pcap = pcap_open_live(cfg->iface.ifname, cfg->snaplen,
+			      true, 1000, errbuf);
+	if (pcap == NULL) {
+		pr_warn("ERROR: Can't open pcap live interface: %s\n", errbuf);
+		goto error_exit;
+	}
+
+        /* Open the pcap handle for pcap file. */
```
tohojo: And here

xdp-dump/xdpdump_bpf.c
```diff
@@ -0,0 +1,98 @@
...
+/*****************************************************************************
+ * trace_to_perf_buffer()
+ *****************************************************************************/
+static inline void trace_to_perf_buffer(struct xdp_buff *xdp, bool fexit)
+{
+	void *data_end = (void *)(long)xdp->data_end;
+	void *data = (void *)(long)xdp->data;
+	struct pkt_trace_metadata metadata;
+
+        if (data >= data_end)
```
tohojo: More indentation weirdness

xdp-dump/xdpdump_bpf.c
```diff
@@ -0,0 +1,98 @@
...
+		return;
+
+	metadata.ifindex = xdp->rxq->dev->ifindex;
+	metadata.rx_queue = xdp->rxq->queue_index;
+	metadata.pkt_len = (__u16)(data_end - data);
+	metadata.cap_len = metadata.pkt_len;
+        metadata.flags = 0;
```
tohojo: And here

xdp-dump/README.org
```diff
@@ -0,0 +1,126 @@
...
+* The options explained
+The =xdpdump= tool tries to mimic the basic =tcpdump= options, but just in case
+below each of the available options is explained:
+
+** --rx-capture <mode>
+Specify where the ingress packet gets captured. Either at the entry of the XDP
+program and/or exit of the XDP program. Valid options are *fentry*, *fexit*,
+or both *fentry,fexit*. The packet at *fexit* can be modified by the XDP
+program. If you are interested to see both the original and modified packet,
+use the *fentryfexit* option. With this, each packet is captured twice.
```
tohojo: Missing comma between fentry and fexit in this last line. Also, maybe specify that what 'fentry' and 'fexit' means? :)

chaudron: `-use the *fentryfexit* option. With this, each packet is captured twice.
+use the *fentry,fexit* option. With this, each packet is captured twice.
+fentry and fexit refer to the kernels function tracer framework, i.e. function
+entry or exit. In the XDP context, it means the XDP program's entry and exit
+points.
`

chaudron:
Thanks for the review, had to tabify the files which fixed the odd indentations (there where a couple of more places, not sure why my emacs messed this up :). 

I also updated the help text. Let me know if there are more comments...

xdp-dump/README.org
tohojo: Hmm, now that you updated this, I'm wondering whether it wouldn't be less confusing to just drop the 'f'? I.e., just name the user-visible flags 'entry' and 'exit' and drop everything except the last bit of that help message you just added? :)

chaudron: Will change it to 

> Specify where the ingress packet gets captured. Either at the entry of the XDP
program and/or exit of the XDP program. Valid options are *entry*, *exit*,
or both *entry,exit*. The packet at *exit* can be modified by the XDP
program. If you are interested to see both the original and modified packet,
use the *entry,exit* option. With this, each packet is captured twice. In the
XDP context, it means the XDP program's entry and exit points.

tohojo: With that change I don't even think you need the last sentence :)

chaudron:
Removed the extra line... Let me know if you see any other changes needed :)

lib/defines.mk
```diff
@@ -1,4 +1,4 @@
-CFLAGS ?= -O2 -g
+CFLAGS ?= -O2 -g -Werror
```
tohojo: Turns out this did actually break things; could you please add this as well?

`BPF_CFLAGS ?= -Wno-visibility`

xdp-dump/README.org
```diff
@@ -0,0 +1,128 @@
...
+The =xdpdump= tool tries to mimic the basic =tcpdump= options, but just in case
+below each of the available options is explained:
+
+** --rx-capture <mode>
+Specify where the ingress packet gets captured. Either at the entry of the XDP
+program and/or exit of the XDP program. Valid options are *entry*, *exit*,
+or both *entry,exit*. The packet at *exit* can be modified by the XDP
+program. If you are interested to see both the original and modified packet,
+use the *entry,exit* option. With this, each packet is captured twice. In the
+XDP context, it means the XDP program's entry and exit points.
```
tohojo: This last sentence is still there?

xdp-dump/xdpdump.c
```diff
@@ -0,0 +1,628 @@
...
+					 packet, h.caplen, i);
+				printf("  %s\n", hline);
+			}
+		}
+		captured_packets++;
+	}
+	exit_pcap = NULL;
+	rc = true;
+
+	fprintf(stderr, "\n%"PRIu64" packets captured\n", captured_packets);
```
tohojo: Your use of PRIu64 causes compile errors on my Ubuntu machine, because of a missing `#include <inttypes.h>`

xdp-dump/README.org
```diff
@@ -0,0 +1,128 @@
...
+** Running xdpdump
+The syntax for running =xdpdump= is:
+
+#+begin_src
+Usage: xdpdump [options]
+
+ XDPDump tool to dump network traffic
+
+Options:
+     --rx-capture <mode>    Capture point for the rx direction (valid values: entry,exit)
```
tohojo: Should also include what the default is

xdp-dump/xdpdump.c
```diff
@@ -0,0 +1,628 @@
...
+		ts = e->time + ctx->epoch_delta;
+		if (ctx->pcap_dumper) {
+			struct pcap_pkthdr h;
+
+			h.ts.tv_sec = ts / 1000000000ULL;
+			h.ts.tv_usec = ts % 1000000000ULL / 1000;
+			h.caplen = min(e->metadata.cap_len, ctx->cfg->snaplen);
+			h.len = e->metadata.pkt_len;
+			pcap_dump((u_char *) ctx->pcap_dumper, &h,
+				  e->packet);
```
tohojo: I wasn't getting any output when piping to tcpdump; adding a pcap_dump_flush() call here helped. Maybe detect if the output is a file and only flush if it isn't, though?

xdp-dump/xdpdump.c
```diff
@@ -0,0 +1,628 @@
...
+		} else {
+			int i;
+			char hline[SNPRINTH_MIN_BUFFER_SIZE];
+			bool fexit = e->metadata.flags & MDF_DIRECTION_FEXIT;
+
+			printf("%llu.%09lld: packet size %u bytes, "
+			       "captured %u bytes on "
+			       "if_index %u, rx queue %u, @%s\n",
```
tohojo: The `@entry`/`@exit` was confusing to me the first time I ran this. Also, on exit we should also print the verdict; otherwise it will be confusing if people run this on a packet filter and they don't see any packets getting filtered. I also think we could make the output a bit shorter... 

So maybe change the output format to something like:

`1584701151.122081500: @entry: pkt size 98 bytes, captured 98 bytes on if_index 4, rx queue 5`
and
`1584701151.122081500: @exit[DROP]: pkt size 98 bytes, captured 98 bytes on if_index 4, rx queue 5`

(obviously we would like this in tcpdump output as well, but I guess we'll have to wait for pcapng support for that).

I'm also wondering if we should have a 'verbose mode' and only show the packet data dump if it is enabled. I.e., without verbose mode we would just do:
`1584701151.122081500: @exit[DROP]: pkt size 98 bytes on if_index 4, rx queue 5`

and with verbose mode we'd add the capture len and hex dump?

xdp-dump/README.org
chaudron: Forgot to push this README.org file :(

lib/defines.mk
chaudron: Odd, did not break anything on my side (compilers :(
Will add it

xdp-dump/README.org
chaudron: Will add this.

xdp-dump/xdpdump.c
chaudron: Good idea will flush if the output is '-'

PR #14, opened 2020-04-21
APIs to get more program information, including xdp-loader using it

chaudron:
This adds API's to get the information needed to see what sub-programs are loaded for multi-programs.

The xdp-loader has been changed to show which programs are loaded on the interface.

These APIs are also needed for xdp-dump to figure out which programs to attach to.

### DIFF
--- a/headers/xdp/libxdp.h
+++ b/headers/xdp/libxdp.h
@@ -27,6 +27,8 @@ struct xdp_program *xdp_program__from_id(__u32 prog_id);
 void xdp_program__free(struct xdp_program *xdp_prog);
 
 const char *xdp_program__name(struct xdp_program *xdp_prog);
+const unsigned char *xdp_program__tag(struct xdp_program *xdp_prog);
+uint32_t xdp_program__id(struct xdp_program *xdp_prog);
 unsigned int xdp_program__run_prio(struct xdp_program *xdp_prog);
 void xdp_program__set_run_prio(struct xdp_program *xdp_prog, unsigned int run_prio);
 bool xdp_program__chain_call_enabled(struct xdp_program *xdp_prog,
@@ -47,3 +49,5 @@ int xdp_multiprog__attach(struct xdp_multiprog *mp,
                           int ifindex, bool force,
                           enum xdp_attach_mode mode);
 struct xdp_multiprog *xdp_multiprog__get_from_ifindex(int ifindex);
+struct xdp_program *xdp_multiprog__next_prog(struct xdp_program *prog,
+					     struct xdp_multiprog *mp);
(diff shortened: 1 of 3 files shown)


tohojo:
Thanks for doing this!

One comment - I'm not sure we should be output'ing the main (dispatcher) program information at all? Rather just leave an empty line, or maybe somehow mark that this is a multi-prog dispatcher; the component programs are the main user-facing piece of information that is relevant in this case, after all, isn't it?

chaudron:
> One comment - I'm not sure we should be output'ing the main (dispatcher) program information at all? Rather just leave an empty line, or maybe somehow mark that this is a multi-prog dispatcher; the component programs are the main user-facing piece of information that is relevant in this case, after all, isn't it?

That was my initial thought also, but other tools like _ip_ and _bpftool_ show the actual program, not the subprograms. So hiding it hear might look like we obscuring something. Also if someone wants to add a kprobe it's more obvious what is present.

tohojo:
Hmm, yeah, but we could format it differently? So instead of
```
Interface        Prio  Program name     Mode     ID   Tag               Chain actions
-------------------------------------------------------------------------------------
lo               10    xdpfilt_blk_all  skb      73   fc947ec6f178c4d7  XDP_PASS
eth0             50    xdp_main         skb      39   c6413e14cc63422d  XDP_PASS
 =>              10    xdp_test_II               43   a768e85830a0f178  XDP_PASS
 =>              50    xdp_test_I                47   b5a46c6e9935298c  XDP_PASS
eth1             10    xdpfilt_blk_all  skb      80   6deef7357e7b4530  XDP_PASS
```

maybe something like

```
Interface        Prio  Program name     Mode     ID   Tag               Chain actions
-------------------------------------------------------------------------------------
lo               10    xdpfilt_blk_all  skb      73   fc947ec6f178c4d7  XDP_PASS
eth0                   (multi-prog dispatcher)   39   c6413e14cc63422d
 =>              10    xdp_test_II               43   a768e85830a0f178  XDP_PASS
 =>              50    xdp_test_I                47   b5a46c6e9935298c  XDP_PASS
eth1             10    xdpfilt_blk_all  skb      80   6deef7357e7b4530  XDP_PASS
```

chaudron:
> Hmm, yeah, but we could format it differently?

I played with this a but, but found not nice way that I liked. I can go with your format, but are you ok not showing the actual program name of the dispatcher?

tohojo:
Hmm, how about I just play around with this a bit and fix up the output format before merging?

chaudron:
> Hmm, how about I just play around with this a bit and fix up the output format before merging?

Yes please, be my guest ;)

tohojo:
Ended up just getting rid of the 'prio' and 'chain call action' columns for
non-multiprog for now... Gives at least some visual separation between the two
types.

chaudron:
> Ended up just getting rid of the 'prio' and 'chain call action' columns for
> non-multiprog for now... Gives at least some visual separation between the two
> types.

Thanks for merging, and looking at the format!

PR #15, opened 2020-04-22
xdp-dump: fix multiple interfaces using the same XDP program

chaudron:
When multiple interfaces are using the same pinned XDP program we
should only capture packets from the interface requested. Without this
fix packets from all interfaces using this XDP program are captured.

To replicate the issues load a XDP program (pinning it), and assign it
to multiple interfaces. For example:

$ bpftool prog load test_bpf.o /sys/fs/bpf/eelcos_prog
$ bpftool net attach xdpgeneric pinned /sys/fs/bpf/eelcos_prog dev eth0
$ bpftool net attach xdpgeneric pinned /sys/fs/bpf/eelcos_prog dev eth1
$ xdpdump -i eth1

The last command will also show packets from eth0.

### DIFF
--- a/xdp-dump/xdpdump.c
+++ b/xdp-dump/xdpdump.c
@@ -395,11 +395,14 @@ static bool capture_on_interface(struct dumpopt *cfg)
 	int                          perf_map_fd;
 	int                          prog_fd;
 	int                          rc = false;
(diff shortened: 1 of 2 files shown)


tohojo:
Yeah, probably a good idea to fix that :)

Issue #1, opened 2019-11-06
Build dependency on bsd/string.h


netoptimizer:
When building xdp-tools, it fails with the following errors:

```
$ make
[...]
make[3]: Entering directory '/home/hawk/git/xdp-tools/lib/util'
cc -I../libbpf/src -Wall -I../../headers -c -o params.o params.c
cc -I../libbpf/src -Wall -I../../headers -c -o logging.o logging.c
cc -I../libbpf/src -Wall -I../../headers -c -o util.o util.c
util.c:9:10: fatal error: bsd/string.h: No such file or directory
 #include <bsd/string.h>
          ^~~~~~~~~~~~~~
compilation terminated.
```

Are there any specific reason for depending on ```bsd/string.h``` ?

netoptimizer:
On Fedora this seems to be avail in package ```libbsd-devel```

```
$ dnf whatprovides '*/bsd/string.h'
[...]
libbsd-devel-0.9.1-2.fc29.i686 : Development files for libbsd
Repo        : fedora
Matched from:
Filename    : /usr/include/bsd/string.h
```

tohojo:
Jesper Dangaard Brouer <notifications@github.com> writes:

> On Fedora this seems to be avail in package ```libbsd-devel```
>
> ```
> $ dnf whatprovides '*/bsd/string.h'
> [...]
> libbsd-devel-0.9.1-2.fc29.i686 : Development files for libbsd
> Repo        : fedora
> Matched from:
> Filename    : /usr/include/bsd/string.h
> ```

It's for strlcpy(); can just switch to using regular old strncpy()...

netoptimizer:
> It's for strlcpy(); can just switch to using regular old strncpy()...

Yes, it might be better to reduce dependencies.

PR #16, opened 2020-04-28
xdp-dump: double rlimit also for ENOTSUP errors

chaudron:
When running some netns tests with xdpdump on the latest kernel I
noticed that sometimes loading fails with ENTOSUP due to a too low
rlimit. Fixed code to also try increasing the rlimit on this error.

### DIFF
--- a/xdp-dump/xdpdump.c
+++ b/xdp-dump/xdpdump.c
@@ -492,9 +492,11 @@ rlimit_loop:
 	if (err) {
 		char err_msg[STRERR_BUFSIZE];
 
-		if (err == -EPERM) {
-			pr_debug("Permission denied when loading eBPF object; "
-				 "raising rlimit and retrying\n");
+		if (err == -EPERM || err == -ENOTSUP) {
+			pr_debug("%s when loading eBPF object; "
+				 "raising rlimit and retrying\n",
+				 err == -EPERM ? "Permission denied" :
+				 "Operation not supported");
 
 			if (!double_rlimit()) {
 				bpf_object__close(trace_obj);


tohojo:
Huh. Any idea which code path in the kernel emits that error?

chaudron:
> Huh. Any idea which code path in the kernel emits that error?

Did not look into the details :) Here is what I get from the detailed libbpf output:
```

libbpf: collecting relocating info for: 'fentry/func'
libbpf: relo for shdr 8, symb 68, value 0, type 1, bind 1, name 16 ('capture_if_ifindex'), insn 7
libbpf: found data map 1 (xdpdump_.data, sec 8, off 0) for insn 7
libbpf: relo for shdr 7, symb 71, value 0, type 1, bind 1, name 116 ('xdpdump_perf_map'), insn 29
libbpf: found map 0 (xdpdump_perf_map, sec 7, off 0) for insn 29
libbpf: collecting relocating info for: 'fexit/func'
libbpf: relo for shdr 8, symb 68, value 0, type 1, bind 1, name 16 ('capture_if_ifindex'), insn 8
libbpf: found data map 1 (xdpdump_.data, sec 8, off 0) for insn 8
libbpf: relo for shdr 7, symb 71, value 0, type 1, bind 1, name 116 ('xdpdump_perf_map'), insn 34
libbpf: found map 0 (xdpdump_perf_map, sec 7, off 0) for insn 34
libbpf: Error in bpf_object__probe_name():Operation not permitted(1). Couldn't load basic 'r0 = 0' BPF program.
libbpf: Probe #0 failed with -1.
libbpf: Error in bpf_object__probe_global_data():Operation not permitted(1). Couldn't create simple array map.
libbpf: Probe #1 failed with -1.
libbpf: kernel doesn't support global data
libbpf: failed to load object './xdpdump_bpf.o'
ERROR: Can't load eBPF object: Operation not supported(-95)
```

Any idea how to trace this quickly back to the kernel?

tohojo:

>> Huh. Any idea which code path in the kernel emits that error?
>
> Did not look into the details :) Here is what I get from the detailed libbpf output:
> ```
>
> libbpf: collecting relocating info for: 'fentry/func'
> libbpf: relo for shdr 8, symb 68, value 0, type 1, bind 1, name 16 ('capture_if_ifindex'), insn 7
> libbpf: found data map 1 (xdpdump_.data, sec 8, off 0) for insn 7
> libbpf: relo for shdr 7, symb 71, value 0, type 1, bind 1, name 116 ('xdpdump_perf_map'), insn 29
> libbpf: found map 0 (xdpdump_perf_map, sec 7, off 0) for insn 29
> libbpf: collecting relocating info for: 'fexit/func'
> libbpf: relo for shdr 8, symb 68, value 0, type 1, bind 1, name 16 ('capture_if_ifindex'), insn 8
> libbpf: found data map 1 (xdpdump_.data, sec 8, off 0) for insn 8
> libbpf: relo for shdr 7, symb 71, value 0, type 1, bind 1, name 116 ('xdpdump_perf_map'), insn 34
> libbpf: found map 0 (xdpdump_perf_map, sec 7, off 0) for insn 34
> libbpf: Error in bpf_object__probe_name():Operation not permitted(1). Couldn't load basic 'r0 = 0' BPF program.
> libbpf: Probe #0 failed with -1.
> libbpf: Error in bpf_object__probe_global_data():Operation not permitted(1). Couldn't create simple array map.
> libbpf: Probe #1 failed with -1.
> libbpf: kernel doesn't support global data
> libbpf: failed to load object './xdpdump_bpf.o'
> ERROR: Can't load eBPF object: Operation not supported(-95)
> ```
>
> Any idea how to trace this quickly back to the kernel?

This seems like the kernel is actually returning EPERM, but libbpf turns
that into ENOTSUP because it's part of its feature probing. So I'd say
this is just a bug in libbpf that we should fix instead of work
around...

chaudron:
Looking at the libbpf bpf capabilities code, they do not really care about the result code, i.e. if the check fails they mark the capability as not supported. Which kind of makes sense, as without the check passing we can not verify it really is supported. In addition, libbpf has no code for increasing the rlimit, so I do not think it makes sense to add it (also as it might not make sense for libbpf to do this on behalf of the application).

So I think the patch as above still makes the most sense...

tohojo:

> Looking at the libbpf bpf capabilities code, they do not really care
> about the result code, i.e. if the check fails they mark the
> capability as not supported. Which kind of makes sense, as without the
> check passing we can not verify it really is supported. In addition,
> libbpf has no code for increasing the rlimit, so I do not think it
> makes sense to add it (also as it might not make sense for libbpf to
> do this on behalf of the application).

No, it doesn't make sense for libbpf to try to raise the rlimit; but I
still think it's a bug that it turns EPERM errors into ENOTSUP error
like that... Something like this (partial and untested):

diff --git a/tools/lib/bpf/libbpf.c b/tools/lib/bpf/libbpf.c
index 8f480e29a6b0..10df143afbde 100644
--- a/tools/lib/bpf/libbpf.c
+++ b/tools/lib/bpf/libbpf.c
@@ -180,6 +180,7 @@ struct bpf_capabilities {
        __u32 btf_func_global:1;
        /* kernel support for expected_attach_type in BPF_PROG_LOAD */
        __u32 exp_attach_type:1;
+       __u32 probe_perm_error:1;
 };
 
 enum reloc_type {
@@ -3228,7 +3229,8 @@ bpf_object__probe_global_data(struct bpf_object *obj)
        if (ret >= 0) {
                obj->caps.global_data = 1;
                close(ret);
-       }
+       } else if (errno == EPERM)
+               obj->caps.probe_perm_error = 1;
 
        close(map);
        return 0;
@@ -5291,7 +5293,7 @@ static int bpf_object__sanitize_maps(struct bpf_object *obj)
                        continue;
                if (!obj->caps.global_data) {
                        pr_warn("kernel doesn't support global data\n");
-                       return -ENOTSUP;
+                       return obj->caps.probe_perm_error ? -EPERM : -ENOTSUP;
                }
                if (!obj->caps.array_mmap)
                        m->def.map_flags ^= BPF_F_MMAPABLE;

chaudron:
Thanks for making me think about this again. Your solution is what I had in mind, but it did not make sense to me. However, after digging trough the capabilities and use cases of EPERM, I think we can say that EPERM is related to memory, and not any of the other "none supporting capabilities" failures. 

So I make the following change, what do you think?

```
diff --git a/src/libbpf.c b/src/libbpf.c
index ff91742..f55c1fd 100644
--- a/src/libbpf.c
+++ b/src/libbpf.c
@@ -3330,10 +3330,14 @@ bpf_object__probe_caps(struct bpf_object *obj)
 
        for (i = 0; i < ARRAY_SIZE(probe_fn); i++) {
                ret = probe_fn[i](obj);
-               if (ret < 0)
+               if (ret < 0) {
                        pr_debug("Probe #%d failed with %d.\n", i, ret);
+                       if (ret == -EPERM) {
+                               pr_perm_msg(ret);
+                               return ret;
+                       }
+               }
        }
-
        return 0;
 }

```

I'll sleep on it, and will send a patch upstream tomorrow ;)

However, we still might need this patch to be backward compatible, or are we ok with this specific fail case? If so I'll add a note some where in the documentation.

tohojo:

> Thanks for making me think about this again.

You're welcome ;)

> Your solution is what I had in mind, but it did not make sense to me.
> However, after digging trough the capabilities and use cases of EPERM,
> I think we can say that EPERM is related to memory, and not any of the
> other "none supporting capabilities" failures.
>
> So I make the following change, what do you think?

Oh yeah, that does seems like a better way to fix it :)

> I'll sleep on it, and will send a patch upstream tomorrow ;)

Great!

> However, we still might need this patch to be backward compatible, or
> are we ok with this specific fail case? If so I'll add a note some
> where in the documentation.

At least until we release a "1.0" of xdp-tools I think it's fine to just
demand a specific minimum libbpf version...

chaudron:
Looks like my patch got accepted on bpf-next, commit fd9eef1a132d. However, how do we make sure this gets included in xdp-tools, @tohojo?

tohojo:

> Looks like my patch got accepted on bpf-next, commit fd9eef1a132d.
> However, how do we make sure this gets included in xdp-tools, @tohojo?

Just wait for it to make it into the github libbpf repo, then bump the
libbpf reference in this repo... :)

PR #18, opened 2020-04-29
Makefile: distclean submodule libbpf

netoptimizer:
It is possible to get stuck with an old git submodule commit id for libbpf, that is too old to e.g. compile against.

Offer a way out via having distclean remove the submodule directory.
As this will caused configure script to call git submodule update.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,12 +33,18 @@ help:
 config.mk:
 	sh configure
 
+clobber_submodules:
+	@if [ -d "lib/libbpf/src" ]; then \
+		rm -r lib/libbpf && mkdir lib/libbpf ;\
+		echo "removed submodule" ;\
+	fi
+
 clobber:
 	touch config.mk
 	$(MAKE) clean
 	rm -f config.mk cscope.*
 
-distclean: clobber
+distclean: clobber clobber_submodules
 
 clean:
 	@for i in $(SUBDIRS); \


tohojo:
This seems a bit scary - it will also wipe out any modifications people have in their libbpf directory. For instance, I have a bunch of branches in that repo, for testing various things with xdp-tools as the consumer...

Couldn't we do something a bit less...destructive? :)

netoptimizer:
But currently the "distclean" doesn't cleanup libbpf subdir, which seems weird.

tohojo:
But wouldn't it be enough to do something like `make -C lib/libbpf clean && git submodule update --force`?

netoptimizer:
> But wouldn't it be enough to do something like `make -C lib/libbpf clean && git submodule update --force`?

That looks fine.

But why do we need to --force?

tohojo:
Hmm, don't think we do, actually...

tohojo:
Also, doesn't 'make clean' already recurse into lib/libbpf?

netoptimizer:
I guess that I really want is something that can trigger a:

```
git submodule update
```

It happens regularly that I get back to a old git tree, do a ```git pull``` , which actually pickup newer git id for the submodule, but I keep using the older version of the submodule code... until it breaks and I notice (and get really really confused, why the build breaks).

tohojo:
Yeah, that's a good point... Hmm, would a warning be enough? If we're just checking (and warning) we could do that on every 'make'. Alternatively (or in addition), we could add `git submodule update` to distclean?

netoptimizer:
I just don't know how to detect that the submodule is out-of-sync, so I/we can generate such a warning.
Do you know?

tohojo:
This seems to match when the submodule is modified:

`git submodule status lib/libbpf | grep -q '^+'`

As for distclean, just adding this at the end of the existing clobber rule should be enough, I think?

`[ -e .git ] && [ -f .gitmodules ] && git submodule update`

netoptimizer:
> This seems to match when the submodule is modified:
> 
> `git submodule status lib/libbpf | grep -q '^+'`

Yes, we can use this to generate a warning.

Maybe also use this in make clean, to trigger a `git submodule update`?
Or does that break your use-case, when you have modificiations in lib/libbpf ?

> As for distclean, just adding this at the end of the existing clobber rule should be enough, I think?
> 
> `[ -e .git ] && [ -f .gitmodules ] && git submodule update`

Sure, **BUT** notice that using this directly in the Makefile have side-effects, if `[ -e .git ]` fails, the Makefile will see it as a make failure.  Hint that is why I used the full `if [ -d XXX ]; then X; fi` in my patch.

tohojo:

>> This seems to match when the submodule is modified:
>> 
>> `git submodule status lib/libbpf | grep -q '^+'`
>
> Yes, we can use this to generate a warning.
>
> Maybe also use this in make clean, to trigger a `git submodule update`?
> Or does that break your use-case, when you have modificiations in
> lib/libbpf ?

Usually the first thing I do after updating the contents of lib/libbpf
is a 'make clean' in the xdp-tools directory so yeah, that would break
things for me. A warning on clean would be fine, though :)

>> As for distclean, just adding this at the end of the existing clobber rule should be enough, I think?
>> 
>> `[ -e .git ] && [ -f .gitmodules ] && git submodule update`
>
> Sure, **BUT** notice that using this directly in the Makefile have
> side-effects, if `[ -e .git ]` fails, the Makefile will see it as a
> make failure.  Hint that is why I used the full `if [ -d XXX ]; then
> X; fi` in my patch.

Oh right, good point; I trust your patch will end up doing the right
thing, though ;)

netoptimizer:
> I trust your patch will end up doing the right thing, though ;)

See if you like PR #20

PR #20, opened 2020-04-29
Makefile: git submodule detect and warn if out-of-sync

netoptimizer:
It is a reoccuring issue that git submodule gets out-of-sync, which cause confusion and issues.

 End-user needs to run:  git submodule update

We cannot run this automatically, because there are legimite reasons for the submodule to be out-of-sync, e.g. when testing local changes.

Thus, this patch detect SHA-1 doesn't match, and just warns end-users, together with a hint to call: git submodule update

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ endif
 
 SUBDIRS=lib xdp-filter xdp-loader xdp-dump
 
-all: config.mk
+all: config.mk check_submodule
 	@set -e; \
 	for i in $(SUBDIRS); \
 	do echo; echo $$i; $(MAKE) -C $$i; done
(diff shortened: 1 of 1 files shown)


netoptimizer:
Per discussion in PR #18

tohojo:
Erm, GitHub seems to think the commit is empty?

netoptimizer:
> Erm, GitHub seems to think the commit is empty?

LOL - It seems I forgot got call ```stg refresh``` get changes integrated, usually [stgit](http://www.procode.org/stgit/) warns on different operations when having uncommitted changes, but I've tried the ```stg publish``` which didn't warn me.

netoptimizer:
I fixed it in the stg branch, and ran `stg publish` again.  This resulted in two commit's one empty and the new with same text and contents. I'll try push a new branch, with a clean history.

tohojo:

> I fixed it in the stg branch, and ran `stg publish` again.  This
> resulted in two commit's one empty and the new with same text and
> contents. I'll try push a new branch, with a clean history.

No worries, I can squash those when merging; going to do a bit of
cleanup of the Makefile anyway :)

netoptimizer:
I will close this one, in favor of PR #21

PR #21, opened 2020-04-30
Makefile: git submodule detect and warn if out-of-sync

netoptimizer:
It is a reoccuring issue that git submodule gets out-of-sync, which cause confusion and issues.

 End-user needs to run:  git submodule update

We cannot run this automatically, because there are legimite reasons for the submodule to be out-of-sync, e.g. when testing local changes.

Thus, this patch detect SHA-1 doesn't match, and just warns end-users, together with a hint to call: git submodule update

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ endif
 
 SUBDIRS=lib xdp-filter xdp-loader xdp-dump
 
-all: config.mk
+all: config.mk check_submodule
 	@set -e; \
 	for i in $(SUBDIRS); \
 	do echo; echo $$i; $(MAKE) -C $$i; done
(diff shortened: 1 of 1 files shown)


netoptimizer:
This related to PR #20 and should be an alternative to PR #18

netoptimizer:
I spotted a spelling error in the commit message.
I'll try to fix that, and push branch again (mostly to see how GitHub react).

netoptimizer:
Did a controversial(?) `git push --force` and GitHub did detect this.

tohojo:

> Did a controversial(?) `git push --force` and GitHub did detect this.

Yeah force-pushing is fine on PR branches - just don't do it on master ;)

tohojo:
Merged as d862849bf8c29d770299d3335ff16f8588c0cf59 (just cherry-picked, and
Github doesn't seem to autodetect that)

PR #22, opened 2020-04-30
xdp-dump: add support for >16 XDP eBPF entry function names

chaudron:
The Linux API does not provide the full name of the attached eBPF
entry function if it's longer than 15 characters. xdpdump will try to
guess the correct function name from the available BTF debug
information. However, if multiple functions exist with the same
leading name, it can not pick the correct one. It will dump the
available functions, and you can choose the correct one, and supply it
with the -m option. For example:

$ xdpdump -i eth0
ERROR: Can't identify the full XDP main function!
The following is a list of candidates:
  xdp_test_prog_with_a_long_name
  xdp_test_prog_with_a_long_name_too
  xdp_test_prog_w
Please use the -m option to pick the correct one.

$ ./xdpdump -i eth0 -m xdp_test_prog_with_a_long_name
1588252724.025055871: @entry: packet size 102 bytes on if_index 2, rx queue 0
1588252724.026463296: @entry: packet size 66 bytes on if_index 2, rx queue 0
1588252724.026786186: @entry: packet size 66 bytes on if_index 2, rx queue 0

### DIFF
--- a/xdp-dump/xdpdump.c
+++ b/xdp-dump/xdpdump.c
@@ -130,6 +130,22 @@ static const char *get_xdp_action_string(enum xdp_action act)
 }
 
 /*****************************************************************************
+ * get_capture_mode_string()
+ *****************************************************************************/
+static const char *get_capture_mode_string(unsigned int mode)
+{
+	switch(mode) {
+	case RX_FLAG_FENTRY:
+		return "entry";
+	case RX_FLAG_FEXIT:
+		return "exit";
+	case RX_FLAG_FENTRY | RX_FLAG_FEXIT:
+		return "entry/exit";
+	}
+	return "unknown";
+}
+
+/*****************************************************************************
  * snprinth()
  *****************************************************************************/
 #define SNPRINTH_MIN_BUFFER_SIZE sizeof("0xffff:  00 11 22 33 44 55 66 77 88" \
@@ -330,6 +346,13 @@ static bool capture_on_legacy_interface(struct dumpopt *cfg)
 			goto error_exit;
 		}
 	}
+
+	/* No more error conditions, display some capture information */
+	fprintf(stderr, "listening on %s, link-type %s (%s), "
+		"capture size %d bytes\n", cfg->iface.ifname,
+		pcap_datalink_val_to_name(pcap_datalink(pcap)),
+		pcap_datalink_val_to_description(pcap_datalink(pcap)),
+		cfg->snaplen);
 
 	/* Loop for receive packets on live interface. */
 	exit_pcap = pcap;
@@ -647,6 +670,12 @@ rlimit_loop:
 			goto error_exit;
 		}
 	}
+
+	/* No more error conditions, display some capture information */
+	fprintf(stderr, "listening on %s, ingress XDP program %s, "
+		"capture mode %s, capture size %d bytes\n",
+		cfg->iface.ifname, prog_name,
+		get_capture_mode_string(cfg->rx_capture), cfg->snaplen);
 
 	/* Setup perf context */
 	memset(&perf_ctx, 0, sizeof(perf_ctx));
--- a/xdp-dump/xdpdump.c
+++ b/xdp-dump/xdpdump.c
@@ -16,6 +16,7 @@
 #include <unistd.h>
 
 #include <bpf/bpf.h>
+#include <bpf/btf.h>
 #include <bpf/libbpf.h>
 
 #include <linux/perf_event.h>
@@ -55,6 +56,7 @@ static const struct dumpopt {
 	struct iface          iface;
 	uint32_t              snaplen;
 	char                 *pcap_file;
+	char                 *program_names;
 	unsigned int          rx_capture;
 } defaults_dumpopt = {
 	.list_interfaces = false,
@@ -76,6 +78,11 @@ static struct prog_option xdpdump_options[] = {
 		      .short_opt = 'i',
 		      .metavar = "<ifname>",
 		      .help = "Name of interface to capture on"),
+	DEFINE_OPTION("program-names", OPT_STRING, struct dumpopt,
+		      program_names,
+		      .short_opt = 'p',
+		      .metavar = "<prog>",
+		      .help = "Specific program to attach to"),
 	DEFINE_OPTION("snapshot-length", OPT_U32, struct dumpopt, snaplen,
 		      .short_opt = 's',
 		      .metavar = "<snaplen>",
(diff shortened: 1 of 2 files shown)


chaudron:
Note that the long option name "--multiprog-attach" does not make 100% sense, but it's because I will re-use it for multi-attach programs where you can specify the actual program(s) to attach to.

tohojo:

> Note that the long option name "--multiprog-attach" does not make 100%
> sense, but it's because I will re-use it for multi-attach programs
> where you can specify the actual program(s) to attach to.

Not sure I like that name even for the multiprog case ;) How about just
'-p, --program-name' ?

chaudron:
> Not sure I like that name even for the multiprog case ;) How about just '-p, --program-name' ?

This might work in the single program name case, but what if it would be a list of programs to attach to? Want to make sure I can re-use the option, for the multiprog case. -p would not collide with tcpdump for our use case so it would be ok.

tohojo:

>> Not sure I like that name even for the multiprog case ;) How about just '-p, --program-name' ?
>
> This might work in the single program name case, but what if it would
> be a list of programs to attach to? Want to make sure I can re-use the
> option, for the multiprog case. -p would not collide with tcpdump for
> our use case so it would be ok.

Oh, you want to be able to attach to multiple programs in one go? That
would potentially capture each packet many times, right? How are you
planning to encode that in pcapng?

As for the switch option, well, --program-names? :)

chaudron:
> Ah, so you're just encoding the capture point in the interface name? I assumed it would be a separate field?

Yes, you can see all interfaces in Wireshark's capture file properties, you can also see some stats per interface. Then the individual packet view will have the capturing interface (and direction once we support XDP egress).

tohojo:

>> Ah, so you're just encoding the capture point in the interface name? I assumed it would be a separate field?
>
> Yes, you can see all interfaces in Wireshark's capture file
> properties, you can also see some stats per interface. Then the
> individual packet view will have the capturing interface (and
> direction once we support XDP egress).

Just to be sure I understand correctly - if you're capturing on eth0,
you would end up with two different 'interface names', 'eth0-fentry' and
'eth0-fexit'?

chaudron:
Changed to option to -p and --program-names. Updated code and documentation.

xdp-dump/xdpdump.c
```diff
@@ -55,6 +56,7 @@ static const struct dumpopt {
 	struct iface          iface;
 	uint32_t              snaplen;
 	char                 *pcap_file;
+	char                 *multiprogs;
```
tohojo, line 59: Why is this still called 'multiprogs'? :)

xdp-dump/xdpdump.c
```diff
@@ -385,6 +392,85 @@ static bool capture_on_legacy_interface(struct dumpopt *cfg)
...
+	int         nr_types;
+	int         matches = 0;
+	char       *func_name_ptr = NULL;
+	struct btf *btf = NULL;
+
+	/*
+	 * If the function name is longer than 15 characters the bpf_prog_info
+	 * structure will truncate it to 15 characters. Here we try to see if
+	 * we need to check BTF information to figure out the full function
+	 * name.
```
tohojo, line 410: Since you already implemented the looping through names, why not just do it unconditionally? That would effectively give us 'prefix matching' so the user only needs to specify the shortest unambiguous function name prefix - kinda like getopt does with long opts...

xdp-dump/xdpdump.c
chaudron: good question changed it :)

xdp-dump/xdpdump.c
chaudron: Decided not to do this, as this is the ambiguity we try to workaround introduced in the kernel. In this case, I rather have them explicitly specify the full name!

tohojo: But since you just print out the list if there's any ambiguity that would still happen whether the ambiguity is caused by the kernel length limit or because someone just entered too short a name? And in both cases we'd ask for a longer name? So what's the harm in saving a bit of typing if there is no ambiguity? Or am I missing something else here?

chaudron: Your suggestion would work, but it might hurt lazy people (like me) in the long run. And anyway if I need to type 16+ chars, I would probably cut/paste the function name :)

I'm quite sure (I will at least) script around xdpdump to get traces in a specific use case, now let's assume it all worked because I used a short name that was unique enough, i.e. xdp_test_prog_with_a_long_name_ with the following in the file:

```
# bpftool btf dump id 1072 | grep xdp_test_prog_w
[7] FUNC 'xdp_test_prog_with_a_long_name' type_id=5
[9] FUNC 'xdp_test_prog_with_a_long_name_too' type_id=8
```

Now a new function xdp_test_prog_with_a_long_name_for_testing, now this will result in my scripts to fail. Not having the auto-expansion will protect people like me from this. In addition, if I sent people output of the xdpdump command line, they understand what I was (they need to) capturing independent of the version of the XDP code running.

In addition in the future if we support multiprogram's it can become even more ambiguous as they could supply a command line like: `-p a,b,er,c`. And I have no idea what programs they are tracing if they supply the command line output to support (guess if they have pcapng we could now from the trace).

tohojo: Heh, your argument basically boils down to "let's not add this convenience feature, because users are are just going to do stupid things with it", doesn't it? ;)

I mean, if you're worried that your scripts are going to fail because you didn't provide the full name, just...make sure you provide the full name?

Not going to insist if you're really against this, though, just thought it was an obvious convenience feature that you had basically already implemented :)

chaudron: > Heh, your argument basically boils down to "let's not add this convenience feature, because users are are just going to do stupid things with it", doesn't it? ;)

Yes, this is correct, and the reason why I'm against it, in this case, is that this tool is mainly used to troubleshoot problems, i.e. used by customer support asking operators who might not have the technical knowledge to gather information on their behave. Now let's assume tech support tested something in their lab, and want "xdpdump -i eth -p xdp_test_prog_with_a_long_name_" (yes the argument can be you should not do this, but my experience is that this will not happen in real life as humans are lazy by nature). Now the operator executes this in his production environment and gets the result this is not possible and needs to pick one more specific function... Here you lose another day (assuming troubleshooting happens in off-hours).

Let's keep the full name for now, and see if we ever get feedback from the field... We can always add it later (removing might be harder ;)

tohojo: Heh, fair enough. BTW, how about printing out the actual function being traced at the beginning of the execution, to aid with something like this? Similar to how tcpdump starts out printing this: `listening on eth1, link-type EN10MB (Ethernet), capture size 262144 bytes`

chaudron: Great idea, done in separate commit :)

xdp-dump/xdpdump.c
```diff
@@ -648,6 +655,11 @@ static bool capture_on_interface(struct dumpopt *cfg)
 		}
 	}
 
+	/* No more error conditions, display some capture information */
+	fprintf(stderr, "listening on %s, ingress XDP program %s, "
+		"capture size %d bytes\n", cfg->iface.ifname, prog_name,
+		cfg->snaplen);
```
tohojo, line 661: ", capture mode entry,exit" ?

chaudron: Lazy me, forgot! I added it to the new commit and also updated the documentation to show it.

chaudron:
Thanks for merging!

PR #23, opened 2020-05-04
xdp-dump: push down snaplen to the bpf capture program

chaudron:
Push down the snaplen to the ebpf program so only the requered number
of bytes are added to the ring buffer.


tohojo:
Heh, yeah, that seems like a good idea ;)

Issue #17, opened 2020-04-29
xdpdump fails when XDP function BPF-prog names are too long


netoptimizer:
xdpdump fails indicating that it cannot find the eBPF object file, which first of all is a VERY confusing error message, because it is not the real problem.

```
$ sudo ./xdpdump -v -i mlx5p1
Looking for './xdpdump_bpf.o'
Loading bpf file 'xdpdump_bpf.o' from './xdpdump_bpf.o'
ERROR: Can't load eBPF object: No such file or directory(-2)
```

The real problem is that xdpdump identify the function name as ```xdp_tailgrow_fu()```, while the real function name is ```xdp_tailgrow_func4()```. Knowing the kernel, this is because the kernel only stores 15 chars of the original ELF function name (details: in kernel ```struct bpf_prog_aux->name[BPF_OBJ_NAME_LEN]```)

```
$ sudo ./xdpdump -D
if_index  if_name           XDP program entry function
--------  ----------------  --------------------------------------------------
1         lo                <No XDP program loaded!>
[... cut ...]
7         i40e2             <No XDP program loaded!>
8         mlx5p1            xdp_tailgrow_fu()
9         mlx5p2            <No XDP program loaded!>
```

**Workaround:**
Keeping function names (not ```SEC``` section names) below 15 chars in the BPF-prog is a workaround to the issue.

**Help/hints to solve:**
I think the xdpdump program can find the associated full-length name via BTF info. The ```struct bpf_prog_aux``` have a (```const char *```) member named ```attach_func_name```, that I think we can extract somehow (according to comment this is "function name for valid attach_btf_id").

netoptimizer:
Issues with xdpdump can be debugged via supplying ```-vv``` for more verbose.
I provided full output in pastebin here: https://pastebin.com/KzSCvMqM

Important line is:
```
libbpf: xdp_tailgrow_fu is not found in prog's BTF
```

tohojo:
> **Help/hints to solve:**

> I think the xdpdump program can find the associated full-length name
> via BTF info. The ```struct bpf_prog_aux``` have a (```const char
> *```) member named ```attach_func_name```, that I think we can extract
> somehow (according to comment this is "function name for valid
> attach_btf_id").

We don't even need to look at bpf_prog_aux - just get the BTF from the
program from the kernel and iterate that. Functions are stored directly
in the BTF info, e.g. as [7] here:

$ bpftool btf dump file xdp_pass_kern.o
[...]

PR #24, opened 2020-05-05
Rework libxdp API and use it in all utilities

tohojo:
This is a rework of the libxdp API to avoid exposing the multiprog and
dispatcher internals, and instead have xdp_program-based attach and detach
functions, so we can have some freedom to change the internals without having to
touch the API.

This also reworks the lib/util functions to build on top of libxdp, and ports
all the command-line utilities to use it.

While I have runtime-tested xdp-loader, the other utilities are only compile
tested.

### DIFF
--- a/lib/libxdp/libxdp.c
+++ b/lib/libxdp/libxdp.c
@@ -1829,18 +1829,10 @@ enum xdp_attach_mode xdp_multiprog__attach_mode(const struct xdp_multiprog *mp)
 	return mp->attach_mode;
 }
 
-uint32_t xdp_multiprog__dispatcher_id(const struct xdp_multiprog *mp)
+struct xdp_program *xdp_multiprog__dispatcher(const struct xdp_multiprog *mp)
 {
 	if (!mp)
-		return 0;
+		return NULL;
 
-	return xdp_program__id(mp->dispatcher);
-}
-
-const unsigned char *xdp_multiprog__dispatcher_tag(const struct xdp_multiprog *mp)
-{
-	if (!mp)
-		return 0;
-
-	return xdp_program__tag(mp->dispatcher);
+	return mp->dispatcher;
 }
--- a/lib/libxdp/libxdp.c
+++ b/lib/libxdp/libxdp.c
@@ -520,7 +520,7 @@ static struct xdp_program *xdp_program__new()
 	return xdp_prog;
 }
 
-void xdp_program__free(struct xdp_program *xdp_prog)
+void xdp_program__close(struct xdp_program *xdp_prog)
 {
 	if (!xdp_prog)
 		return;
@@ -595,7 +595,7 @@ struct xdp_program *xdp_program__from_bpf_obj(struct bpf_object *obj,
 
 	return xdp_prog;
 err:
-	xdp_program__free(xdp_prog);
+	xdp_program__close(xdp_prog);
 	return ERR_PTR(err);
 }
 
@@ -627,7 +627,7 @@ struct xdp_program *xdp_program__open_file(const char *filename,
 
 	return xdp_prog;
 err:
-	xdp_program__free(xdp_prog);
+	xdp_program__close(xdp_prog);
 	bpf_object__close(obj);
 	return ERR_PTR(err);
 }
@@ -812,17 +812,17 @@ static struct xdp_program *xdp_program__clone(struct xdp_program *prog)
 }
 
 
-void xdp_multiprog__free(struct xdp_multiprog *mp)
+void xdp_multiprog__close(struct xdp_multiprog *mp)
 {
 	struct xdp_program *p, *next = NULL;
 
 	if (!mp)
 		return;
 
-	xdp_program__free(mp->dispatcher);
+	xdp_program__close(mp->dispatcher);
 	for (p = mp->first_prog; p; p = next) {
 		next = p->next;
-		xdp_program__free(p);
+		xdp_program__close(p);
 	}
 }
 
@@ -977,7 +977,7 @@ err:
 	prog = mp->first_prog;
 	while (prog) {
 		p = prog->next;
-		xdp_program__free(prog);
+		xdp_program__close(prog);
 		prog = p;
 	}
 	mp->first_prog = NULL;
@@ -1079,7 +1079,7 @@ out:
 	return err;
 
 err:
-	xdp_program__free(prog);
+	xdp_program__close(prog);
 	goto out;
 }
 
@@ -1225,7 +1225,7 @@ static int xdp_multiprog__link_prog(struct xdp_multiprog *mp,
 	return 0;
 
 err_free:
-	xdp_program__free(new_prog);
+	xdp_program__close(new_prog);
 err:
 	return err;
 }
@@ -1300,7 +1300,7 @@ struct xdp_multiprog *xdp_multiprog__generate(struct xdp_program **progs,
 	return mp;
 
 err:
-	xdp_multiprog__free(mp);
+	xdp_multiprog__close(mp);
 	return ERR_PTR(err);
 }
 
--- a/lib/libxdp/libxdp.c
+++ b/lib/libxdp/libxdp.c
@@ -336,6 +336,14 @@ uint32_t xdp_program__id(const struct xdp_program *xdp_prog)
 	return xdp_prog->prog_id;
 }
 
+int xdp_program__fd(const struct xdp_program *xdp_prog)
+{
+	if (!xdp_prog)
+		return -1;
+
+	return xdp_prog->prog_fd;
+}
+
 int xdp_program__print_chain_call_actions(const struct xdp_program *prog,
 					  char *buf,
 					  size_t buf_len)
--- a/lib/libxdp/libxdp.c
+++ b/lib/libxdp/libxdp.c
@@ -233,7 +233,7 @@ static int xdp_lock_release(int lock_fd)
 	return err;
 }
 
-struct btf *xdp_program__btf(struct xdp_program *xdp_prog)
+const struct btf *xdp_program__btf(struct xdp_program *xdp_prog)
 {
 	if (!xdp_prog)
 		return NULL;
@@ -273,17 +273,18 @@ out:
 	return ret;
 }
 
-void xdp_program__set_chain_call_enabled(struct xdp_program *prog,
-					 unsigned int action, bool enabled)
+int xdp_program__set_chain_call_enabled(struct xdp_program *prog,
+					unsigned int action, bool enabled)
 {
-	if (!prog)
-		return;
+	if (!prog || prog->prog_fd)
+		return -EINVAL;
 
-	/* FIXME: Should this also update the BTF info? */
 	if (enabled)
 		prog->chain_call_actions |= (1<<action);
 	else
 		prog->chain_call_actions &= ~(1<<action);
+
+	return 0;
 }
 
 bool xdp_program__chain_call_enabled(const struct xdp_program *prog,
@@ -303,13 +304,13 @@ unsigned int xdp_program__run_prio(const struct xdp_program *prog)
 	return prog->run_prio;
 }
 
-void xdp_program__set_run_prio(struct xdp_program *prog, unsigned int run_prio)
+int xdp_program__set_run_prio(struct xdp_program *prog, unsigned int run_prio)
 {
-	if (!prog)
-		return;
+	if (!prog || prog->prog_fd)
+		return -EINVAL;
 
-	/* FIXME: Should this also update the BTF info? */
 	prog->run_prio = run_prio;
+	return 0;
 }
 
 const char *xdp_program__name(const struct xdp_program *prog)
@@ -587,7 +588,7 @@ static const struct btf_type *btf_get_section_var(const struct btf *btf,
  */
 static int xdp_program__parse_btf(struct xdp_program *xdp_prog)
 {
-	struct btf *btf = xdp_program__btf(xdp_prog);
+	const struct btf *btf = xdp_program__btf(xdp_prog);
 	const struct btf_type *def, *sec;
 	const struct btf_member *m;
 	char struct_name[100];
--- a/lib/libxdp/libxdp.c
+++ b/lib/libxdp/libxdp.c
@@ -1863,7 +1863,7 @@ static int xdp_multiprog__attach(struct xdp_multiprog *old_mp,
 		xdp_flags |= XDP_FLAGS_DRV_MODE;
 		break;
 	case XDP_MODE_HW:
-		return -EOPNOTSUPP;
+		xdp_flags |= XDP_FLAGS_HW_MODE;
 		break;
 	case XDP_MODE_UNSPEC:
 		break;
@@ -1880,7 +1880,7 @@ static int xdp_multiprog__attach(struct xdp_multiprog *old_mp,
 			pr_debug("XDP already loaded on device\n");
 			break;
 		case EOPNOTSUPP:
-			pr_debug("Native XDP not supported; try using SKB mode\n");
+			pr_debug("XDP mode not supported; try using SKB mode\n");
 			break;
 		default:
 			break;
--- a/lib/libxdp/libxdp.c
+++ b/lib/libxdp/libxdp.c
@@ -791,6 +791,21 @@ err:
 	return ERR_PTR(err);
 }
 
+static bool try_bpf_file(char *buf, size_t buf_size,
+			 char *path, const char *progname)
+{
+	struct stat sb = {};
+
+	if (try_snprintf(buf, buf_size, "%s/%s", path, progname))
+		return false;
+
+	pr_debug("Looking for '%s'\n", buf);
+	if (stat(buf, &sb))
+		return false;
+
+	return true;
+}
+
 static int find_bpf_file(char *buf, size_t buf_size, const char *progname)
 {
 	static char *bpf_obj_paths[] = {
@@ -800,21 +815,15 @@ static int find_bpf_file(char *buf, size_t buf_size, const char *progname)
 		BPF_OBJECT_PATH,
 		NULL
 	};
-	struct stat sb = {};
-	char **path;
-	int err;
+	char *path, **p;
 
-	for (path = bpf_obj_paths; *path; path++) {
-		err = try_snprintf(buf, buf_size, "%s/%s", *path, progname);
-		if (err)
-			return err;
-
-		pr_debug("Looking for '%s'\n", buf);
-		err = stat(buf, &sb);
-		if (err)
-			continue;
-
+	path = secure_getenv(XDP_OBJECT_ENVVAR);
+	if (path && try_bpf_file(buf, buf_size, path, progname)) {
 		return 0;
+	} else if (!path) {
+		for (p = bpf_obj_paths; *p; p++)
+			if (try_bpf_file(buf, buf_size, *p, progname))
+				return 0;
 	}
 
 	pr_warn("Couldn't find a BPF file with name %s\n", progname);
--- a/lib/libxdp/libxdp.c
+++ b/lib/libxdp/libxdp.c
@@ -941,7 +941,8 @@ struct xdp_program *xdp_program__from_pin(const char *pin_path)
 	fd = bpf_obj_get(pin_path);
 	if (fd < 0) {
 		err = -errno;
-		pr_warn("couldn't get program fd: %s", strerror(-err));
+		pr_warn("couldn't get program fd from %s: %s",
+			pin_path, strerror(-err));
 		return ERR_PTR(err);
 	}
 
--- a/lib/libxdp/libxdp.c
+++ b/lib/libxdp/libxdp.c
@@ -694,6 +694,8 @@ void xdp_program__close(struct xdp_program *xdp_prog)
 		else if (xdp_prog->btf)
 			btf__free(xdp_prog->btf);
 	}
+
+	free(xdp_prog);
 }
 
 static int xdp_program__fill_from_obj(struct xdp_program *xdp_prog,
@@ -1182,6 +1184,8 @@ void xdp_multiprog__close(struct xdp_multiprog *mp)
 		next = p->next;
 		xdp_program__close(p);
 	}
+
+	free(mp);
 }
 
 static int xdp_multiprog__main_fd(struct xdp_multiprog *mp)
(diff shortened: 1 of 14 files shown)


tohojo:
@chaudron could you please take a look at the xdpdump patch? And if you have comments on the libxdp API in general, I'm all ears as well; that last bit goes for @netoptimizer as well! :)

xdp-dump/xdpdump.c
```diff
@@ -416,94 +417,30 @@ static bool capture_on_legacy_interface(struct dumpopt *cfg)
...
 
-		pr_warn("Please use the -p option to pick the correct one.\n");
-		func_name_ptr = NULL;
-	} else {
-		func_name_ptr = strdup(func_name_ptr);
-	}
-
-	btf__free(btf);
-	return func_name_ptr;
+	return NULL;
```
chaudron, line 434: Will schedule a full review of these changes for early next week, but in this case we should still have something like thise: 

```
pr_warn("ERROR: Can't find function, %s, in BTF information!\n"
```

xdp-dump/xdpdump.c
```diff
@@ -756,21 +692,25 @@ static bool list_interfaces(struct dumpopt *cfg)
...
+			struct xdp_program *prog = NULL;
+
 			printf("%-8d  %-16.16s  %s()\n",
-			       iface.ifindex, iface.ifname,
-			       info.name);
+			       idx->if_index, idx->if_name,
+			       xdp_program__name(xdp_multiprog__dispatcher(mp)));
+
+			while ((prog = xdp_multiprog__next_prog(prog, mp)))
+				printf("%-25s %s()\n", "=>",
```
chaudron, line 710: The main reason for adding the guess_full_main_function() function was to solve the long program name issue, issue #17. But all of this is undone :) Even this command is not showing the long program loaded:

```
[root@xdp-dev xdp-dump]# ./xdpdump -D
if_index  if_name           XDP program entry function
--------  ----------------  --------------------------------------------------
1         lo                <No XDP program loaded!>
2         eth0              <No XDP program loaded!>
3         eth1              <No XDP program loaded!>
```

Old code shows this:

```
# ./xdpdump -D
if_index  if_name           XDP program entry function
--------  ----------------  --------------------------------------------------
1         lo                <No XDP program loaded!>
2         eth0              xdp_test_prog_w()
3         eth1              <No XDP program loaded!>
```

And the real long name is:

```
SEC("xdp_test_prog_long")
int xdp_test_prog_with_a_long_name(struct xdp_md *ctx)
{
    bpf_debug("PASS[1]: prog %u\n", ctx->ingress_ifindex);
    return XDP_PASS;
}
```

Also do not forget to take care of the case where there is more than one program matching, see the other error you removed in find_target_program() above.

I've attached a simeple ebpf program below that can be used two simulate the use cases:
[test_long_func_name.txt](https://github.com/xdp-project/xdp-tools/files/4585973/test_long_func_name.txt)

xdp-dump/xdpdump.c
tohojo: What? Telling the user what went wrong instead of just silently erroring out and leaving them guessing? Luxury, I say! ;)

xdp-dump/xdpdump.c
tohojo: Yes, I do realise I undid all your careful matching :)
As it says in the commit log, though, it's not completely accidental: I decided to sidestep the whole issue instead by just having libxdp enforce a max length. This is not ideal, of course, but I think it's something we ought to solve at the libxdp level.

My current idea is that since libxdp will pin all component programs anyway, we can put more information into the pin filenames. Such as the BTF ID of the function, which libxdp can then read back out when generating a struct xdp_program from a kernel fd, so we can make sure we always have the full name in xdp_program__name(), and we won't have to guess.

As for the other issue with more than one program name matching, that is actually worse than what you say: With XDP multiprogs we can have more than one sub-program using the same function name, since they are completely independent BPF objects. This means that program name is not guaranteed to be a valid disambiguator between functions; so I think xdpdump needs to switch to using IDs instead...

chaudron: > Yes, I do realise I undid all your careful matching :)
> As it says in the commit log, though, it's not completely accidental: I decided to sidestep the whole issue instead by just having libxdp enforce a max length. This is not ideal, of course, but I think it's something we ought to solve at the libxdp level.

Well, this is a problem, as you assume we will always have multi prog. The issues I raise are all in the non-multiprog setup. I.e. someone just loading a program. Which xdpdump must also support.

> 
> My current idea is that since libxdp will pin all component programs anyway, we can put more information into the pin filenames. Such as the BTF ID of the function, which libxdp can then read back out when generating a struct xdp_program from a kernel fd, so we can make sure we always have the full name in xdp_program__name(), and we won't have to guess.

Same as above, this is only in the non-mulitprog case.
 
> As for the other issue with more than one program name matching, that is actually worse than what you say: With XDP multiprogs we can have more than one sub-program using the same function name, since they are completely independent BPF objects. This means that program name is not guaranteed to be a valid disambiguator between functions; so I think xdpdump needs to switch to using IDs instead...

My idea was still to do our best guessing, i.e. match first xchars on program entries on various programs, and then walk that specific BPF tables. Because we use an exact match this is still fine. The only problem is when multiple programs use the same first x chars, this we can probably solve by allowing a program NR/index instead of a name in the --program-names option. So something like `--program-names=1,5` to trace the first and 5th program.


I think for now, as xdpdump does not yet support multi-programs, leave the code as is, and I fix it up when we add multi-program support as this area needs to change anyway. I think your patchset should still work with the old code.

tohojo:

> @chaudron commented on this pull request.
>
>
>
>>  			printf("%-8d  %-16.16s  %s()\n",
> -			       iface.ifindex, iface.ifname,
> -			       info.name);
> +			       idx->if_index, idx->if_name,
> +			       xdp_program__name(xdp_multiprog__dispatcher(mp)));
> +
> +			while ((prog = xdp_multiprog__next_prog(prog, mp)))
> +				printf("%-25s %s()\n", "=>",
>
>> Yes, I do realise I undid all your careful matching :)
>> As it says in the commit log, though, it's not completely accidental: I decided to sidestep the whole issue instead by just having libxdp enforce a max length. This is not ideal, of course, but I think it's something we ought to solve at the libxdp level.
>
> Well, this is a problem, as you assume we will always have multi prog.
> The issues I raise are all in the non-multiprog setup. I.e. someone
> just loading a program. Which xdpdump must also support.

Not if we assume all programs are loaded through libxdp ;) Which is the
fantasy world I'm currently living in...

>> My current idea is that since libxdp will pin all component programs
>> anyway, we can put more information into the pin filenames. Such as
>> the BTF ID of the function, which libxdp can then read back out when
>> generating a struct xdp_program from a kernel fd, so we can make sure
>> we always have the full name in xdp_program__name(), and we won't
>> have to guess.
>
> Same as above, this is only in the non-mulitprog case.
>  
>> As for the other issue with more than one program name matching, that
>> is actually worse than what you say: With XDP multiprogs we can have
>> more than one sub-program using the same function name, since they
>> are completely independent BPF objects. This means that program name
>> is not guaranteed to be a valid disambiguator between functions; so I
>> think xdpdump needs to switch to using IDs instead...
>
> My idea was still to do our best guessing, i.e. match first xchars on
> program entries on various programs, and then walk that specific BPF
> tables. Because we use an exact match this is still fine. The only
> problem is when multiple programs use the same first x chars, this we
> can probably solve by allowing a program NR/index instead of a name in
> the --program-names option. So something like `--program-names=1,5` to
> trace the first and 5th program.

Hmm, I guess we could support the case where there is no ambiguity. But
are program names really better than IDs in the first place?

> I think for now, as xdpdump does not yet support multi-programs, leave
> the code as is, and I fix it up when we add multi-program support as
> this area needs to change anyway. I think your patchset should still
> work with the old code.

Well with this patch xdpdump *does* support multi-programs. Just only
one of them at a time :)

chaudron:
> Not if we assume all programs are loaded through libxdp ;) Which is the fantasy world I'm currently living in...

I would like xdpdump to support both, as it might not only be used on RHEL, but in other areas...

> Hmm, I guess we could support the case where there is no ambiguity. But are program names really better than IDs in the first place?

Yes, I think so, as the IDs are not consequent, they might not be the same on every system. So from a user perspective, it might be easier to ask for `xdpdump -i eth0 -p firewall_entry` than ask them to do xdpdump -D, get the id for firewall_entry and use it with -p.

> I think for now, as xdpdump does not yet support multi-programs, leave the code as is, and I 
fix it up when we add multi-program support as this area needs to change anyway. I think your patchset should still work with the old code.
> Well with this patch xdpdump *does* support multi-programs. Just only one of them at a time :)

Did you try? As I could not attach with my previous code...

tohojo:

>> Not if we assume all programs are loaded through libxdp ;) Which is
>> the fantasy world I'm currently living in...
>
> I would like xdpdump to support both, as it might not only be used on
> RHEL, but in other areas...

Yeah, sure. For this I was thinking we could have the libxdp
xdp_multiprog__get_from_ifindex() generate a "dummy" multiprog with just
a single item underneath it if it encounters a program that's not a
dispatcher. So that a consumer of the library API doesn't have to
care... And if we make libxdp walk the BTF to expand the full program
names, all xdpdump has to care about is the case when a single BPF
program has ambiguous (>15 chars) function names...

>> Hmm, I guess we could support the case where there is no ambiguity. But are program names really better than IDs in the first place?
>
> Yes, I think so, as the IDs are not consequent, they might not be the
> same on every system. So from a user perspective, it might be easier
> to ask for `xdpdump -i eth0 -p firewall_entry` than ask them to do
> xdpdump -D, get the id for firewall_entry and use it with -p.

Ah, right, good point; the names stay consistent for an application,
hadn't thought of that :)

>> I think for now, as xdpdump does not yet support multi-programs,
>> leave the code as is, and I fix it up when we add multi-program
>> support as this area needs to change anyway. I think your patchset
>> should still work with the old code. Well with this patch xdpdump
>> *does* support multi-programs. Just only one of them at a time :)
>
> Did you try? As I could not attach with my previous code...

Heh, didn't say it was working, just that the support was there ;)
But now that you mention it, yeah, there was that bug where the verifier
rejects it...

tohojo:
Right, brought back the function name checking for xdpdump (still only compile-tested, though). Also added function name expansion when building xdp_program objects, but this can still be ambiguous if there are multiple programs with the same prefix, so had to basically re-implement the matching in xdpdump... Suggestions for how to avoid that welcome :)

lib/libxdp/libxdp.c
```diff
@@ -240,6 +240,30 @@ static struct btf *xdp_program__btf(struct xdp_program *xdp_prog)
 	return xdp_prog->btf;
 }
 
+bool xdp_program__is_attached(const struct xdp_program *xdp_prog, int ifindex)
+{
+	struct xdp_program *prog = NULL;
+	struct xdp_multiprog *mp;
+	bool ret = false;
+
+	if (!xdp_prog->prog_id)
```
chaudron, line 250: Maybe: `if (!xdp_prog || !xdp_prog->prog_id)`

lib/libxdp/libxdp.c
```diff
@@ -861,6 +903,14 @@ static int cmp_xdp_programs(const void *_a, const void *_b)
 	return 0;
 }
 
+int xdp_program__pin(struct xdp_program *prog, const char *pin_path)
+{
+	if (!prog->prog_fd)
```
chaudron, line 991: Add `!prog || `

lib/util/util.c
```diff
@@ -229,180 +118,62 @@ int make_dir_subdir(const char *parent, const char *dir)
...
-		 prog_name ?: bpf_program__title(prog, false),
-		 iface->ifname, mode == XDP_MODE_SKB ? " in skb mode" : "");
+	char pin_path[PATH_MAX];
+	int err = 0;
 
 	if (!pin_root_path)
-		return 0;
+		goto load;
 
 	err = make_dir_subdir(pin_root_path, "programs");
```
chaudron, line 174: Should we add an error message in this case? It might be useful if for some reason we can't create the pinned directory it gives an indication which directory and where we were trying to add it.

lib/util/util.c
```diff
@@ -470,23 +243,26 @@ int get_pinned_program(const struct iface *iface, const char *pin_root_path,
 			continue;
 		}
 
-		if (!program_is_loaded(iface->ifindex, pin_path, mode, info)) {
-			ret = -ENOENT;
+		prog = xdp_program__from_pin(pin_path);
+		if (IS_ERR_OR_NULL(prog) ||
+		    !xdp_program__is_attached(prog, iface->ifindex)) {
+			ret = IS_ERR(prog) ? PTR_ERR(prog) : -ENOENT;
 			pr_debug("Program %s no longer loaded on %s\n",
```
chaudron, line 294: Maybe add ret to the pr_debug message as it can be overwritten by unlink.

lib/libxdp/libxdp.c
```diff
@@ -1393,20 +1408,23 @@ static int xdp_multiprog__fill_from_fd(struct xdp_multiprog *mp, int prog_fd)
...
-	if (err)
-		goto err;
+	if (!mp->is_legacy) {
+		err = xdp_multiprog__link_pinned_progs(mp);
+		if (err)
+			goto err;
+	}
 
 	mp->is_loaded = true;
 	pr_debug("Found multiprog with id %d and %zu component progs\n",
```
chaudron, line 1426: Should we do something like: 

```
pr_debug("Found %s with id %d and %zu component progs\n",
         mp->is_legacy ? "legacy program" : "multiprog",

```

lib/libxdp/libxdp.c
```diff
@@ -1668,7 +1686,7 @@ static int xdp_multiprog__pin(struct xdp_multiprog *mp)
 		return lock_fd;
 
 	pr_debug("Pinning multiprog fd %d beneath %s\n",
-		 mp->dispatcher->prog_fd, pin_path);
+		 mp->main_prog->prog_fd, pin_path);
```
chaudron, line 1689: same here on `multiprog` string

xdp-dump/xdpdump.c
```diff
@@ -416,27 +416,119 @@ static bool capture_on_legacy_interface(struct dumpopt *cfg)
...
+	}
+
+	if (matches == 1)
+		*found_name = btf__name_by_offset(btf, match->name_off);
+
+	return matches;
+}
+
 /*****************************************************************************
  * find_target_program()
```
chaudron, line 468: Need updated program name :)

xdp-dump/xdpdump.c
```diff
@@ -416,27 +416,119 @@ static bool capture_on_legacy_interface(struct dumpopt *cfg)
 	return rc;
 }
 
+static size_t find_func_matches(const struct btf *btf,
```
chaudron, line 419: Missing function marker

```
/*****************************************************************************
 * find_func_matches()
 *****************************************************************************/
```

lib/libxdp/libxdp.c
```diff
@@ -632,6 +664,51 @@ struct xdp_program *xdp_program__open_file(const char *filename,
 	return ERR_PTR(err);
 }
 
+static int find_bpf_file(char *buf, size_t buf_size, const char *progname)
+{
+	static char *bpf_obj_paths[] = {
+#ifdef DEBUG
+		".",
+#endif
+		BPF_OBJECT_PATH,
```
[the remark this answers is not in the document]
tohojo: Yeah, thought about that; couldn't decide, so left it out. But probably a good idea, so will add this!

lib/libxdp/libxdp.c
```diff
@@ -69,6 +70,32 @@ static const char *xdp_action_names[] = {
...
+{
+	return libbpf_get_error(ptr);
+}
+
+int libxdp_strerror(int err, char *buf, size_t size)
+{
+	return libbpf_strerror(err, buf, size);
+}
+
+static int check_snprintf(char *buf, size_t buf_len, const char *format, ...)
```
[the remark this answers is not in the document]
tohojo: Heh, right, can fix that :)

xdp-dump/xdpdump.c
```diff
@@ -416,27 +416,119 @@ static bool capture_on_legacy_interface(struct dumpopt *cfg)
...
+
+		name = btf__name_by_offset(btf, t->name_off);
+		if (!strncmp(name, func_name, len)) {
+			pr_debug("Found func %s matching %s\n",
+				 name, func_name);
+
+			if (print)
+				pr_warn("  %s\n", name);
+
+			if (strlen(name) == len) {
```
chaudron, line 450: This should be removed 

```
			/* if (strlen(name) == len) { */
			/* 	*found_name = name; */
			/* 	return 1; /\* exact match *\/ */
			/* } */
```

As it's failing the test case where the exact name is not the function loaded.

xdp-dump/xdpdump.c
```diff
@@ -756,21 +692,25 @@ static bool list_interfaces(struct dumpopt *cfg)
...
 		} else {
+			struct xdp_program *prog = NULL;
+
 			printf("%-8d  %-16.16s  %s()\n",
-			       iface.ifindex, iface.ifname,
-			       info.name);
+			       idx->if_index, idx->if_name,
+			       xdp_program__name(xdp_multiprog__dispatcher(mp)));
+
+			while ((prog = xdp_multiprog__next_prog(prog, mp)))
```
chaudron, line 806: This format look odd:

```
# ./xdpdump -i eth0 -D
if_index  if_name           XDP program entry function
--------  ----------------  --------------------------------------------------
1         lo                <No XDP program loaded!>
2         eth0              xdp_main()
=>                        xdp_test_II()
=>                        xdp_test_I()
3         eth1              <No XDP program loaded!>
```

Please change to:

```
                        while ((prog = xdp_multiprog__next_prog(prog, mp)))
-                               printf("%-25s %s()\n", "=>",
+                               printf("%-29s %s()\n", "",
                                       xdp_program__name(prog));
```

So it looks like before:

```
# make && ./xdpdump -i eth0 -D
make: Nothing to be done for 'all'.
if_index  if_name           XDP program entry function
--------  ----------------  --------------------------------------------------
1         lo                <No XDP program loaded!>
2         eth0              xdp_main()
                              xdp_test_II()
                              xdp_test_I()
3         eth1              <No XDP program loaded!>
```

xdp-dump/xdpdump.c
tohojo: Not sure I understand your comment; what is failing?

chaudron: If I have three functions in my BTF (file) and one is an exact match but not the one actually being used it assumes this is the one needed. Which might not be the case, its better to fail and let the user pick the correct one and not assume the exact match is the correct one.

tohojo: But how is a user supposed to specify the right candidate if an exact match gets rejected when it's also a prefix of another function?

I.e., say there's `funcname1()` and `funcname1_more()` and the user **wants** to use `funcname1()`, if we don't have this 'exact match' short-circuiting that won't be possible, unless I'm missing something?

chaudron: You are right, that test case I did before I made the change :(

The exact match should only be valid for the -p option, and probably also when the length is less than 15 chars as than no collision can happen. Think I had that in the original code.

tohojo: Yeah, you did. The reason I removed it was that the name is now expanded by libxdp on load, so the truncated name will only survive if that initial expansion is ambiguous. Hmm, I guess we could have if (-p set || len != 15) to automatically match both longer and shorter names in one go, then?

lib/libxdp/libxdp.c
```diff
@@ -1551,60 +1705,59 @@ int xdp_multiprog__attach(struct xdp_multiprog *mp,
...
 	return err;
 }
 
+int xdp_multiprog__detach(struct xdp_multiprog *mp, int ifindex)
+{
+	int err;
+
+	if (!mp || !mp->is_loaded)
+		return -EINVAL;
+
```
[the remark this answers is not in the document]
tohojo: Yeah, good question. I think maybe we should just remove the mode parameter from the detach functions? Since the API always asks to detach a specific program, should it matter which mode it is attached in?

xdp-dump/xdpdump.c
tohojo: Just out of curiosity, is there a particular reason you add these headings? I.e., does your editor do something specific with them, or is it purely aesthetics?

chaudron: It's my OCD kicking in, can't stand long files without a clear function/data boundary.

Thanks for making all the review changes, I guess it kept you buys based on the number of commits :)

tohojo: Right fair enough. And yeah, this has kept me busy for a bit, but getting close now! Thanks for reviewing :)

lib/util/util.c
```diff
@@ -723,11 +723,12 @@ int prog_lock_get(const char *progname)
...
+			len = read(fd, buf, sizeof(buf)-1);
 			close(fd);
-			buf[len] = '\0';
-
-			pid = strtoul(buf, NULL, 10);
+			if (len > 0) {
+				buf[len] = '\0';
+				pid = strtoul(buf, NULL, 10);
+			}
 			if (!pid) {
```
chaudron, line 730: Bit of a nity comment, but why not check errno here also, as ret on error could be ULONG_MAX.
So `if (!pid || errno)`

lib/util/util.c
```diff
@@ -481,13 +481,11 @@ static int bpf_mnt_check_target(const char *target)
 	struct stat sb = {};
```
chaudron, line 481: This should have given you an unused variable error.

xdp-dump/xdpdump.c
```diff
@@ -813,7 +813,7 @@ static bool list_interfaces(struct dumpopt *cfg)
 			       xdp_program__name(xdp_multiprog__main_prog(mp)));
 
 			while ((prog = xdp_multiprog__next_prog(prog, mp)))
-				printf("%-25s %s()\n", "=>",
+				printf("%-29s %s()\n", "=>",
```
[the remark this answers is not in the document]
tohojo: Oh, right. That was not on purpose, I'm fine with removing it :)

lib/util/util.c
tohojo: ACK, can do

lib/util/util.c
tohojo: And indeed it does; guess maybe I forgot to compile-test this before pushing? Will fix

PR #43, opened 2020-06-03
xdp-dump: add first set of selftests

chaudron:

### DIFF
--- a/xdp-dump/test-xdpdump.sh
+++ b/xdp-dump/test-xdpdump.sh
@@ -1,0 +1,254 @@
+#
+# Test scrip to do basic xdpdump checks
+#
+# shellcheck disable=2039
+#
+ALL_TESTS="test_help test_interfaces test_capt_pcap test_capt_term test_exitentry test_snap test_none_xdp"
+
+XDPDUMP=./xdpdump
+XDP_LOADER=../xdp-loader/xdp-loader
+
+RESULT=""
+
+print_result()
+{
+    echo "$RESULT"
+    if [ -n "$1" ]; then
+	echo "ERROR: $1"
+    fi
+}
+
+start_background()
+{
+    local TMP_FILE="/tmp/${NS}_PID_$$_$RANDOM"
+    eval "${1} >& ${TMP_FILE} &"
+    local PID=$!
+    sleep 1 # Wait to make sure the command is executed in the background
+
+    mv "$TMP_FILE" "/tmp/${NS}_PID_${PID}" >& /dev/null
+
+    echo "$PID"
+}
+
+stop_background()
+{
+    local OUTPUT_FILE="/tmp/${NS}_PID_${1}"
+    kill -SIGINT "$1"
+    sleep 1 # Wait to make sure the buffer is flushed after the shutdown
+    cat "$OUTPUT_FILE"
+    rm "$OUTPUT_FILE" >& /dev/null
+}
+
+test_help()
+{
+    local XDPDUMP_HELP_TEXT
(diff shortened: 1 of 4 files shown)


xdp-dump/test-xdpdump.sh
```diff
@@ -0,0 +1,252 @@
...
+#
+# shellcheck disable=2039
+#
+XDPDUMP=./xdpdump
+XDP_LOADER=../xdp-loader/xdp-loader
+ALL_TESTS="test_help test_interfaces test_capt_pcap test_capt_term test_exitentry test_snap test_none_xdp"
+
+print_result()
+{
+    echo "$RESULT"
```
tohojo, line 12: Please either declare this as a global variable, or just pass it in as a parameter?

xdp-dump/test-xdpdump.sh
```diff
@@ -0,0 +1,252 @@
...
+    local OUTPUT_FILE="/tmp/${NS}_PID_${1}"
+    kill -SIGINT "$1"
+    sleep 1 # Wait to make sure the buffer is flushed after the shutdown
+    cat "$OUTPUT_FILE"
+    rm "$OUTPUT_FILE" >& /dev/null
+}
+
+test_help()
+{
+    local RESULT=""
```
tohojo, line 41: This breaks with the expectation of a global variable above, doesn't it?

chaudron: Yes I forgot to cleanup this one :( RESULT should be global, will make the change

xdp-dump/test-xdpdump.sh
```diff
@@ -0,0 +1,254 @@
...
+    if ! [[ $RESULT =~ $PASS_EXIT_REGEX ]]; then
+        print_result "IPv6 exit packet not received"
+        return 1
+    fi
+
+    $XDP_LOADER unload "$NS" --all || return 1
+    $XDP_LOADER load "$NS" "$TEST_PROG_DIR/xdp_drop.o" || return 1
+
+    PID=$(start_background "$XDPDUMP -i $NS --rx-capture=exit")
+    ping6 -W 2 -c 1 "$INSIDE_IP6" # Note that this ping will fail!!
```
tohojo, line 191: Do we really need a 2-second wait for a ping we know is going to fail? Makes the test run slower than it needs to be...

chaudron: You are right, I'll update it with
"ping6 -W 0.1 -c 1 "$INSIDE_IP6" # Note that this ping will fail!!"

chaudron:
FYI I did notice that some times ping is failing in the veth/netns, re-running the test makes it work. I could not quickly figure out what is wrong in these cases, but I suspect my vagrant fedora32 image.

tohojo:
> FYI I did notice that some times ping is failing in the veth/netns, re-running the test makes it work. I could not quickly figure out what is wrong in these cases, but I suspect my vagrant fedora32 image.

Well you're relying on 'sleep' for timing, so I guess the tests are inherently racy. I suppose we can live with that for a first pass, though :)

chaudron:
> > FYI I did notice that some times ping is failing in the veth/netns, re-running the test makes it work. I could not quickly figure out what is wrong in these cases, but I suspect my vagrant fedora32 image.
> 
> Well you're relying on 'sleep' for timing, so I guess the tests are inherently racy. I suppose we can live with that for a first pass, though :)

This is why there is an inial ping in some tests ` ping6 -W 2 -c 1 "$INSIDE_IP6" || return 1 ` as this even fails. This is just after you've setup the netns part. I'll blame my testup for now :)

PR #12, opened 2020-03-24
xdpdump: add support for PcapNG

chaudron:
Signed-off-by: Eelco Chaudron <echaudro@redhat.com>

PR #47, opened 2020-06-05
xdp-dump: add PcapNG support

chaudron:
Adding support for PcapNG, this fixes issue #12

### DIFF
--- a/lib/testing/test_runner.sh
+++ b/lib/testing/test_runner.sh
@@ -26,7 +26,7 @@ GENERATED_NAME_PREFIX="xdptest"
 ALL_TESTS=""
 TEST_PROG_DIR="${TEST_PROG_DIR:-$(dirname $0)}"
 
-NEEDED_TOOLS="ethtool ip ping tc tcpdump"
+NEEDED_TOOLS="capinfos ethtool ip ping tc tcpdump"
 MAX_NAMELEN=15
 
 # Odd return value for skipping, as only 0-255 is valid.
--- a/lib/testing/test_runner.sh
+++ b/lib/testing/test_runner.sh
@@ -26,7 +26,7 @@ GENERATED_NAME_PREFIX="xdptest"
 ALL_TESTS=""
 TEST_PROG_DIR="${TEST_PROG_DIR:-$(dirname $0)}"
 
-NEEDED_TOOLS="capinfos ethtool ip ping tc tcpdump"
+NEEDED_TOOLS="capinfos ethtool ip ping sed tc tcpdump timeout"
 MAX_NAMELEN=15
 
 # Odd return value for skipping, as only 0-255 is valid.
--- a/lib/util/params.c
+++ b/lib/util/params.c
@@ -21,6 +21,7 @@
 
 #define BUFSIZE 30
 #define FIRST_PRINTABLE 65 /* ord('A') = 65 */
+#define VERSION_SHORT_OPT 0
 
 static bool opt_needs_arg(const struct prog_option *opt)
 {
@@ -436,6 +437,7 @@ void usage(const char *prog_name, const char *doc,
 	printf("Options:\n");
 	_print_options(poptions, false);
 	printf(" -v, --verbose              Enable verbose logging (-vv: more verbose)\n");
+	printf("     --version              Display version information\n");
 	printf(" -h, --help                 Show this help\n");
 	printf("\n");
 }
@@ -444,7 +446,7 @@ static int prog_options_to_options(struct prog_option *poptions,
 				   struct option **options,
 				   char **optstring)
 {
-	int num = 0, num_cmn = 0, n_sopt = 0;
+	int num = 0, num_cmn = 0, n_sopt = VERSION_SHORT_OPT + 1;
 	struct option *new_options, *nopt;
 	struct prog_option *opt;
 	char buf[100], *c = buf;
@@ -452,12 +454,14 @@ static int prog_options_to_options(struct prog_option *poptions,
 	struct option common_opts[] = {
 	       {"help", no_argument, NULL, 'h'},
 	       {"verbose", no_argument, NULL, 'v'},
+	       {"version", no_argument, NULL, VERSION_SHORT_OPT},
 	       {}
 	};
 
 	for (nopt = common_opts; nopt->name; nopt++) {
 		num++; num_cmn++;
-		*c++ = nopt->val;
+		if (nopt->val != VERSION_SHORT_OPT)
+			*c++ = nopt->val;
 	}
 
 	FOR_EACH_OPTION(poptions, opt)
@@ -587,6 +591,10 @@ int parse_cmdline_args(int argc, char **argv,
 		case 'v':
 			increase_log_level();
 			break;
+		case VERSION_SHORT_OPT:
+			printf("%s version %s\n", prog, TOOLS_VERSION);
+			err = EXIT_FAILURE;
+			goto out;
 		default:
 			if (set_opt(cfg, poptions, opt, optarg)) {
 				usage(prog, doc, poptions, full_help);
--- a/lib/util/util.mk
+++ b/lib/util/util.mk
@@ -1,1 +1,1 @@
-UTIL_OBJS := params.o logging.o util.o stats.o
+UTIL_OBJS := params.o logging.o util.o stats.o xpcapng.o
--- a/lib/util/xpcapng.c
+++ b/lib/util/xpcapng.c
@@ -1,0 +1,634 @@
+// SPDX-License-Identifier: GPL-2.0
+/*
+ *  Description:
+ *      Simple PcapNG library developed from scratch as no library existed that
+ *      met the requirements for xdpdump. It can also be used by other XDP
+ *      applications that would like to capture packets for debugging purposes.
+ */
+
+/*****************************************************************************
+ * Include files
+ *****************************************************************************/
+#include <ctype.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <inttypes.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+
+#include <sys/uio.h>
+
+#include "xpcapng.h"
+
+/*****************************************************************************
+ * Simple roundup() macro
+ *****************************************************************************/
+#ifndef roundup
+#define roundup(x, y) (                  \
+{                                        \
+	typeof(y) __y = y;               \
+	(((x) + (__y - 1)) / __y) * __y; \
+}                                        \
+)
+#endif
+
+/*****************************************************************************
+ * pcapng_dumper structure
+ *****************************************************************************/
+struct xpcapng_dumper {
+	int      pd_fd;
+	uint32_t pd_interfaces;
+};
+
+/*****************************************************************************
+ * general pcapng block and option definitions
+ *****************************************************************************/
+enum pcapng_block_types {
+	PCAPNG_SECTION_BLOCK = 0x0A0D0D0A,
+	PCAPNG_INTERFACE_BLOCK = 1,
+	PCAPNG_PACKET_BLOCK,
+	PCAPNG_SIMPLE_PACKET_BLOCK,
+	PCAPNG_NAME_RESOLUTION_BLOCK,
+	PCAPNG_INTERFACE_STATS_BLOCK,
+	PCAPNG_ENHANCED_PACKET_BLOCK
+};
+
+struct pcapng_option {
+	uint16_t po_type;
+	uint16_t po_length;
+	uint8_t  po_data[];
+} __attribute__((__packed__));
+
+enum pcapng_opt {
+	PCAPNG_OPT_END	= 0,
+	PCAPNG_OPT_COMMENT = 1,
+	PCAPNG_OPT_CUSTOME_2988 = 2988,
+	PCAPNG_OPT_CUSTOME_2989 = 2989,
+	PCAPNG_OPT_CUSTOME_19372 = 19372,
+	PCAPNG_OPT_CUSTOME_19373 = 29373
+};
+
+/*****************************************************************************
+ * pcapng section header block definitions
+ *****************************************************************************/
+struct pcapng_section_header_block {
+	uint32_t shb_block_type;
+	uint32_t shb_block_length;
+	uint32_t shb_byte_order_magic;
+	uint16_t shb_major_version;
+	uint16_t shb_minor_version;
+	uint64_t shb_section_length;
+	uint8_t  shb_options[];
+	/* The options are followed by another:
+	 *   uint32_t shb_block_length;
+	 */
+} __attribute__((__packed__));
+
+#define PCAPNG_BYTE_ORDER_MAGIC 0x1A2B3C4D
+#define PCAPNG_MAJOR_VERSION    1
+#define PCAPNG_MINOR_VERSION    0
+
+enum pcapng_opt_shb {
+	PCAPNG_OPT_SHB_HARDWARE = 2,
+	PCAPNG_OPT_SHB_OS,
+	PCAPNG_OPT_SHB_USERAPPL
+};
+
+/*****************************************************************************
+ * pcapng interface description block definitions
+ *****************************************************************************/
+struct pcapng_interface_description_block {
+	uint32_t idb_block_type;
+	uint32_t idb_block_length;
+	uint16_t idb_link_type;
+	uint16_t idb_reserved;
+	uint32_t idb_snap_len;
+	uint8_t  idb_options[];
+	/* The options are followed by another:
+	 *   uint32_t idb_block_length;
+	 */
+} __attribute__((__packed__));
+
+enum pcapng_opt_idb {
+	PCAPNG_OPT_IDB_IF_NAME = 2,
+	PCAPNG_OPT_IDB_IF_DESCRIPTION,
+	PCAPNG_OPT_IDB_IF_IPV4_ADDR,
+	PCAPNG_OPT_IDB_IF_IPV6_ADDR,
+	PCAPNG_OPT_IDB_IF_MAC_ADDR,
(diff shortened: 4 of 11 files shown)


xdp-dump/Makefile
```diff
@@ -6,5 +6,6 @@ TEST_FILE := test-xdpdump.sh
 
 LIB_DIR   = ../lib
 USER_LIBS = -lpcap
+USER_CFLAGS = -DXDPDUMP_VERSION=\"$(LIBXDP_VERSION)\"
```
chaudron, line 9: After re-thinking this, I will change this to an XDPDUMP specific version in the Makefile, rather than use the libxdp version. Let me know if you still want to keep the USER_CFLAGS as it might be handy for other use cases?

tohojo: If a tool needs a particular CFLAG, can't the Makefile in that directory just do 'CFLAGS+=foo' ?

As for the libbpf version detection, I started hacking on it yesterday, but realised that we may be better off doing what configure scripts usually do: detect the feature itself, instead of matching on versions.

I.e., we can have configure try to compile a stub program with the perf buffer flush call in it, and if that succeeds, we set HAVE_PERF_FLUSH or some-such... I think this will be more maintainable in the long run :)

chaudron: The 'CFLAGS+=foo' does not work as common.mk is included below the use definitions. Maybe common.mk needs changing to make this work. For now I'll remove this patch and make it direct version reference.

tohojo: Well just add the += underneath the include statement?

chaudron:
My last push removed the .short_opt numeric values.

xdp-dump/Makefile
```diff
@@ -1,4 +1,5 @@
 # SPDX-License-Identifier: (GPL-2.0 OR BSD-2-Clause)
+XDPDUMP_VERSION := 0.1.0
```
tohojo, line 2: I was kinda thinking we could just reuse LIBXDP_VERSION for all the tools as well? Seems a bit of hassle to have to maintain several different version series in the same repository?

xdp-dump/xdpdump.c
```diff
@@ -99,17 +113,115 @@ static struct prog_option xdpdump_options[] = {
...
+
+	memset(&ereq, 0, sizeof(ereq));
+	ereq.req.cmd = ETHTOOL_GLINKSETTINGS;
+
+	memset(&ifr, 0, sizeof(ifr));
+	strncpy(ifr.ifr_name, iface->ifname, sizeof(ifr.ifr_name) - 1);
+	ifr.ifr_data = (void *)&ereq;
+
+	/* First query the kernel to see how many masks we need to ask for. */
+	if (ioctl(fd, SIOCETHTOOL, &ifr) != 0)
```
tohojo, line 161: Since we're dependent on a 5.7+ kernel anyway, why not use the ethtool netlink interface?

xdp-dump/xdpdump.c
```diff
@@ -99,17 +113,115 @@ static struct prog_option xdpdump_options[] = {
...
+	info.cmd = ETHTOOL_GDRVINFO;
+
+	memset(&ifr, 0, sizeof(ifr));
+	strncpy(ifr.ifr_name, iface->ifname, sizeof(ifr.ifr_name) - 1);
+	ifr.ifr_data = (void *)&info;
+
+	if (ioctl(fd, SIOCETHTOOL, &ifr) != 0)
+		goto exit;
+
+	snprintf(buffer, len,
```
tohojo, line 213: try_snprintf() + check return?

(This goes for all uses of snprintf)

xdp-dump/xdpdump.c
```diff
@@ -99,17 +113,115 @@ static struct prog_option xdpdump_options[] = {
...
+	/* Open socket, and initialize structures. */
+	fd = socket(AF_INET, SOCK_DGRAM, 0);
+	if (fd < 0)
+		return 0;
+
+	memset(&ereq, 0, sizeof(ereq));
+	ereq.req.cmd = ETHTOOL_GLINKSETTINGS;
+
+	memset(&ifr, 0, sizeof(ifr));
+	strncpy(ifr.ifr_name, iface->ifname, sizeof(ifr.ifr_name) - 1);
```
tohojo, line 157: Interfaces can have (longer) aliases these days; not sure what it would take to support that, but I guess we should add that throughout at some point?

xdp-dump/xdpdump.c
```diff
@@ -709,16 +852,73 @@ static bool capture_on_interface(struct dumpopt *cfg)
...
+
+			pcap_dumper = pcap_dump_open(pcap, cfg->pcap_file);
+			if (!pcap_dumper) {
+				pr_warn("ERROR: Can't open pcap file for writing!\n");
+				goto error_exit;
+			}
+		} else {
+			char           if_name[IFNAMSIZ + 7];
+			char           if_descr[BPF_OBJ_NAME_LEN + IFNAMSIZ + 10];
+			char           if_drv[260];
```
tohojo, line 871: I'm not doubting that you calculated the size of these buffers correctly, but it does make me a bit twitchy to see these + no checks of snprintf() return codes. At least use try_snprintf(), as I pointed out above :)

xdp-dump/xdpdump.c
```diff
@@ -847,6 +1051,12 @@ int main(int argc, char **argv)
 			       &defaults_dumpopt) != 0)
 		return EXIT_FAILURE;
 
+	/* Do we need to dump version information? */
+	if (cfg_dumpopt.version) {
```
tohojo, line 1055: For this I think we should just add a version output to params.c (to go along with --help and --verbose).

xdp-dump/xdpdump.h
```diff
@@ -9,7 +9,7 @@
 /******************************************************************************
  * General definitions
  ******************************************************************************/
-#define PERF_MMAP_PAGE_COUNT	16
+#define PERF_MMAP_PAGE_COUNT	256
```
tohojo, line 12: Why is this changing?

lib/testing/test_runner.sh
```diff
@@ -26,7 +26,7 @@ GENERATED_NAME_PREFIX="xdptest"
 ALL_TESTS=""
 TEST_PROG_DIR="${TEST_PROG_DIR:-$(dirname $0)}"
 
-NEEDED_TOOLS="ethtool ip ping tc tcpdump"
+NEEDED_TOOLS="capinfos ethtool ip ping tc tcpdump"
```
tohojo: Which package is this coming from? Please add that to the RPM spec and .travis.yml

xdp-dump/Makefile
chaudron: Are we ok with updating the libxdp version if nothing changes in the libraty itself? If so I can use that version (I had that initially).

xdp-dump/xdpdump.c
chaudron: We could, be do we want to pull in all the TLV parsing for Netlink? I thought this would be the easiest way. At the time I implemented this is still how ethtool does it.

lib/testing/test_runner.sh
chaudron: Its part of wireshark, will add the package.

xdp-dump/xdpdump.h
chaudron: With 16 it was easy to run out of perf buffer space, so I increased it to a decent number.

xdp-dump/xdpdump.c
chaudron: It makes sense if we decide to use the same version for all apps and libxdp.

xdp-dump/Makefile
tohojo: Hmm, that's a good point. I guess we may need to split them in the future. But for now at least, we need to keep them together, if nothing else because of the packaging. But how about we define TOOLS_VERSION := $(LIBXDP_VERSION) and use that as the output in --version, man page etc? Then we can always split the two versions down the line; I think it'll be fine to bump all the tool versions even when only updating one of them...

xdp-dump/xdpdump.h
tohojo: Right, OK. Just curious :)

xdp-dump/Makefile
chaudron: Will add this in the next commit

xdp-dump/xdpdump.c
chaudron: Will add a --version in the param.c

xdp-dump/xdpdump.c
chaudron: We do not support aliases for the -i option either so we are good for now.

xdp-dump/xdpdump.c
chaudron: Will fix in next commit

xdp-dump/xdpdump.c
chaudron: Will fix all snprintf() calls in the code

chaudron:
Update code, guess Travis and LGTM failure are due to man page changes you introduced.

lib/util/xpcapng.c
```diff
@@ -0,0 +1,630 @@
+// SPDX-License-Identifier: GPL-2.0
+/*
+ *  Description:
+ *      Simple PcapNG library developed from scratch as no library existed that
+ *      met the requirements for xdpdump. It can also be used by other XDP
+ *      applications that would like to capture packets for debugging purposes.
+ *
+ *  Notes:
```
tohojo, line 8: Notes? :)

lib/util/xpcapng.c
```diff
@@ -0,0 +1,630 @@
...
+	 */
+} __attribute__((__packed__));
+
+enum pcapng_opt_epb {
+	PCAPNG_OPT_EPB_FLAGS = 2,
+	PCAPNG_OPT_EPB_HASH,
+	PCAPNG_OPT_EPB_DROPCOUNT,
+
+	/* The below are suggestions based on the following pull request:
+	 *   https://github.com/pcapng/pcapng/pull/91
```
tohojo, line 159: This comment seems out of date now? :)

lib/util/xpcapng.c
```diff
@@ -0,0 +1,630 @@
...
+
+	if (user_application)
+		opt = pcapng_add_option(opt, PCAPNG_OPT_SHB_USERAPPL,
+					strlen(user_application),
+					user_application);
+
+	opt = pcapng_add_option(opt, PCAPNG_OPT_END, 0, NULL);
+	memcpy(opt, &shb->shb_block_length, sizeof(shb->shb_block_length));
+
+	/* Write the SHB, and free it's memory. */
```
tohojo, line 271: its memory

lib/util/util.mk
```diff
@@ -1 +1 @@
-UTIL_OBJS := params.o logging.o util.o stats.o
+UTIL_OBJS := params.o logging.o util.o stats.o xpcapng.o
```
tohojo, line 1: Why the 'x' prefix in the file name?

lib/util/xpcapng.c
```diff
@@ -0,0 +1,630 @@
...
+	shb->shb_byte_order_magic = PCAPNG_BYTE_ORDER_MAGIC;
+	shb->shb_major_version = PCAPNG_MAJOR_VERSION;
+	shb->shb_minor_version = PCAPNG_MINOR_VERSION;
+	shb->shb_section_length = UINT64_MAX;
+
+	/* Add the options and block_length value */
+	opt = (struct pcapng_option *) &shb->shb_options;
+
+	if (comment)
+		opt = pcapng_add_option(opt, PCAPNG_OPT_COMMENT,
```
tohojo, line 252: I get that these can't fail because the length was calculated above, but it still makes me a bit twitchy. Is there any way we could guard against, say, someone adding a new option in this part and forgetting to add it to the length calculation?

lib/util/xpcapng.c
```diff
@@ -0,0 +1,630 @@
...
+	idb->idb_block_type = PCAPNG_INTERFACE_BLOCK;
+	idb->idb_block_length = idb_length;
+	idb->idb_link_type = 1; /* Ethernet */
+	idb->idb_snap_len = snap_len;
+
+	/* Add the options and block_length value */
+	opt = (struct pcapng_option *) &idb->idb_options;
+
+	if (name)
+		opt = pcapng_add_option(opt, PCAPNG_OPT_IDB_IF_NAME,
```
tohojo, line 339: Same comment as above here, obviously :)

lib/util/xpcapng.c
```diff
@@ -0,0 +1,630 @@
...
+	epb.epb_timestamp_hi = timestamp >> 32;
+	epb.epb_timestamp_low = (uint32_t) timestamp;
+	epb.epb_captured_length = caplen;
+	epb.epb_original_length = len;
+
+	/* Add the flag/end option and block_length value */
+	opt = (struct pcapng_option *) options;
+
+	if (flags)
+		opt = pcapng_add_option(opt, PCAPNG_OPT_EPB_FLAGS,
```
tohojo, line 456: And here...

xdp-dump/xdpdump.c
tohojo: I guess we can change this if/when we add interface alias support :)

lib/util/xpcapng.c
```diff
@@ -0,0 +1,630 @@
...
+
+/*****************************************************************************
+ * xpcapng_dump_enhanced_pkt()
+ *****************************************************************************/
+bool xpcapng_dump_enhanced_pkt(struct xpcapng_dumper *pd, uint32_t ifid,
+			       enum pcapng_epb_flags flags, uint64_t timestamp,
+			       uint32_t len, uint32_t caplen,
+			       const uint8_t *pkt, uint64_t dropcount,
+			       const char *comment, uint64_t *packetid,
+			       uint32_t *queue, int64_t *xdp_verdict)
```
tohojo, line 625: 12 arguments? Wow. Could we maybe put these into a struct or something? The call to this becomes quite unreadable...

xdp-dump/xdpdump.c
```diff
@@ -250,24 +365,45 @@ static enum bpf_perf_event_ret handle_perf_event(void *private_data,
...
+			xpcapng_dump_enhanced_pkt(ctx->pcapng_dumper,
+						  fexit ? 1 : 0,
+						  PCAPNG_EPB_FLAG_INBOUND,
+						  ts, e->metadata.pkt_len,
+						  min(e->metadata.cap_len,
+						      ctx->cfg->snaplen),
+						  e->packet,
+						  ctx->last_missed_events,
+						  NULL, &pktid, &queue,
+						  fexit ? &action : NULL);
```
tohojo, line 383: This is quite unreadable... See also previous comment on the definition of this function.

.travis.yml
```diff
@@ -11,6 +11,7 @@ addons:
       - libpcap-dev
       - libelf-dev
       - gcc-multilib
+      - wireshark
```
tohojo, line 14: Travis doesn't actually run the tests yet, and when it does it'll be in a VM (since we need a newer kernel). So no reason to add this here; I meant that you should add it as a BuildRequires in `packaging/rpm/xdp-tools.spec`

lib/util/xpcapng.c
chaudron: Ooops, had some notes which I removed, forgot the header. Wil remove in the next update

lib/util/xpcapng.c
chaudron: thanks, fixed

lib/util/util.mk
chaudron: x was for xdp-tools, I can remove it if you prefer plain pcapng.c/h

lib/util/xpcapng.c
chaudron: Yes, removed.

lib/util/xpcapng.c
chaudron: It will definitely trash memory if they forget to add the code above for length. But the good part is that it will mess up the length in the structure and the first try to load the capture file in Wireshark or the self-test will fail as the file is invalid.

Will add a comment at the last option as a reminder to update length calculation above :)

lib/util/xpcapng.c
chaudron: See above

lib/util/xpcapng.c
chaudron: See above, will add comment

.travis.yml
chaudron: Will remove, and add wireshark to the spec file.

lib/util/xpcapng.c
chaudron: Yes, I guess it grew on me :) I'll re-order it, and add a structure for the non mendatory options.

xdp-dump/xdpdump.c
chaudron: Yes the above change will fix this...

lib/util/util.mk
tohojo: Right, I see. Hmm, I kinda see the point with that, so I guess we can keep it. I just thought it was odd because 'xpcapng' looks a lot like someone misspelled 'xdp' :)

lib/util/params.c
```diff
@@ -21,6 +21,7 @@
 
 #define BUFSIZE 30
 #define FIRST_PRINTABLE 65 /* ord('A') = 65 */
+#define VERSION_SHORT_OPT 0
```
tohojo, line 24: Why not just use 'V'?

chaudron: I decided that -V is not really a convention, and it's used by a lot of other command-line tools for other purposes. I did not want to restrict tools from using -V, as -v is already taken.

I did think of using -V and then check to see if the user options include a -V and then fall back to no short option. But I thought it was not worth the complexity.

tohojo: Nah, better to be consistent. I can live with no short option for `--version` :)

PR #48, opened 2020-06-05
xdp-dump: add support for multiple buffer wakeup

chaudron:


chaudron:
> There seem to be a lot of unrelated whitespace changes that got in here somehow. Could you try to get rid of those, please?

I cleaned up the file as there where tabs and space used. So made it all spaces... You want a separate commit for this?

tohojo:
Yes, please. Makes it clearer what actually changed for the new check :)

chaudron:
> Yes, please. Makes it clearer what actually changed for the new check :)

Done, added additional commit at the end.

xdp-dump/test-xdpdump.sh
```diff
@@ -62,6 +64,10 @@ Options:
 END
 	  )
 
+    if grep -q "HAVE_LIBBPF_PERF_BUFFER__CONSUME:=n" "$CONFIG_MK"; then
```
tohojo, line 67: I would like to make the tests runnable from the installed RPM. In which case config.mk won't be available. So can we come up with a better way to detect this?

Some ideas:
- grep the --help output
- have the Makefile write a test config that we can install along with the test scripts

xdp-dump/xdpdump.c
```diff
@@ -79,6 +81,12 @@ static struct prog_option xdpdump_options[] = {
 		      .short_opt = 'i',
 		      .metavar = "<ifname>",
 		      .help = "Name of interface to capture on"),
+#ifdef HAVE_LIBBPF_PERF_BUFFER__CONSUME
+	DEFINE_OPTION("perf-wakeup", OPT_U32, struct dumpopt, perf_wakeup,
+		      .short_opt = 4,
```
tohojo, line 86: Are these non-printable short opts really needed? If so, I'd prefer to fix params.c so they aren't...

chaudron: Yes, they are, else you would crash :( Your current code needs them.

tohojo: Right, okay, let's just merge this as-is, and I'll look into fixing params.c later...

xdp-dump/test-xdpdump.sh
tohojo: Didn't see any resolution of this?

chaudron: I missed this comment, let me change this, and ping you!

xdp-dump/xdpdump.c
chaudron: Don't forget you need to merge PcapNG first for the tests to work.

chaudron:
Fixed test check to use --help option to determine if wakeup feature is present.

xdp-dump/test-xdpdump.sh
```diff
@@ -3,10 +3,11 @@
 #
 # shellcheck disable=2039
 #
-ALL_TESTS="test_help test_interfaces test_capt_pcap test_capt_term test_exitentry test_snap test_none_xdp"
+ALL_TESTS="test_help test_interfaces test_capt_pcap test_capt_term test_exitentry test_snap test_perf_wakeup test_none_xdp"
 
 XDPDUMP=./xdpdump
 XDP_LOADER=../xdp-loader/xdp-loader
+CONFIG_MK=../config.mk
```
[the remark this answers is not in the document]
chaudron: oops, will fix in the next batch when I cleanup the short parameter.

chaudron:
Last commit change fixed/removed "CONFIG_MK=../config.mk", and fixed short option.

PR #62, opened 2020-08-12
lib/util: Fix skipping of keys while iterating map

shoracek:
Sometimes an added address was not printed out by the status subcommand, even though it was filtered correctly.


tohojo:
Ah, thank you so much for fixing this! Was just about getting to it, so you just beat me to it; great! :)

Please add the missing S-o-b lines, then this should be fine to merge :)

tohojo:
Awesome, thanks!

PR #63, opened 2020-08-17
xdp-filter: Fix printing of filtered MAC addresses

shoracek:
If the filter was loaded with a feature set not containing ip, status
subcommand did not print out the filtered MAC addresses.

Signed-off-by: Štěpán Horáček <shoracek@redhat.com>


tohojo:
Looks good, thanks!

PR #65, opened 2020-08-26
xdp-filters: fixed self test issue with tcpdump

chaudron:
Some times tcpdump gets terminated before it finishing reading the
buffers from the kernel. Rather than increasing the timeout, add the
"--immediate-mode" option to not delay any packet gathering.

In addition also removed the first 1 second delay, as it's already
part of the start_background shell function.


tohojo:
Thanks for fixing this! :)

PR #66, opened 2020-08-31
xdp-filter: Add basic IPv6 extension header support

shoracek:
Signed-off-by: Štěpán Horáček <shoracek@redhat.com>

### DIFF
--- a/headers/xdp/parsing_helpers.h
+++ b/headers/xdp/parsing_helpers.h
@@ -24,6 +24,7 @@
 #include <linux/icmpv6.h>
 #include <linux/udp.h>
 #include <linux/tcp.h>
+#include <linux/in.h>
 
 /* Header cursor to keep track of current parsing position */
 struct hdr_cursor {
@@ -54,6 +55,12 @@ struct icmphdr_common {
 #ifndef VLAN_MAX_DEPTH
 #define VLAN_MAX_DEPTH 4
 #endif
+
+/* Longest chain of IPv6 extension headers to resolve */
+#ifndef IPV6_EXT_MAX_CHAIN
+#define IPV6_EXT_MAX_CHAIN 6
+#endif
+
 
 static __always_inline int proto_is_vlan(__u16 h_proto)
 {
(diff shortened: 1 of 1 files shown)


headers/xdp/parsing_helpers.h
```diff
@@ -101,6 +102,39 @@ static __always_inline int parse_ethhdr(struct hdr_cursor *nh, void *data_end,
 	return h_proto; /* network-byte-order */
 }
 
+static __always_inline int skip_ip6hdrext(struct hdr_cursor *nh,
+					  void *data_end,
+					  __u8 currhdr)
+{
+	for (int i = 0; i < 8; ++i) {
```
tohojo, line 109: The number of loop iterations should be a define, like VLAN_MAX_DEPTH above

headers/xdp/parsing_helpers.h
```diff
@@ -101,6 +102,39 @@ static __always_inline int parse_ethhdr(struct hdr_cursor *nh, void *data_end,
 	return h_proto; /* network-byte-order */
 }
 
+static __always_inline int skip_ip6hdrext(struct hdr_cursor *nh,
+					  void *data_end,
+					  __u8 currhdr)
```
tohojo, line 107: Please find a better variable name than 'currhdr'. Maybe 'next_proto'?

headers/xdp/parsing_helpers.h
tohojo: Also, you can probably lower it to 6, since the code only handles 6 different headers, and none of them should occur more than once (according to the RFC).

headers/xdp/parsing_helpers.h
```diff
@@ -101,6 +102,39 @@ static __always_inline int parse_ethhdr(struct hdr_cursor *nh, void *data_end,
...
+			nh->pos = (char *)hdr + (hdr->hdrlen + 1) * 8;
+			break;
+		case IPPROTO_AH:
+			nh->pos = (char *)hdr + (hdr->hdrlen + 2) * 4;
+			break;
+		case IPPROTO_FRAGMENT:
+			nh->pos = (char *)hdr + 8;
+			break;
+		default:
+			return currhdr;
```
tohojo, line 129: A comment here making it explicit that this is the common case where we've found the "real" next proto header would help readability.

headers/xdp/parsing_helpers.h
```diff
@@ -101,6 +102,39 @@ static __always_inline int parse_ethhdr(struct hdr_cursor *nh, void *data_end,
...
+		}
+
+		currhdr = hdr->nexthdr;
```
tohojo, line 132: Having this assignment here makes the whole logic more difficult to follow, I think; so better have it in each case: statement along with the assignment to nh->pos, even if it means duplicating it a couple of times...

tohojo (changes requested):
Very nice! A few comments to aid readability, but otherwise LGTM :)

headers/xdp/parsing_helpers.h
shoracek: I decided to name it that way to mirror the "nexthdr" field of the headers struct. Also, I think that calling it protocol is not accurate, since the headers are part of the IPv6 protocol. But I don't really feel strongly about the naming.

tohojo: But it's not the 'current header' as the name applies. Rather it's the 'current next header' - which obviously is not a very good variable name. But just using 'currhdr' is confusing; if you can come up with something better, feel free! :)

PR #67, opened 2020-09-17
libxdp: Fix installing without supported Emacs

shoracek:
Makefile fails and prints
```
install: missing destination file operand after '/usr/local/share/man/man3'
Try 'install --help' for more information.
```
while installing without proper Emacs, since the MAN_FILES variable was empty.


tohojo:
Ah, oops - thanks for the fix! :)

PR #68, opened 2020-09-18
Adds tests using the xdp-test-harness library

shoracek:


chaudron:
Did not look at all the details, but I think we need to update the pre-run check to make sure we have python3 and the required python modules (scapy/etc.).

https://github.com/xdp-project/xdp-tools/blob/be624a9f12227a61c96707a2ae0ef52b07fa21f8/lib/testing/test_runner.sh#L118

tohojo:
> Did not look at all the details, but I think we need to update the pre-run check to make sure we have python3 and the required python modules (scapy/etc.).

Actually, I would rather skip the test if that's not available; but maybe we should make the tool detection a bit more generic (let each test function define which tools it needs and automatically skip it if they are missing)...

xdp-filter/test-xdp-filter.sh
```diff
@@ -261,6 +261,38 @@ test_print()
     check_run $XDP_FILTER unload $NS -v
 }
 
+run_python_test()
+{
+    python3 -c "import xdp_test_harness; import bcc;"
```
tohojo, line 266: See the other comment. For now, please just check that python itself exists (and it may be just 'python' not 'python3' - so maybe a detection and support for a $PYTHON variable is in order?)

xdp-filter/python-tests/xdp-filter-exec.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+cd ..
+./xdp-filter $@
```
tohojo, line 4: Why is this needed? When run from the test harness, $XDP_FILTER should always be set, no?

tohojo (changes requested):
Looks good, but a few nits :)

xdp-filter/test-xdp-filter.sh
shoracek: > Did not look at all the details, but I think we need to update the pre-run check to make sure we have python3 and the required python modules (scapy/etc.).

This line here returns non-zero if one of the required libraries is not installed. I am checking only for XDP test harness and BCC, since both Scapy and Pyroute2 are listed as dependencies for XDP test harness and so should be also installed by the package manager.
This should also check whether Python exists, since if that is the case the shell will return non-zero value. Python 2 will not work, because the library uses some features of Python 3.

xdp-filter/test-xdp-filter.sh
```diff
@@ -261,6 +261,38 @@ test_print()
...
+    if [ $? -ne 0 ]; then
+        # Libraries are not installed.
+        return "$SKIPPED_TEST"
+    fi
+
+    # Tests are in python-tests directory in a build tree and in tests directory after install.
```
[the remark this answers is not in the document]
shoracek: I did not want to clutter the `xdp-filter` directory and having `tests` directory contain only the new tests seemed misleading. Maybe move all the tests (even xdpdump and xdp-loader) into `tests` directories, even though with the other tools it will contain only one file?

lib/common.mk
```diff
@@ -76,6 +76,7 @@ install: all install_local
 	$(if $(SCRIPTS_FILES),install -m 0755 $(SCRIPTS_FILES) $(DESTDIR)$(SCRIPTSDIR))
 	$(if $(TEST_FILE),install -m 0755 -d $(DESTDIR)$(SCRIPTSDIR)/tests)
 	$(if $(TEST_FILE),install -m 0644 $(TEST_FILE) $(DESTDIR)$(SCRIPTSDIR)/tests)
+	$(if $(TEST_FILE_DEPS),install -m 0644 $(TEST_FILE_DEPS) $(DESTDIR)$(SCRIPTSDIR)/tests)
```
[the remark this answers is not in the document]
shoracek: What about adding just a prefix to the test file names? E.g. `xdp-filter-common.py`.

xdp-filter/python-tests/xdp-filter-exec.sh
shoracek: I used this since I was running the tests from the test folder and so xdp-filter could not find object files for the XDP programs. I will rewrite it so that the working directory is the the one containing the compiled files and remove this file. `XDP_FILTER` variable actually does not get inherited into the Python, since it is just local, will fix that.

xdp-filter/test-xdp-filter.sh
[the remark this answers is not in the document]
tohojo: Hmm, yeah having all test files in $TOOL/tests, and installing them into /usr/share/xdp-tools/tests/$TOOL seems fine. And then export the directory from the test harness so you don't need to have a conditional in the script itself...

xdp-filter/test-xdp-filter.sh
tohojo: Yup, you should definitely check that python itself exists as well. The version check will be taken care of by the import check, I think (the module won't import correctly in py2). However, python3 can be called just 'python' on some systems; and we may have build systems where it's set up weirdly, so supporting a $PYTHON variable seems sensible...

xdp-filter/python-tests/xdp-filter-exec.sh
tohojo: Setting the work directory to where the binary is doesn't work after install (we don't want to run with workdir in /usr/bin). Just exporting XDP_FILTER so the Python code can get it from the environment should be enough, no?

lib/common.mk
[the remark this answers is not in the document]
tohojo: If we do the above ($TOOL/tests installed into tests/$TOOL) this problem goes away :)

xdp-filter/test-xdp-filter.sh
chaudron: Note that these tests can also be run from the Github clone, so I think checks for scapy and pyroute2 should also be added and any other dependency.

tohojo: Hmm, I was assuming that was implicit in 'import xdp_test_harness' as that would in turn try to import scapy and pyroute2. But not sure that's actually the case for just this import? If not, I agree, we should check for the other modules as well...

shoracek: Using `import xdp_test_harness` does not check for Scapy or pyroute2. What I meant was that pip while installing XDP test harness will also install its dependencies: Scapy and pyroute2, meaning if XDP test harness is installed then so should be the dependencies. But I am okay with adding the libraries explicitly into the check.

tohojo: Hmm, right. Well I think it's better to make sure the dependencies are also there, but that could also be done by just importing them from the top-level xdp_test_harness module. That would make sure we check for broken installations while still keeping the dependencies in the module itself, not in this script...

lib/common.mk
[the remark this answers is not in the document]
tohojo: You can just drop this entire commit now, since it's folded into the next one anyway...

lib/testing/run_tests.sh
```diff
@@ -7,8 +7,8 @@ TEST_RUNNER="$TEST_PROG_DIR/test_runner.sh"
 RET=0
 
 echo "Running all tests from $TESTS_DIR"
-for f in "$TESTS_DIR"/*.sh; do
-    "$TEST_RUNNER" "$f" || RET=1
+for f in "$TESTS_DIR"/*; do
+    "$TEST_RUNNER" "$f/test-$(basename $f).sh" || RET=1
```
tohojo, line 11: Took me a few tries to understand what's going on here; please split this up so it becomes more clean that $(basename $f) is the tool name.

Or maybe this could just be a two-level glob (for f in "$TESTS_DIR"/*/*.sh) ? Did you try that?

lib/testing/test_runner.sh
```diff
@@ -446,6 +446,8 @@ TEST_DEFINITIONS="${1:-}"
 [ -f "$TEST_DEFINITIONS" ] || usage
 source "$TEST_DEFINITIONS"
 
+TOOL_TESTS_DIR=$(dirname $TEST_DEFINITIONS)
```
tohojo, line 477: What is this for?

xdp-filter/tests/common.py
```diff
@@ -0,0 +1,59 @@
+import os
+import subprocess
+
+from xdp_test_harness.xdp_case import XDPCase, usingCustomLoader
+from xdp_test_harness.utils import XDPFlag
+
+
+if "XDP_FILTER" in os.environ:
```
tohojo, line 8: `XDP_FILTER = os.environ.get("XDP_FILTER", "xdp-filter")` ?

lib/testing/test_runner.sh
```diff
@@ -251,6 +251,28 @@ get_vlan_prefix()
     (IFS=:; set -- $prefix; printf "%s:%s:%s:%x::" "$1" "$2" "$3" $(($4 + $vid * 4096)))
 }
 
+check_python()
+{
+    if [ -z ${PYTHON+unset} ]; then
```
tohojo, line 256: What's the `+unset`?

Just `if [[ -z "$PYTHON" ]]` ?

(Please use double-brace tests for all string tests, we're running this through bash deliberately so we don't have to dump down to /bin/sh syntax)

xdp-filter/tests/test-xdp-filter.sh
```diff
@@ -261,6 +261,36 @@ test_print()
     check_run $XDP_FILTER unload $NS -v
 }
 
+
+run_python_test()
+{
+    if [ "${PYTHON}" = "invalid" ]; then
```
tohojo, line 267: Please use double braces and the `==` test here

xdp-filter/tests/test-xdp-filter.sh
```diff
@@ -261,6 +261,36 @@ test_print()
...
+        return "$SKIPPED_TEST"
+    fi
+
+    MODULE_PATH=$(realpath --relative-to=. "$TOOL_TESTS_DIR" | sed "s/\//./g")
+    if [ $? -ne 0 ] || [ "$MODULE_PATH" = "." ]; then
```
tohojo, line 272: and here

xdp-filter/tests/test-xdp-filter.sh
```diff
@@ -261,6 +261,36 @@ test_print()
...
 }
 
+
+run_python_test()
+{
+    if [ "${PYTHON}" = "invalid" ]; then
+        return "$SKIPPED_TEST"
+    fi
+
+    MODULE_PATH=$(realpath --relative-to=. "$TOOL_TESTS_DIR" | sed "s/\//./g")
```
tohojo, line 271: Ah, that's where you're using TOOL_TESTS_DIR - please add it in the same commit, then :)

xdp-filter/Makefile
```diff
@@ -5,10 +5,12 @@ XDP_TARGETS  := xdpfilt_dny_udp xdpfilt_dny_tcp xdpfilt_dny_ip \
 	xdpfilt_alw_udp xdpfilt_alw_tcp xdpfilt_alw_ip \
 	xdpfilt_alw_eth xdpfilt_alw_all
 
+TOOL_NAME := xdp-filter
 USER_TARGETS := xdp-filter
 EXTRA_DEPS := xdpfilt_prog.h
 MAN_PAGE := xdp-filter.8
-TEST_FILE := test-xdp-filter.sh
+TEST_FILE := tests/test-xdp-filter.sh
+TEST_FILE_DEPS := tests/*.py
```
tohojo, line 13: This should probably use a make `glob()`...

tohojo (changes requested):
A few more comments, but getting there! :)

lib/testing/run_tests.sh
shoracek: I wanted to use files called `test-$(TOOL_NAME).sh` as an entry point to the tests (since there might be also other shell scripts), but maybe just files with a `test-` prefix?

xdp-filter/Makefile
shoracek: Do you mean `$(wildcard`? Couldn't find any mention of `glob()` in make manual.

lib/testing/test_runner.sh
shoracek: I used the substitution since `nounset` is set and so `if [[ -z "$PYTHON" ]]` gives unbound variable error. And I used single brackets because they were used in the rest of the file, but I will use double brackets instead.

lib/testing/run_tests.sh
tohojo: `for f in "$TESTS_DIR"/*/test-*.sh` then? I think it would be fine to support multiple entry points in case we want to split things up in the future...

Also, probably test that the file exists before executing it (in case the glob doesn't match anything)

lib/testing/test_runner.sh
tohojo: Ah, I see... The fact that you can omit the colon is buried somewhat in the bash man page.

However, in this case I do actually think we'd want to match on both unset and null, so maybe `if [[ -z "${PYTHON:-}" ]]` is better?

xdp-filter/Makefile
tohojo: Yeah, wildcard, sorry about that - hard to keep track on all the different tool names for the same thing ;)

lib/testing/test_runner.sh
```diff
@@ -251,6 +251,28 @@ get_vlan_prefix()
     (IFS=:; set -- $prefix; printf "%s:%s:%s:%x::" "$1" "$2" "$3" $(($4 + $vid * 4096)))
 }
 
+check_python()
+{
+    if [[ -z ${PYTHON:-} ]]; then
+        if command -v python3; then
+            PYTHON=python3
```
tohojo, line 258: Since you're running 'command' anyway, why not assign the output to $PYTHON?

tohojo (changes requested):
A few more cosmetic issues...

lib/testing/test_runner.sh
shoracek: Couldn't it be an alias? Then it would break.

lib/testing/test_runner.sh
```diff
@@ -251,6 +251,28 @@ get_vlan_prefix()
     (IFS=:; set -- $prefix; printf "%s:%s:%s:%x::" "$1" "$2" "$3" $(($4 + $vid * 4096)))
 }
 
+check_python()
```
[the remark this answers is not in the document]
shoracek: Since a test function is run in a subshell, changed variables are not reflected in the parent shell. So I don't think it could easily be in the test file without checking Python every time.

lib/testing/test_runner.sh
tohojo: Hmm, right, 'command' can return alias definitions as well. I don't think that's good thing, though, if an alias is defined (and actually picked up by the script) that is more likely to just lead to confusion. So maybe switch to using 'which' instead?

lib/testing/test_runner.sh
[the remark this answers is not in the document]
tohojo: True, but does that matter? It's not like it's a huge overhead...

xdp-filter/tests/test-xdp-filter.sh
```diff
@@ -261,6 +261,64 @@ test_print()
...
 
+
+get_python()
+{
+    if [[ -z "${PYTHON:-}" ]]; then
+        local -a possible=(python3 python)
+        local -a available
+
+        available=($(which --skip-alias --skip-functions "${possible[@]}" 2> /dev/null))
+        if [[ $? == "${#possible[@]}" ]]; then
```
tohojo, line 272: Here you're comparing the return code to the number of elements in the array? With a string comparison? Shouldn't it just be `$? -ne 0` like below?

shoracek: The exit code is the number of failed arguments, so I am checking whether it failed to find them all.

tohojo: Ah, I see. You're really bringing out all the esoteric shell features here, huh? ;)

Please add a comment explaining this; and it should be `-eq` instead of `==` for numeric comparison :)

tohojo (approved):
And with that I'm out of things to complain about - nice work! :)

PR #71, opened 2020-11-08
Clean processes

shoracek:
While trying to fix the problem with leftover processes due to stop_background not being called, I found that some sometimes stop_background did not stop the command completely, so I fixed that also.

### DIFF
--- a/lib/testing/test_runner.sh
+++ b/lib/testing/test_runner.sh
@@ -359,6 +359,13 @@ teardown()
         [[ "$CUR" == "$NS" ]] && rm -f "$STATEDIR/current"
     fi
 
+    for f in ${STATEDIR}/proc/*; do
+        if [ -f "$f" ]; then
+            local pid="${f/${STATEDIR}\/proc\//}"
+            stop_background "$pid" &> /dev/null || true
+        fi
+    done
+
     rm -rf "$STATEDIR"
 
     CLEANUP_FUNC=
--- a/lib/testing/test_runner.sh
+++ b/lib/testing/test_runner.sh
@@ -78,7 +78,7 @@ die()
 start_background()
 {
     local TMP_FILE="${STATEDIR}/tmp_proc_$$_$RANDOM"
-    eval "$* >& ${TMP_FILE} &"
+    setsid bash -c "$*" &> ${TMP_FILE} &
     local PID=$!
     sleep 1 # Wait to make sure the command is executed in the background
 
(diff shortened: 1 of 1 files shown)


lib/testing/test_runner.sh
```diff
@@ -357,6 +357,15 @@ teardown()
 
     rm -rf "$STATEDIR"
 
+    for i in /tmp/"${NS}"_PID_*; do
```
chaudron, line 360: This won't work, as it will kill all processes, even those that are running in parallel. To support this some global per test run ID needs to be added.

tohojo: I think the right thing here is to change the tmpfile to a directory (with mktemp -d) and dump everything in there; that also makes it easier to find all the pid files for a run...

shoracek: I'm not sure what do you mean by processes running in parallel. If you mean processes from tests that are running in parallel, then they all would have to have a different $NS, wouldn't they?

chaudron: I was under the impression we could run the tests in parallel (or maybe I was hoping for this ;), but looking at the code this is not the case, so we might not need to worry about it for now.

shoracek:
I added commit moving all state files of the test into a temporary folder, as suggested by @tohojo .

lib/testing/test_runner.sh
```diff
@@ -123,7 +123,11 @@ check_prereq()
         die "This script needs root permissions to run."
     fi
 
-    [ -d "$STATEDIR" ] || mkdir -p "$STATEDIR" || die "Unable to create state dir $STATEDIR"
+    STATEDIR="$(mktemp -d --tmpdir=${TMPDIR:-/tmp} --suffix=.xdptest)" 
+    if [ $? -ne 0 ]; then
+        die "Unable to create state dir in $TMPDIR"
+    fi
+    mkdir ${STATEDIR}/proc
```
tohojo, line 130: cleanup() does `rmdir $STATEDIR`, so with this I think it'll fail, no? I think the easiest is to just change the cleanup to 'rm -rf' :)

shoracek: Looking at the code, I'm having trouble finding where the cleanup() function is even used. I thought that the cleaning was done in teardown().

tohojo: Huh, you're right! Looks like cleanup() is just dead code left over from when this was copied from the tutorial runner. My bad!

tohojo:
Looks good, thanks! :)

PR #73, opened 2020-11-20
xdp-loader: Change to loading programs in native mode by default

tohojo:
Previously we were loading programs without specifying an explicit mode by
default, letting the kernel pick. This causes problems for a couple of
reasons:

- It causes silent fallback from native to generic mode, making it harder
  to predict which mode a program would be loaded in.

- There's an nfp driver bug which makes it harder to unload such programs.

Arguably, the default should have been to just load in native mode from the
start. So let's change the default, but let the user go back to the
kernel-picked mode by specifying '-m unspecified' when loading. Document
the change in defaults in the man page.

----

@shoracek discovered why the nfp driver was having issues with unloading programs.
Arguably we should have been using native mode from the beginning, but let's
discuss whether we should change it at this point.

@chaudron, @netoptimizer, @shoracek comments?

### DIFF
--- a/xdp-loader/README.org
+++ b/xdp-loader/README.org
@@ -48,8 +48,12 @@ The supported options are:
 ** -m, --mode <mode>
 Specifies which mode to load the XDP program to be loaded in. The valid values
 are 'native', which is the default in-driver XDP mode, 'skb', which causes the
-so-called /skb mode/ (also known as /generic XDP/) to be used, or 'hw' which
-causes the program to be offloaded to the hardware.
+so-called /skb mode/ (also known as /generic XDP/) to be used, 'hw' which causes
+the program to be offloaded to the hardware, or 'unspecified' which leaves it up
+to the kernel to pick a mode (which it will do by picking native mode if the
+driver supports it, or generic mode otherwise). Note that using 'unspecified'
+can make it difficult to predict what mode a program will end up being loaded
+in. For this reason, the default is 'native'.
 
 ** -p, --pin-path <path>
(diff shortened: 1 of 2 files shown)


chaudron:
I'm ok with this change, as it makes sense to be the default. We should avoid auto fall back to native by default. I think we are early in the adoption phase to be able to change it ;)

xdp-loader/README.org
```diff
@@ -48,8 +48,13 @@ The supported options are:
 ** -m, --mode <mode>
 Specifies which mode to load the XDP program to be loaded in. The valid values
 are 'native', which is the default in-driver XDP mode, 'skb', which causes the
-so-called /skb mode/ (also known as /generic XDP/) to be used, or 'hw' which
-causes the program to be offloaded to the hardware.
+so-called /skb mode/ (also known as /generic XDP/) to be used, 'hw' which
+causes the program to be offloaded to the hardware, or 'unspecified' which
+leaves it up to the kernel to pick a mode (which it will do by picking native
+mode if the driver supports it, or generic mode otherwise). The default is
+'native'; this is a change from the previous default of 'unspecified', which
```
chaudron, line 55: I think you should remove the "; this is a change from...." part, as it makes no sense to keep this history in the documentation file. Maybe we should add a NEWS or RELEASE file to keep track of new features/changes?

tohojo: Yeah, good point, I'll remove the 'previous' bit and just note the difficulty with unspecified without any history.

PR #75, opened 2020-11-25
Fix compiling on OpenWrt

PolynomialDivision:
Just a draft to track progress.
Comments are welcome. No idea if this is a good way of doing such stuff...

I still have the error:
```
xdp-dispatcher.c:5:10: fatal error: 'bpf/bpf_helpers.h' file not found
#include <bpf/bpf_helpers.h>
         ^~~~~~~~~~~~~~~~~~~
1 error generated.
```

### DIFF
--- a/configure
+++ b/configure
@@ -242,6 +242,26 @@ EOF
     echo "HAVE_LIBBPF_PERF_BUFFER__CONSUME:=y" >>"$CONFIG"
 }
 
+check_secure_getenv()
+{
+    cat >$TMPDIR/secure_getenv.c <<EOF
+#define _GNU_SOURCE
+#include <stdlib.h>
+int main(int argc, char **argv) {
+	secure_getenv("test");
+    return 0;
+}
+EOF
+    secure_getenv_err=$($CC -o $TMPDIR/secure_getenv $TMPDIR/secure_getenv.c 2>&1)
+    if [ "$?" -eq "0" ]; then
+        echo "HAVE_SECURE_GETENV:=y" >>"$CONFIG"
+        echo "yes"
+    else
+        echo "HAVE_SECURE_GETENV:=n" >>"$CONFIG"
+        echo "no"
+    fi
+}
+
 quiet_config()
 {
     cat <<EOF
@@ -282,6 +302,9 @@ check_toolchain
 
 echo -n "libbpf support: "
 check_libbpf
+
+echo -n "secure_getenv support: "
+check_secure_getenv
 
 if [ -n "$KERNEL_HEADERS" ]; then
     echo "kernel headers: $KERNEL_HEADERS"
--- a/lib/libxdp/libxdp.c
+++ b/lib/libxdp/libxdp.c
@@ -28,6 +28,8 @@
 #include <xdp/libxdp.h>
 #include <xdp/prog_dispatcher.h>
 
+#include "compat.h"
+
 #define XDP_RUN_CONFIG_SEC ".xdp_run_config"
 
 #define ARRAY_SIZE(x) (sizeof(x) / sizeof(*(x)))
--- a/lib/util/params.c
+++ b/lib/util/params.c
@@ -54,8 +54,11 @@ static int handle_multistring(char *optarg, void *tgt, void *typearg)
 	struct multistring *opt_set = tgt;
 	void *ptr;
 
-	ptr = reallocarray(opt_set->strings, sizeof(*opt_set->strings),
-			   opt_set->num_strings + 1);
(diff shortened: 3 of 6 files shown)


PolynomialDivision:
Ah, I think I have to checkout git submodules. ;)

PolynomialDivision:
Maybe u want to cherry-pick 60c4db670cda9ce4d6b55e2b5454c06db36987c6

tohojo:
For reallocarray() I think it's simpler to just open-code the overflow check and always switch to realloc(), since we're only using it in that one place...

As for the secure_getenv() wrapper, instead of checking specifically for musl, could you please just add a check to 'configure' for that specific function and define it based on that? See check_perf_consume() for an example.

configure
```diff
@@ -242,6 +242,29 @@ EOF
...
+check_secure_getenv()
+{
+    cat >$TMPDIR/secure_getenv.c <<EOF
+#define _GNU_SOURCE
+#include <stdlib.h>
+#define XDP_OBJECT_ENVVAR "LIBXDP_OBJECT_PATH"
+int main(int argc, char **argv) {
+    char *path, **p;
+
+	path = secure_getenv(XDP_OBJECT_ENVVAR);
```
tohojo, line 254: Don't really need the `#define XDP_OBJECT_ENVVAR` or the variables; you can just do `secure_getenv("test");` and not assign the return value anywhere...

configure
```diff
@@ -242,6 +242,29 @@ EOF
...
+    return 0;
+}
+EOF
+    secure_getenv_err=$($CC -o $TMPDIR/secure_getenv $TMPDIR/secure_getenv.c $LIBBPF_CFLAGS $LIBBPF_LDLIBS 2>&1)
```
tohojo, line 258: No need for the LIBBPF_{CFLAGS,LDLIBS}

lib/libxdp/libxdp.c
```diff
@@ -28,6 +28,10 @@
 #include <xdp/libxdp.h>
 #include <xdp/prog_dispatcher.h>
 
+#if !HAVE_SECURE_GETENV
```
tohojo, line 31: `#ifndef` - also you need to add the -D to defines.mk

lib/util/secure_getenv.c
```diff
@@ -0,0 +1,9 @@
+#if !HAVE_SECURE_GETENV
+#include <stdlib.h>
+
+// Source: https://www.openwall.com/lists/musl/2019/05/28/3
+char *secure_getenv(const char *name)
```
tohojo, line 5: Just turn this into a 'static inline' function and keep it entirely in the header file?

tohojo (changes requested):
On the right track :)

lib/util/secure_getenv.c
PolynomialDivision: Yep. Much better. :)

PolynomialDivision:
> For reallocarray() I think it's simpler to just open-code the overflow check and always switch to realloc(), since we're only using it in that one place...

How would u do the overflow check? I just now switched to realloc Just checking the size after the multiplication? I will try implement something.

PolynomialDivision:
@tohojo Like that? https://github.com/xdp-project/xdp-tools/pull/75/files#diff-4ef1b5443fb0e5b06a85d68a016050cb03d7b40d2d91a08c8d94341d3743c7baR56-R68

lib/util/secure_getenv.h
```diff
@@ -0,0 +1,14 @@
+#ifndef HAVE_SECURE_GETENV
```
tohojo, line 1: Put this around the function definition only - makes it easy to add more compat stuff later if needed...

lib/util/params.c
```diff
@@ -53,9 +53,20 @@ static int handle_multistring(char *optarg, void *tgt, void *typearg)
...
+	num_strings = opt_set->num_strings + 1;
+	max_strings = SIZE_MAX / sizeof(*opt_set->strings);
+	if (max_strings < num_strings) {
+		// overflow
+		ptr = NULL;
+	}
+	else {
+		// no overflow
+		ptr = realloc(opt_set->strings, num_strings);
+	}
```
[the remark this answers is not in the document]
PolynomialDivision: I guess you mean `	ptr = realloc(opt_set->strings, sizeof(*opt_set->strings) * (opt_set->num_strings +1));` ;)

lib/util/secure_getenv.h
PolynomialDivision: I wrapped that around the include statement, too. Or is that bad practice?

tohojo (approved):
LGTM - let's wait for CI and see if it agrees :)

PR #76, opened 2020-11-27
Fix hw mode

shoracek:


lib/libxdp/libxdp.c
```diff
@@ -947,9 +948,11 @@ struct xdp_program *xdp_program__open_file(const char *filename,
 		goto err;
 	}
 
-	err = reuse_bpf_maps(obj, filename, opts);
-	if (err)
-		goto err_close_obj;
+	if (reuse) {
+		err = reuse_bpf_maps(obj, filename, opts);
```
tohojo, line 952: Can't we detect that reuse is not possible in reuse_bpf_maps() instead of having the condition?

lib/libxdp/libxdp.c
```diff
@@ -1712,6 +1721,9 @@ struct xdp_multiprog *xdp_multiprog__get_from_ifindex(int ifindex)
 	if (xinfo.attach_mode == XDP_ATTACHED_SKB) {
 		prog_id = xinfo.skb_prog_id;
 		mode = XDP_MODE_SKB;
+	} else if (xinfo.attach_mode == XDP_ATTACHED_HW) {
+		prog_id = xinfo.hw_prog_id;
+		mode = XDP_MODE_HW;
```
tohojo, line 1726: A program can be attached in HW mode at the same time as a native/generic mode program is attached; we should handle that properly...

tohojo (changes requested):

lib/libxdp/libxdp.c
shoracek: I couldn't find any. The point in the execution where reuse_bpf_maps() gets called is at the part where the execution is the same for native and for hw mode. The arguments with which xdp_program__open_file()/reuse_bpf_maps() do not give any information about that. So maybe introduce xdp_program__open_file() that does not reuse?

lib/libxdp/libxdp.c
shoracek: But that would also require change of API, right? Since right now we return one multiprogram. I don't think we can just put the hw program as one of the dispatched programs, so maybe a new field in the multiprog just for offloaded programs?

tohojo: Yeah, I was thinking a new member of `struct xdp_multiprog` for the hw prog, and a new accessor like `xdp_multiprog__main_prog()`. And we can introduce a new MULTI attach_mode like the kernel has, and return that from `xdp_multiprog__attach_mode()` if there are multiple programs attached.

lib/libxdp/libxdp.c
tohojo: Hmm, yeah we may need a new open function specifically for that. Bugger.

What happens if you leave out this bit? HW mode fails completely if it's using maps, right?

shoracek: It will try to use maps with different ifindexes than the program, and fail.

tohojo: Right, so sorry for taking so long to circle back to this. I think maybe in this instance we should make use of the fact that the call to reuse_bpf_maps() is a stopgap solution until we can use the freplace reattach functionality in the kernel (at which point the call to reuse is no longer necessary as we can handle already-loaded programs).

So in other words, if you just remove the call to reuse_bpf_maps() entirely as part of this series, we can at least make progress on fixing the HW mode stuff :)

tohojo:
> @shoracek pushed 5 commits.
>
> 8c7cc3d1b0a617176aea9d49b07005435bb23c9b  libxdp: Do not reuse maps
> 66f0fa8b53fefd226e8de750e17f3ff58139f616  libxdp: Fix offloading
> ec5c1755be0e597ee53d87ed980ce284d0d24e5d  xdp-filter: Add failure while trying to load in hw mode
> 30231553b23be6a2f1f7933e6fbe6f4c9426d0dc  libxdp: Add HW prog support for multiprog struct
> 54175270ad7acbd8b37865ee864b0c5272760586  xdp-loader: Support multiple programs on one interface

Great! I'll take a closer look at this tomorrow; but one thing that
immediately springs to mind is the lack of commit messages.

I think it's great that you're adding this in smaller commits, but
please write some text for each describing what each patch does and how
it fits together with the next ones. And for the new API for dealing
with hw progs in struct xdp_multiprog, a few short usage examples in the
commit message would be helpful, as it was not quite clear from a quick
glance through the code how a user is supposed to use this (e.g., how to
discover which programs are loaded on an interface through libxdp, how
to unload one or multiple of them, etc)...

lib/libxdp/libxdp.c
```diff
@@ -1292,18 +1299,28 @@ int xdp_program__detach_multi(struct xdp_program **progs, size_t num_progs,
...
 	mp = xdp_multiprog__get_from_ifindex(ifindex);
-	if (IS_ERR_OR_NULL(mp) || mp->is_legacy) {
+	if (IS_ERR_OR_NULL(mp)) {
 		pr_warn("No XDP dispatcher found on ifindex %d\n", ifindex);
 		return -ENOENT;
 	}
 
-	if (mode != XDP_MODE_UNSPEC && mp->attach_mode != mode) {
+	if (mp->is_legacy) {
+		err = xdp_multiprog__attach(mp, NULL, mode);
```
tohojo, line 1308: This should check that the main_prog matches what we are trying to detach...

tohojo: And also bail if num_progs >1

lib/libxdp/libxdp.c
```diff
@@ -2112,40 +2169,61 @@ static int xdp_multiprog__attach(struct xdp_multiprog *old_mp,
...
 	if (mp) {
+		if (mode == XDP_MODE_HW)
+			return -EINVAL;
+
 		prog_fd = xdp_multiprog__main_fd(mp);
 		if (prog_fd < 0)
 			return -EINVAL;
 		ifindex = mp->ifindex;
+
+		err = xdp_attach_fd(prog_fd, old_fd, ifindex, mode);
```
tohojo, line 2186: Erm, what? This is an unconditional detach (old_fd is always -1 here)?

lib/libxdp/libxdp.c
```diff
@@ -2112,40 +2169,61 @@ static int xdp_multiprog__attach(struct xdp_multiprog *old_mp,
 				 struct xdp_multiprog *mp,
 				 enum xdp_attach_mode mode)
 {
-	int err = 0, prog_fd = -1, old_fd = -1, ifindex = -1;
+	int err = 0, prog_fd = -1, old_fd = -1, hw_fd = -1, ifindex = -1;
```
tohojo, line 2172: I think it's better to leave HW mode out of this function entirely, and just create a separate function to handle that...

lib/libxdp/libxdp.c
```diff
@@ -2157,11 +2235,11 @@ int xdp_multiprog__detach(struct xdp_multiprog *mp)
 	if (!mp || !mp->is_loaded)
 		return -EINVAL;
 
-	err = xdp_multiprog__attach(mp, NULL, mp->attach_mode);
+	err = xdp_multiprog__attach(mp, NULL, XDP_MODE_UNSPEC);
```
tohojo, line 2238: Eh? Why? :)

lib/libxdp/libxdp.c
```diff
@@ -2157,11 +2235,11 @@ int xdp_multiprog__detach(struct xdp_multiprog *mp)
...
 	if (err)
 		return err;
 
-	if (!mp->is_legacy)
+	if (!mp->is_legacy && mp->main_prog)
```
tohojo, line 2242: How can main_prog be unset if is_legacy is false?

lib/libxdp/libxdp.c
```diff
@@ -1531,7 +1565,8 @@ static int xdp_multiprog__link_pinned_progs(struct xdp_multiprog *mp)
 	goto out;
 }
 
-static int xdp_multiprog__fill_from_fd(struct xdp_multiprog *mp, int prog_fd)
+static int xdp_multiprog__fill_from_fd(struct xdp_multiprog *mp, int prog_fd,
+				       bool fill_hw)
```
tohojo, line 1569: Instead of a bool, how about turning this into an 'int hw_fd' and handle that at the bottom after the main programs are added...

lib/libxdp/libxdp.c
```diff
@@ -1292,18 +1299,28 @@ int xdp_program__detach_multi(struct xdp_program **progs, size_t num_progs,
...
+
+	if (mode != XDP_MODE_UNSPEC && mp->attach_mode != mode && mode != XDP_MODE_HW) {
 		pr_warn("XDP dispatcher attached in mode %d, requested %d\n",
 			mp->attach_mode, mode);
 		err = -ENOENT;
 		goto out;
 	}
 
+	if (mode == XDP_MODE_HW) {
+		err = xdp_multiprog__attach(mp, NULL, mode);
```
tohojo, line 1320: See below re: not using xdp_multiprog__attach() here...

xdp-loader/xdp-loader.c
```diff
@@ -271,31 +287,43 @@ int print_iface_status(const struct iface *iface,
...
-		if (err)
-			return err;
-
-		printf("%-16s %-5d %-16s %-8s %-4u %-17s %s\n",
-		       " =>", xdp_program__run_prio(prog),
-		       xdp_program__name(prog),
-		       "", xdp_program__id(prog),
-		       print_bpf_tag(tag, xdp_program__tag(prog)),
-		       buf);
+	hw_prog = xdp_multiprog__hw_prog(mp);
```
tohojo, line 318: I think we should print this before the others, as HW progs are executed first...

xdp-loader/xdp-loader.c
```diff
@@ -219,8 +218,26 @@ int do_unload(const void *cfg, __unused const char *pin_root_path)
...
+			}
+		}
+		if (!prog) {
+			prog = xdp_multiprog__hw_prog(mp);
+			if (xdp_program__id(prog) == opt->prog_id) {
+				mode = XDP_MODE_HW;
+			} else {
+				prog = NULL;
+			}
 		}
```
tohojo, line 241: Instead of setting and reset'ing 'prog' I think you can simplify this a bit by just using 'goto found' if you find the right program...

lib/libxdp/libxdp.c
tohojo: Oh, if there's *only* an HW prog. In that case I think we should also set is_legacy...

tohojo (changes requested):
This is definitely on the right track! It gets a little convoluted to also have the hw_prog in xdp_multiprog, but I don't really think that can be helped.

However, there are some places where the logic is a bit **more** convoluted that it needs to be; see detailed comments below. In particular, I think it would be clearer to not handle the hw_prog in xdp_multiprog__attach(), but have a separate routine only for attach/detach of the hw mode prog (and just call that from xdp_multiprog__detach() and xdp_program__detach_multi() as needed).

Also, be careful with mixing validation and program logic, it makes the code hard to follow. I've pointed out some instances, but keep it in mind when you're fixing things :)

lib/libxdp/libxdp.c
shoracek: Since xdp_multiprog__attach detaches only the specified mode, and now we can have two modes at once, I've changed it so that XDP_MODE_UNSPEC detaches every mode. Since xdp_multiprog__detach detaches entire multiprog.

lib/libxdp/libxdp.c
shoracek: I tried to keep the is_legacy so that it corresponds to the main_prog only, since hw_prog is always "legacy" and there might be some ambiguity if both main_prog and hw_prog would be attached.

lib/libxdp/libxdp.c
tohojo: Right, but see the comment about just not handling hw mode in xdp_multiprog__attach() at all :)

lib/libxdp/libxdp.c
tohojo: Hmm, it does become a bit weird that we can basically have a 'null' value for main_prog now with hw-mode only. But I think is_legacy should mean "doesn't have a dispatcher". Which can then either mean "doesn't have a dispatcher because it's a single program attached in software" in which case main_prog is != null, or it can mean "doesn't have a dispatcher because it's only an hw prog", in which case main_prog == null.

In any case, these semantics should be documented; please add that to both the commit message and the README file :)

shoracek: Also, is is_legacy field really needed? Isn't is_legacy equivalent to first_prog == NULL?

tohojo: Yeah, guess so; if you want to get rid of it, go ahead, but you'd need to change all the internal references to use xdp_multiprog__is_legacy() instead first :)

tohojo (approved):

tohojo:
This is much cleaner, good job!

I found a few fd leaks in the error paths, but rather than have you go through another round I'll just fix them up in separate commits before merging this, so we can all enjoy a nice holiday without this hanging over our heads :)

PR #77, opened 2020-12-11
configure: check endianness for cross-compiling

PolynomialDivision:
The XDP programms are compiled with clang from the hostsystem. If we want to cross-compile, e.g. for big endian, we can just give "-target bpfeb" to clang.

For this we need to check what endianness our cross-compile-toolchain is.

Just a draft, just to check if you are fine with something like this.

### DIFF
--- a/configure
+++ b/configure
@@ -145,6 +145,29 @@ EOF
     fi
 }
 
+check_is_little_endian()
+{
+    cat >$TMPDIR/is_little_endian.c <<EOF
+int main(int argc, char **argv) { // based on https://stackoverflow.com/a/4240014
+    short int number = 0x1;
+    char *numPtr = (char*)&number;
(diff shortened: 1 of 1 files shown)


PolynomialDivision:
Probably I can make this as OpenWrt patch, that checks for
```
ifeq ($(CONFIG_BIG_ENDIAN),y)
```

tohojo:
Why not just make the BPF target itself configurable?

PolynomialDivision:
> Why not just make the BPF target itself configurable?

What do u mean with that?

tohojo:
Add a BPF_TARGET to configure that openwrt can set, and pass that to llvm?

PolynomialDivision:
Closed for https://github.com/xdp-project/xdp-tools/pull/78

PR #78, opened 2020-12-13
configure: add BPF_TARGET for cross-compile

PolynomialDivision:
Although we cross-compile, we use clang from the host. Setting
"-target=bpf" will result in the right format for the host system but
not for the actual target.

Clang has the feature that we can also cross compile directly for some
other architecture, i.e. we can use "bpfeb" and "bpfel".

If we want to use e.g. OpenWrt, then we would like to be able to compile
directly for the given architecture. OpenWrt has the nice feature
that we can check in the Makefile for "ifeq ($(CONFIG_BIG_ENDIAN),y)".
If we combine both, we can set the endianness automatically.

The default behavior remains by setting "-target=bpf".

### DIFF
--- a/configure
+++ b/configure
@@ -16,9 +16,11 @@ check_opts()
     : ${PRODUCTION:=0}
     : ${DYNAMIC_LIBXDP:=0}
     : ${MAX_DISPATCHER_ACTIONS:=10}
+    : ${BPF_TARGET:=bpf}
     echo "PRODUCTION:=${PRODUCTION}" >>$CONFIG
     echo "DYNAMIC_LIBXDP:=${DYNAMIC_LIBXDP}" >>$CONFIG
     echo "MAX_DISPATCHER_ACTIONS:=${MAX_DISPATCHER_ACTIONS}" >>$CONFIG
+    echo "BPF_TARGET:=${BPF_TARGET}" >>$CONFIG
 }
 
 check_toolchain()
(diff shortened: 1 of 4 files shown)


PolynomialDivision:
```
gcc: error: unrecognized command line option ‘-std=gnu17’; did you mean ‘-std=gnu11’?
```
Hmmm. :/ Travis CI does not recognize the new standard.

configure
```diff
@@ -21,6 +21,15 @@ check_opts()
...
 }
 
+check_bpf_target()
+{
+    if [ -z $BPF_TARGET ]; then
+        BPF_TARGET=bpf
+    fi
+    echo "BPF_TARGET:=${BPF_TARGET}" >>$CONFIG
+    echo "$BPF_TARGET"
+}
```
tohojo, line 31: Instead of a whole new function, let's just put this into check_opts and add a `BPF_TARGET?=bpf` to defines.mk.

tohojo (changes requested):
The CI error should go away if you rebase... Sorry about that.

configure
PolynomialDivision: updated

PolynomialDivision: wait something is wrong

PolynomialDivision: `BPF_TARGET?=bpf` Does not work?

tohojo: You're setting it to empty in the configure script. You want '-n' instead of '-z'

PolynomialDivision: I think I have to completely remove it? Then it is working.

tohojo: Hmm, no I think we do want this setable in configure so it persists. But we can put a default in there to avoid setting it to an empty variable. So in check_opts() you can add a default like of the other variables: `: ${BPF_TARGET:=bpf}` 

and then echo it to config.mk. And then move the `BPF_TARGET?=bpf` to the top of defines.mk (along with CFLAGS, before the includes) and I think things should work in all cases...

PolynomialDivision: Changed. :)

tohojo (approved):
Right, LGTM :)

PR #79, opened 2020-12-14
xdpdump: add multi-program capture support

chaudron:


tohojo:
I'll wait with looking through the code in detail until you've fixed the compile
errors, but one UI thing I think we need to fix is the discrepancy with
xdp-loader. I.e., in xdp-loader we output the list of subprograms in a different
format, and to refer to individual subprograms we use --id with the kernel
program ID. Here you're proposing to use name+index instead.

I don't mind adding match by name to xdp-loader as well, but I'm not a huge fan
of the @index syntax. And I think we should be using the same output format when
printing the list of programs (maybe move it into utils.c?).

chaudron:
The compile issue should already be solved, however, you are right we should align the interface output.

For the UI if I use the ID, we do not need the @sign, we should not need it anyway, as the number only would have been enough ;)

Let me rework the patch. I'll move the display function out to a general library function and use it in xdp-loader and xdp-dump. Will change my UI parser to either take a name, or ID.

tohojo:

> The compile issue should already be solved, however, you are right we
> should align the interface output.

Ah, great! I keep forgetting that github doesn't notify on force pushes
to an existing PR.

> For the UI if I use the ID, we do not need the @sign, we should not
> need it anyway, as the number only would have been enough ;)
>
> Let me rework the patch. I'll move the display function out to a
> general library function and use it in xdp-loader and xdp-dump. Will
> change my UI parser to either take a name, or ID.

Sounds good! Could you add the "select subprog by name" option to
xdp-loader while you're at it? :)

chaudron:
Closing this merge req for now, as I'll be pushing stuff to my branch which might get messy ;)

PR #81, opened 2020-12-22
xdpdump: Add multiple program attach support

chaudron:


tohojo:
This LGTM apart from the HW mode issue (#84) - do you wanna fix that as part of this PR or should we merge and fix it afterwards?

chaudron:
Please apply this pr and once the other is also in I’ll fix the issue. Probably next week.

PR #87, opened 2021-01-21
Support incremental attach and detach of subprograms

tohojo:
The kernel now supports re-attaching an freplace program in multiple
places, which means we can finally implement incremental loading of
multiprogs. This involves no change in the API, it just means that
attaching a program to an interface that already has program(s) attached
now will work if the kernel is new enough (5.10+).

This also means we can support detaching a subset of the programs, by the same
mechanism of creating a new dispatcher and swapping it out atomically.

### DIFF
--- a/lib/libxdp/libxdp.c
+++ b/lib/libxdp/libxdp.c
@@ -127,9 +127,10 @@ __printf(2, 3) static void libxdp_print(enum libxdp_print_level level,
 static int xdp_multiprog__attach(struct xdp_multiprog *old_mp,
 				 struct xdp_multiprog *mp,
 				 enum xdp_attach_mode mode);
-static struct xdp_multiprog *xdp_multiprog__generate(struct xdp_program **progs,
-						     size_t num_progs,
-						     int ifindex);
+static struct xdp_multiprog *xdp_multiprog__generate(struct xdp_program **new_progs,
+						     size_t num_new_progs,
+						     int ifindex,
+						     struct xdp_multiprog *old_mp);
 static int xdp_multiprog__pin(struct xdp_multiprog *mp);
 static int xdp_multiprog__unpin(struct xdp_multiprog *mp);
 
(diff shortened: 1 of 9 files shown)


chaudron (approved):

chaudron (approved):

chaudron:
Messed up, reviewed only first two patches :)

tohojo:

> Messed up, reviewed only first two patches :)

So you're not done, or you are done? ;)

lib/libxdp/libxdp.c
```diff
@@ -2119,9 +2221,14 @@ static struct xdp_multiprog *xdp_multiprog__generate(struct xdp_program **progs,
 			goto err;
 	}
 
+	if (num_progs > num_new_progs)
```
chaudron: Think the check should be if (old_mp), so it will cover the case where old_mp->num_links == 0. I think this can happen if the program loaded is a hw offload case only.

lib/libxdp/libxdp.c
```diff
@@ -2119,9 +2221,14 @@ static struct xdp_multiprog *xdp_multiprog__generate(struct xdp_program **progs,
...
+		free(progs);
+
 	return mp;
 
 err:
+	if (num_progs > num_new_progs)
```
chaudron: Same as above

lib/libxdp/libxdp.c
```diff
@@ -2154,18 +2184,33 @@ static struct xdp_multiprog *xdp_multiprog__generate(struct xdp_program **new_pr
...
+				/* remove_new means new_progs is an array of
+				 * programs we should remove from old_mp instead
+				 * of adding them.
+				 */
+				bool found = false;
+
+				for (j = 0; j < num_new_progs; j++)
+					if (new_progs[j]->prog_id == prog->prog_id)
+						found = true;
+				if (found)
```
chaudron, line 2206: Should we at least put a warning in when we do not find the entry?!?

lib/libxdp/libxdp.c
```diff
@@ -2130,9 +2157,10 @@ static int xdp_multiprog__link_prog(struct xdp_multiprog *mp,
 static struct xdp_multiprog *xdp_multiprog__generate(struct xdp_program **new_progs,
```
chaudron, line 2157: The name new_progs is really confusing, not sure what would be better though...  Maybe chg_progs? 

So thinks like ```if (!remove_new)``` look really odd. What are we doing are we not removing_new?! But you mean new_progs is not used as new but as remove...

But I'm ok with new_progs, but maybe this function needs a comment?

lib/libxdp/libxdp.c
```diff
@@ -2143,6 +2171,8 @@ static struct xdp_multiprog *xdp_multiprog__generate(struct xdp_program **new_pr
 	if (!new_progs || !num_new_progs)
 		return ERR_PTR(-EINVAL);
```
chaudron, line 2172: maybe add || (!old_mp && remove_new)?

lib/libxdp/libxdp.c
chaudron: Looks like you fixed it in the next patch :)

xdp-loader/tests/test-xdp-loader.sh
```diff
@@ -13,6 +13,52 @@ test_section()
...
+    ret=$?
+
+    if [ "$ret" -ne "0" ] && echo $output | grep -q "Falling back to loading single prog"; then
+        ret=$SKIPPED_TEST
+    else
+        check_progs_loaded $NS 2
+
+        id=$($XDP_LOADER status $NS | grep xdp_pass | awk '{print $4}')
+        check_run $XDP_LOADER unload $NS --id $id
+        check_progs_loaded $NS 1
```
chaudron, line 56: I think the following needs to be added, so we test the unload all corner case trough IDs:

```
        id=$($XDP_LOADER status $NS | grep xdp_drop | awk '{print $4}')                                                                                                                                                                                         
        check_run $XDP_LOADER unload $NS --id $id                                                                                                                                                                                                               
        check_progs_loaded $NS 0  
```

For now it's failing as check_progs_loaded() needs fixing.

chaudron (changes requested):
Some small changes... Done with t he review :)

tohojo:

> @chaudron requested changes on this pull request.

Updated - I didn't bother to go back and rewrite history, so the diff in
the detach patch is a bit larger than it could have been...

chaudron (approved):

chaudron:
Changes look good to me, merged the pull req.

Issue #84, opened 2021-01-19
xdpdump: handle the case where only an XDP_MODE_HW program is loaded.


chaudron:
If only XDP_ MODE_HW is loaded, fall back to legacy or XDP load based capture.

PR #88, opened 2021-01-25
xdpdump: make xdpdump aware of XDP_MODE_HW

chaudron:
This fixes issue #84

### DIFF
--- a/xdp-dump/xdpdump.c
+++ b/xdp-dump/xdpdump.c
@@ -1603,14 +1603,19 @@ static bool capture_on_interface(struct dumpopt *cfg)
 	struct capture_programs      tgt_progs = {};
 
 	mp = xdp_multiprog__get_from_ifindex(cfg->iface.ifindex);
-	if (IS_ERR_OR_NULL(mp)) {
+	if (IS_ERR_OR_NULL(mp) || xdp_multiprog__main_prog(mp) == NULL) {
+
 		if (!cfg->load_xdp) {
-			pr_warn("WARNING: Specified interface does not have an XDP "
-				"program loaded, capturing\n         in legacy mode!\n");
+			pr_warn("WARNING: Specified interface does not have an XDP program loaded%s,"
+				"\n         capturing in legacy mode!\n",
+				IS_ERR_OR_NULL(mp) ? "" : " in software");
+
+			xdp_multiprog__close(mp);
 			return capture_on_legacy_interface(cfg);
 		}
-		pr_warn("WARNING: Specified interface does not have an XDP program loaded!\n"
-			"         Will load a capture only XDP program!\n");
+		pr_warn("WARNING: Specified interface does not have an XDP program loaded%s!\n"
+			"         Will load a capture only XDP program!\n",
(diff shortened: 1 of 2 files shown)


xdp-dump/xdpdump.c
```diff
@@ -1603,14 +1603,23 @@ static bool capture_on_interface(struct dumpopt *cfg)
 	struct capture_programs      tgt_progs = {};
 
 	mp = xdp_multiprog__get_from_ifindex(cfg->iface.ifindex);
-	if (IS_ERR_OR_NULL(mp)) {
+	if (IS_ERR_OR_NULL(mp) ||
+	    (!IS_ERR_OR_NULL(mp) && xdp_multiprog__main_prog(mp) == NULL)) {
```
tohojo, line 1607: Boolean logic short-circuiting makes that second `!IS_ERR_OR_NULL()` redundant...

xdp-dump/xdpdump.c
```diff
@@ -1603,14 +1603,23 @@ static bool capture_on_interface(struct dumpopt *cfg)
...
+			xdp_multiprog__close(mp);
+			mp = NULL;
+		}
+
 		if (!cfg->load_xdp) {
-			pr_warn("WARNING: Specified interface does not have an XDP "
-				"program loaded, capturing\n         in legacy mode!\n");
+			pr_warn("WARNING: Specified interface does not have an XDP program loaded%s,"
+				"\n         capturing in legacy mode!\n",
+				IS_ERR_OR_NULL(mp) ? "" : " in software");
```
tohojo, line 1617: You just closed `mp` and set it to `NULL` above, so this `IS_ERR_OR_NULL()` will always be true...

xdp-dump/xdpdump.c
```diff
@@ -1603,14 +1603,23 @@ static bool capture_on_interface(struct dumpopt *cfg)
...
 			return capture_on_legacy_interface(cfg);
 		}
-		pr_warn("WARNING: Specified interface does not have an XDP program loaded!\n"
-			"         Will load a capture only XDP program!\n");
+		pr_warn("WARNING: Specified interface does not have an XDP program loaded%s!\n"
+			"         Will load a capture only XDP program!\n",
+			IS_ERR_OR_NULL(mp) ? "" : " in software");
```
tohojo, line 1622: ...as will this one...

tohojo (changes requested):
Some odd logic to fix... :)

xdp-dump/xdpdump.c
chaudron: Messed this up, when I was adding the mp = NULL to fix another error corner case I ran into. Will fix it differently ;)

xdp-dump/xdpdump.c
chaudron: You are right, I normally rely on the compiler to do this. But in this case, it's obvious, will change.

chaudron:
@tohojo new version is available...

tohojo:
Yeah, that's better, thanks :)

PR #90, opened 2021-02-02
Xsk migration to libxdp

magnus-karlsson:
These are the commits for the migration of the xsk functionality from libbpf to libxdp. AFAIK, only samples are missing and need to be added to the bpf-examples repo. At that point, the documentation also need to change to point to that repo for the samples.

### DIFF
--- a/headers/xdp/xsk.h
+++ b/headers/xdp/xsk.h
@@ -1,0 +1,269 @@
+/* SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause) */
+
+/*
+ * AF_XDP user-space access library.
+ *
+ * Copyright(c) 2018 - 2019 Intel Corporation.
+ *
+ * Author(s): Magnus Karlsson <magnus.karlsson@intel.com>
+ */
+
+/* So as not to clash with these functions when they where part of libbpf */
+#ifndef __LIBBPF_XSK_H
+#define __LIBBPF_XSK_H
+
+#include <stdio.h>
+#include <stdint.h>
+#include <bpf/libbpf.h>
+#include <linux/if_xdp.h>
+#include <xdp/libxdp_util.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/* Do not access these members directly. Use the functions below. */
+#define DEFINE_XSK_RING(name) \
+struct name { \
+	__u32 cached_prod; \
+	__u32 cached_cons; \
+	__u32 mask; \
+	__u32 size; \
+	__u32 *producer; \
+	__u32 *consumer; \
+	void *ring; \
+	__u32 *flags; \
+}
+
+DEFINE_XSK_RING(xsk_ring_prod);
+DEFINE_XSK_RING(xsk_ring_cons);
+
+/* For a detailed explanation on the memory barriers associated with the
+ * ring, please take a look at net/xdp/xsk_queue.h in the Linux kernel source tree.
+ */
+
+struct xsk_umem;
+struct xsk_socket;
+
+static inline __u64 *xsk_ring_prod__fill_addr(struct xsk_ring_prod *fill,
+					      __u32 idx)
+{
+	__u64 *addrs = (__u64 *)fill->ring;
+
+	return &addrs[idx & fill->mask];
+}
+
+static inline const __u64 *
+xsk_ring_cons__comp_addr(const struct xsk_ring_cons *comp, __u32 idx)
+{
+	const __u64 *addrs = (const __u64 *)comp->ring;
+
+	return &addrs[idx & comp->mask];
+}
+
+static inline struct xdp_desc *xsk_ring_prod__tx_desc(struct xsk_ring_prod *tx,
+						      __u32 idx)
+{
+	struct xdp_desc *descs = (struct xdp_desc *)tx->ring;
+
+	return &descs[idx & tx->mask];
+}
+
+static inline const struct xdp_desc *
+xsk_ring_cons__rx_desc(const struct xsk_ring_cons *rx, __u32 idx)
+{
+	const struct xdp_desc *descs = (const struct xdp_desc *)rx->ring;
+
+	return &descs[idx & rx->mask];
+}
+
+static inline int xsk_ring_prod__needs_wakeup(const struct xsk_ring_prod *r)
+{
+	return *r->flags & XDP_RING_NEED_WAKEUP;
+}
+
+static inline __u32 xsk_prod_nb_free(struct xsk_ring_prod *r, __u32 nb)
+{
+	__u32 free_entries = r->cached_cons - r->cached_prod;
+
+	if (free_entries >= nb)
+		return free_entries;
+
+	/* Refresh the local tail pointer.
+	 * cached_cons is r->size bigger than the real consumer pointer so
+	 * that this addition can be avoided in the more frequently
+	 * executed code that computs free_entries in the beginning of
+	 * this function. Without this optimization it whould have been
+	 * free_entries = r->cached_prod - r->cached_cons + r->size.
+	 */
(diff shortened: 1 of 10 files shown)


lib/libxdp/Makefile
```diff
@@ -9,7 +9,7 @@ include $(LIB_DIR)/util/util.mk
 OBJDIR ?= .
 SHARED_OBJDIR := $(OBJDIR)/sharedobjs
 STATIC_OBJDIR := $(OBJDIR)/staticobjs
-OBJS := libxdp.o
+OBJS := libxdp.o xsk.o
```
tohojo, line 12: Please make sure the new header files are also added as dependencies so they cause recompiles. At least libxdp_internal.h is not right now...

lib/libxdp/README.org
```diff
@@ -8,17 +8,20 @@
...
+attaching XDP programs to network interfaces and using AF_XDP
+sockets. The library is fairly lightweight and relies on =libbpf= to
+do the heavy lifting for processing eBPF object files etc.
 
-The primary feature that =libxdp= provides on top of =libbpf= is the ability to
-load multiple XDP programs in sequence on a single network device (which is not
-natively supported by the kernel). This support relies on the =freplace=
-functionality in the kernel, which makes it possible to attach an eBPF program
-as a replacement for a global function in another (already loaded) eBPF program.
+One of the two primary featurea that =libxdp= provides on top of
```
tohojo, line 18: s/featurea/features/

lib/libxdp/README.org
```diff
@@ -8,17 +8,20 @@
...
+=libbpf= is the ability to load multiple XDP programs in sequence on a
+single network device (which is not natively supported by the
+kernel). This support relies on the =freplace= functionality in the
+kernel, which makes it possible to attach an eBPF program as a
+replacement for a global function in another (already loaded) eBPF
+program.
```
tohojo, line 24: Also, since this paragraph starts with 'one of the two features', it seems odd that the other one is not mentioned at all; so maybe fix that? :)

lib/libxdp/README.org
```diff
@@ -212,9 +215,165 @@ follows:
 If set, the =LIBXDP_BPFFS= environment variable will override the location of
 =bpffs=, but the =xdp= subdirectory is always used.
 
+* Using AF_XDP sockets
+
+Libxdp implements helper functions for configuring AF_XDP sockets as
+well as reading and writing packets from these sockets. AF_XDP sockets
+can be used to redirect packets to user-space at high throughputs from
```
tohojo, line 222: at high rates?

lib/libxdp/README.org
```diff
@@ -212,9 +215,165 @@ follows:
...
+an XDP program. Note that this functionality used to reside in libbpf,
+but has now been moved over to libxdp as it is a better fit for this
+library. As of the 1.0 release of libbpf, the AF_XDP socket support
+will (or has) be removed and all future development will be performed
```
tohojo, line 226: maybe just drop the (or has) for now, and we can change it once libbpf actually removes the functionality?

lib/libxdp/libxdp.c
```diff
@@ -100,10 +101,8 @@ libxdp_print_fn_t libxdp_set_print(libxdp_print_fn_t fn)
 	return old_print_fn;
 }
 
-#define __printf(a, b) __attribute__((format(printf, a, b)))
-
-__printf(2, 3) static void libxdp_print(enum libxdp_print_level level,
-					const char *format, ...)
+LIBXDP_HIDE_SYMBOL __printf(2, 3) void libxdp_print(enum libxdp_print_level level,
```
tohojo, line 104: This is not actually needed, it's enough to put the attribute in the header file...

lib/libxdp/libxdp.map
```diff
@@ -39,4 +39,13 @@ LIBXDP_1.0.0 {
...
 LIBXDP_1.1.0 {
+	        xsk_setup_xdp_prog;
+		xsk_socket__create;
+		xsk_socket__create_shared;
+		xsk_socket__delete;
+		xsk_socket__fd;
+		xsk_socket__update_xskmap;
+		xsk_umem__create;
+		xsk_umem__delete;
+		xsk_umem__fd;
```
tohojo, line 50: So since we already tagged 1.1.0, these should really be in LIBXDP_1.2.0. I guess since there are no symbols in 1.1.0, you can just rename the section...

lib/libxdp/xsk.c
```diff
@@ -0,0 +1,987 @@
+// SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause)
+
+/*
+ * AF_XDP user-space access library.
+ *
+ * Copyright(c) 2018 - 2019 Intel Corporation.
```
tohojo, line 6: and 2020 and 2021? :)

lib/libxdp/xsk.c
```diff
@@ -0,0 +1,987 @@
...
+		/* call bpf_redirect_map */
+		BPF_EMIT_CALL(BPF_FUNC_redirect_map),
+		/* The jumps are to this instruction */
+		BPF_EXIT_INSN(),
+	};
+	size_t insns_cnt = sizeof(prog) / sizeof(struct bpf_insn);
+
+	prog_fd = bpf_load_program(BPF_PROG_TYPE_XDP, prog, insns_cnt,
+				   "LGPL-2.1 or BSD-2-Clause", 0, log_buf,
+				   log_buf_size);
```
tohojo, line 404: This should really use the libxdp loader functions so it's compatible with multi-prog. However, that won't work with just the raw instructions, `freplace()` needs BTF information. So moving the BPF program into a separate file and compiling it (like we do with xdp-dispatcher.c) is probably easier. You can then use `xdp_program__find_file()` to load it.

The only remaining problem, then, becomes performance: there's an overhead of a single function call + a branch for even a single-program dispatcher. The trouble is we can't really skip it, since it's not possible to change an XDP program to be attached as an freplace program later. This should really be fixed in the kernel, but I'm not sure whether this is really feasible, and it doesn't help us for old kernels anyway.

I think it may be feasible to simply teach libxdp to recognise the AF_XDP stub program and just replace it with a new one when building the dispatcher. This kind of special-casing is a bit ugly, but it's the only way I can see to keep compatibility with multi-prog without incurring performance overhead for AF_XDP. WDYT?

headers/xdp/xsk.h
```diff
@@ -0,0 +1,269 @@
...
+/* Flags for the libbpf_flags field.
+ * We still call this field libbpf_flags for compatibility reasons.
+ */
+#define XSK_LIBBPF_FLAGS__INHIBIT_PROG_LOAD (1 << 0)
+#define XSK_LIBXDP_FLAGS__INHIBIT_PROG_LOAD (1 << 0)
+
+struct xsk_socket_config {
+	__u32 rx_size;
+	__u32 tx_size;
+	__u32 libbpf_flags;
```
tohojo, line 219: We could make this into a union so it can also be accessed as libxdp_flags?

tohojo (changes requested):
Mostly nits, but the one major(ish) issue is compatibility with multi-prog and the rest of the library. See that specific comment...

lib/libxdp/xsk.c
magnus-karlsson: Let me implement this and check how much the overhead is. If it is negligible, then let just do it this way. If not, then we need to think of a solution.

headers/xdp/xsk.h
magnus-karlsson: Will fix this and all the other suggestions you had. Thanks!

lib/libxdp/xsk.c
tohojo: OK, sounds good! :)

tohojo:
Superseded by #92

PR #91, opened 2021-02-12
xdpdump: skip mulitprog selftest if multiprog is not supported

chaudron:


tohojo:
Thanks! :)

PR #92, opened 2021-02-15
Xsk migration with multiprog support

magnus-karlsson:
Pushed a whole new branch. Rewrote the handling of loading and manipulating the default xsk program using your xdp_program__* APIs and fixed all your other comments. Also updated with the latest changes from the Linux repo version.

### DIFF
--- a/lib/libxdp/Makefile
+++ b/lib/libxdp/Makefile
@@ -9,8 +9,8 @@ include $(LIB_DIR)/util/util.mk
 OBJDIR ?= .
 SHARED_OBJDIR := $(OBJDIR)/sharedobjs
 STATIC_OBJDIR := $(OBJDIR)/staticobjs
-OBJS := libxdp.o
-XDP_OBJS := xdp-dispatcher.o
+OBJS := libxdp.o xsk.o
+XDP_OBJS := xdp-dispatcher.o xsk_def_xdp_prog.o
 SHARED_OBJS := $(addprefix $(SHARED_OBJDIR)/,$(OBJS))
 STATIC_OBJS := $(addprefix $(STATIC_OBJDIR)/,$(OBJS))
 STATIC_LIBS := $(OBJDIR)/libxdp.a
@@ -20,11 +20,11 @@ MAN_OBJ := ${MAN_PAGE:.3=.man}
 SHARED_CFLAGS += -fPIC -DSHARED
 LIB_HEADERS := $(wildcard $(HEADER_DIR)/xdp/*.h)
 BPF_HEADERS := $(wildcard $(HEADER_DIR)/bpf/*.h) $(wildcard $(HEADER_DIR)/xdp/*.h)
-EXTRA_LIB_DEPS := $(OBJECT_LIBBPF) $(LIBMK) $(LIB_OBJS) $(LIB_HEADERS) compat.h
+EXTRA_LIB_DEPS := $(OBJECT_LIBBPF) $(LIBMK) $(LIB_OBJS) $(LIB_HEADERS) compat.h libxdp_internal.h
 PC_FILE := $(OBJDIR)/libxdp.pc
 TEMPLATED_SOURCES := xdp-dispatcher.c
 
-CFLAGS += -I$(HEADER_DIR) -I$(LIB_DIR)/util
+CFLAGS += -I$(HEADER_DIR) -I$(LIB_DIR)/util -O0 -g
 BPF_CFLAGS += -I$(HEADER_DIR)
 
 
@@ -87,7 +87,7 @@ $(STATIC_OBJDIR)/%.o: %.c $(EXTRA_LIB_DEPS) | $(STATIC_OBJDIR)
 $(SHARED_OBJDIR)/%.o: %.c $(EXTRA_LIB_DEPS) | $(SHARED_OBJDIR)
 	$(QUIET_CC)$(CC) $(CFLAGS) $(SHARED_CFLAGS) -Wall -I../../headers -c $< -o $@
 
-XDP_IN_SHARED	:= $(SHARED_OBJDIR)/libxdp.o
+XDP_IN_SHARED	:= $(SHARED_OBJDIR)/libxdp.o $(SHARED_OBJDIR)/xsk.o
 
 GLOBAL_SYM_COUNT = $(shell readelf -s --wide $(XDP_IN_SHARED) | \
 			   cut -d "@" -f1 | sed 's/_v[0-9]_[0-9]_[0-9].*//' | \
--- a/lib/libxdp/xsk.c
+++ b/lib/libxdp/xsk.c
@@ -1,0 +1,947 @@
+// SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause)
+
+/*
+ * AF_XDP user-space access library.
+ *
+ * Copyright(c) 2018 - 2021 Intel Corporation.
+ *
+ * Author(s): Magnus Karlsson <magnus.karlsson@intel.com>
+ */
+
+#include <errno.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include <arpa/inet.h>
+#include <bpf/bpf.h>
+#include <bpf/libbpf.h>
+#include <linux/err.h>
+#include <linux/ethtool.h>
+#include <linux/filter.h>
+#include <linux/if_ether.h>
+#include <linux/if_link.h>
+#include <linux/if_packet.h>
+#include <linux/if_xdp.h>
+#include <linux/list.h>
+#include <linux/sockios.h>
+#include <net/if.h>
+#include <sys/ioctl.h>
+#include <sys/mman.h>
+#include <sys/socket.h>
+#include <sys/types.h>
+#include <xdp/xsk.h>
+
+#include "libxdp_internal.h"
+#include "util.h"
+
+#ifndef SOL_XDP
+ #define SOL_XDP 283
+#endif
+
+#ifndef AF_XDP
+ #define AF_XDP 44
+#endif
+
+#ifndef PF_XDP
+ #define PF_XDP AF_XDP
+#endif
+
+struct xsk_umem {
+	struct xsk_ring_prod *fill_save;
+	struct xsk_ring_cons *comp_save;
+	char *umem_area;
+	struct xsk_umem_config config;
+	int fd;
+	int refcount;
+	struct list_head ctx_list;
+};
+
+struct xsk_ctx {
+	struct xsk_ring_prod *fill;
+	struct xsk_ring_cons *comp;
+	__u32 queue_id;
+	struct xsk_umem *umem;
+	int refcount;
+	int ifindex;
+	struct list_head list;
+	struct xdp_program *xdp_prog;
+	int prog_fd;
+	int xsks_map_fd;
+	char ifname[IFNAMSIZ];
+};
+
+struct xsk_socket {
+	struct xsk_ring_cons *rx;
+	struct xsk_ring_prod *tx;
+	__u64 outstanding_tx;
+	struct xsk_ctx *ctx;
+	struct xsk_socket_config config;
+	int fd;
+};
+
+struct xsk_nl_info {
+	bool xdp_prog_attached;
+	int ifindex;
+	int fd;
+};
+
+/* Up until and including Linux 5.3 */
+struct xdp_ring_offset_v1 {
+	__u64 producer;
+	__u64 consumer;
+	__u64 desc;
+};
+
+/* Up until and including Linux 5.3 */
+struct xdp_mmap_offsets_v1 {
+	struct xdp_ring_offset_v1 rx;
+	struct xdp_ring_offset_v1 tx;
+	struct xdp_ring_offset_v1 fr;
+	struct xdp_ring_offset_v1 cr;
+};
+
+int xsk_umem__fd(const struct xsk_umem *umem)
+{
+	return umem ? umem->fd : -EINVAL;
+}
+
+int xsk_socket__fd(const struct xsk_socket *xsk)
+{
+	return xsk ? xsk->fd : -EINVAL;
+}
+
+static bool xsk_page_aligned(void *buffer)
+{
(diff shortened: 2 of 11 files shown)


lib/libxdp/Makefile
```diff
@@ -20,11 +20,11 @@ MAN_OBJ := ${MAN_PAGE:.3=.man}
 SHARED_CFLAGS += -fPIC -DSHARED
 LIB_HEADERS := $(wildcard $(HEADER_DIR)/xdp/*.h)
 BPF_HEADERS := $(wildcard $(HEADER_DIR)/bpf/*.h) $(wildcard $(HEADER_DIR)/xdp/*.h)
-EXTRA_LIB_DEPS := $(OBJECT_LIBBPF) $(LIBMK) $(LIB_OBJS) $(LIB_HEADERS) compat.h
+EXTRA_LIB_DEPS := $(OBJECT_LIBBPF) $(LIBMK) $(LIB_OBJS) $(LIB_HEADERS) compat.h libxdp_internal.h
 PC_FILE := $(OBJDIR)/libxdp.pc
 TEMPLATED_SOURCES := xdp-dispatcher.c
 
-CFLAGS += -I$(HEADER_DIR) -I$(LIB_DIR)/util
+CFLAGS += -I$(HEADER_DIR) -I$(LIB_DIR)/util -O0 -g
```
tohojo, line 27: Why is this needed?

lib/libxdp/xsk.c
```diff
@@ -0,0 +1,947 @@
...
+struct xsk_ctx {
+	struct xsk_ring_prod *fill;
+	struct xsk_ring_cons *comp;
+	__u32 queue_id;
+	struct xsk_umem *umem;
+	int refcount;
+	int ifindex;
+	struct list_head list;
+	struct xdp_program *xdp_prog;
+	int prog_fd;
```
tohojo, line 68: Why have both? It's possible to build an xdp_program from an fd (xdp_program__from_fd()), so just keeping the xdp_prog pointer saves branching (see comments below).

lib/libxdp/xsk.c
```diff
@@ -0,0 +1,947 @@
...
+
+static void xsk_delete_map_entry(struct xsk_socket *xsk)
+{
+	struct xsk_ctx *ctx = xsk->ctx;
+
+	bpf_map_delete_elem(ctx->xsks_map_fd, &ctx->queue_id);
+	close(ctx->xsks_map_fd);
+}
+
+static int xsk_lookup_bpf_map(struct xsk_socket *xsk)
```
tohojo, line 412: Looking at this function I wonder if it would make sense to have libxdp parse the list of maps in `xdp_program__fill_from_fd()` so they are always available to callers? A bit orthogonal to this patch, but may be generally useful? WDYT?

lib/libxdp/xsk.c
```diff
@@ -0,0 +1,947 @@
...
+		if (fd < 0)
+			continue;
+
+		err = bpf_obj_get_info_by_fd(fd, &map_info, &map_len);
+		if (err) {
+			close(fd);
+			continue;
+		}
+
+		if (!strcmp(map_info.name, "xsks_map")) {
```
tohojo, line 452: I am a little worried about forward compatibility here; what if we need to change the layout of xsks_map in the future? I think we should at least checking that the key and value size matches what we expect...

lib/libxdp/xsk.c
```diff
@@ -0,0 +1,947 @@
...
+	if (!multi_prog)
+		return -1;
+
+	while ((prog = xdp_multiprog__next_prog(prog, multi_prog)))
+		if (!strcmp(xdp_program__name(prog), "xsk_def_prog"))
+			break;
+
+	if (!prog)
+		return -1;
+	return xdp_program__fd(prog);
```
tohojo, line 533: This leaks the multi_prog reference... Given what I said above re: just keeping the xdp_program reference, maybe the right thing here is to have this function return an xdp_program pointer, and expose xdp_program__clone() so it can still close the multiprog...

lib/libxdp/xsk.c
```diff
@@ -0,0 +1,947 @@
...
+}
+
+static int __xsk_setup_xdp_prog(struct xsk_socket *xsk, int *xsks_map_fd)
+{
+	struct xsk_ctx *ctx = xsk->ctx;
+	int prog_fd, err;
+
+	prog_fd = xsk_lookup_program(xsk);
```
tohojo, line 541: This could just be `ctx->xdp_prog = xsk_lookup_program()`

lib/libxdp/xsk.c
```diff
@@ -0,0 +1,947 @@
...
+			return -errno;
+		err = xsk_lookup_bpf_map(xsk);
+		if (err) {
+			close(ctx->prog_fd);
+			return err;
+		}
+	}
+
+	if (xsk->rx) {
+		err = xsk_add_map_entry(xsk);
```
tohojo, line 569: So if no XDP program is loaded, this will attach the program before the map is populated. Isn't it better to populate the map first so we're reasonable sure everything is setup before it starts processing packets? Seeing as program attach can have side effects, we'd want to avoid doing attach-then-detach if possible?

lib/libxdp/xsk.c
```diff
@@ -0,0 +1,947 @@
...
+	if (xsks_map_fd)
+		*xsks_map_fd = ctx->xsks_map_fd;
+
+	return 0;
+
+err_detach:
+	xdp_program__detach(ctx->xdp_prog, ctx->ifindex,
+			    xsk_convert_xdp_flags(xsk->config.xdp_flags), 0);
+err_prog_load:
+	xdp_program__close(ctx->xdp_prog);
```
tohojo, line 588: This should clear the pointer as well.

lib/libxdp/xsk.c
```diff
@@ -0,0 +1,947 @@
...
+
+	err = bind(xsk->fd, (struct sockaddr *)&sxdp, sizeof(sxdp));
+	if (err) {
+		err = -errno;
+		goto out_mmap_tx;
+	}
+
+	ctx->prog_fd = -1;
+
+	if (!(xsk->config.libbpf_flags & XSK_LIBBPF_FLAGS__INHIBIT_PROG_LOAD)) {
```
tohojo, line 854: This flag is not mentioned in the documentation :)

lib/libxdp/xsk.c
```diff
@@ -0,0 +1,947 @@
...
+		return;
+
+	ctx = xsk->ctx;
+	umem = ctx->umem;
+	if (ctx->prog_fd != -1) {
+		xsk_delete_map_entry(xsk);
+		if (ctx->xdp_prog)
+			xdp_program__close(ctx->xdp_prog);
+		else
+			close(ctx->prog_fd);
```
tohojo, line 923: If we just always store ctx->xdp_prog this disambiguation is not needed anymore...

lib/libxdp/xsk.c
```diff
@@ -0,0 +1,947 @@
...
+	struct xdp_multiprog *multi_prog;
+	struct xdp_program *prog = NULL;
+	struct xsk_ctx *ctx = xsk->ctx;
+
+	multi_prog = xdp_multiprog__get_from_ifindex(ctx->ifindex);
+	if (!multi_prog)
+		return -1;
+
+	while ((prog = xdp_multiprog__next_prog(prog, multi_prog)))
+		if (!strcmp(xdp_program__name(prog), "xsk_def_prog"))
```
tohojo, line 528: This is also a potential forward-compatibility concern. If we ever need to change the default program in an incompatible way, we'd get subtle breakage when mixing library versions. I put a version check into the dispatcher program for just this reason - see `check_dispatcher_version()`, and feel free to just copy that (or better, refactor it to be reusable for this :))

lib/libxdp/xsk.c
```diff
@@ -0,0 +1,947 @@
...
+	free(xsk);
+}
+
+int xsk_socket__update_xskmap(struct xsk_socket *xsk, int fd)
+{
+	xsk->ctx->xsks_map_fd = fd;
+	return xsk_add_map_entry(xsk);
+}
+
+int xsk_setup_xdp_prog(int ifindex, int *xsks_map_fd)
```
tohojo, line 684: How is the application supposed to tear down the XDP prog again? Or rather, how is an application supposed to do this without racing against another program trying to install itself? Since we abstract away the creation and attachment of the XDP program, I think we need to handle the race condition as well...

We do already have a locking mechanism (in xdp_lock_acquire/release()) we could use, but a bit of care is needed to prevent deadlocking with outselves (xdp_program__attach() will grab the lock). So may need a bit more knowledge of the internals of the rest of the library... WDYT?

tohojo (changes requested):
A few comments; biggest issue is probably the race condition / locking thing.

Also, did you have a chance to test if using the XDP dispatcher has any measurable performance overhead? :)

lib/libxdp/Makefile
magnus-karlsson: It is not ;-). Just garbage left after a debug session.

lib/libxdp/xsk.c
magnus-karlsson: Might make sense, but if this is never performance critical it might not make sense to waste the memory. In my case, performance is not an issue.

lib/libxdp/xsk.c
magnus-karlsson: Good point. Will fix.

lib/libxdp/xsk.c
magnus-karlsson: I will make a stab at just using the program pointer and make sure to destroy it when necessary.

lib/libxdp/xsk.c
magnus-karlsson: Will fix.

lib/libxdp/xsk.c
magnus-karlsson: Good point. Will steal your code ;-).

lib/libxdp/xsk.c
magnus-karlsson: I have kept all locking outside the library by design. I think this is best left up to the user to fix with a synchronization primitive of its choosing which might even be none with certain designs.

lib/libxdp/xsk.c
magnus-karlsson: Oops! Will add.

lib/libxdp/xsk.c
magnus-karlsson: I can reverse the order. Actually saves some processing time during setup.

magnus-karlsson:
> A few comments; biggest issue is probably the race condition / locking thing.

I think we should not implement any locking as discussed above. Let the user do this if it needs it.

> Also, did you have a chance to test if using the XDP dispatcher has any measurable performance overhead? :)

No, I have not done this yet. I take an action to do it.

lib/libxdp/xsk.c
tohojo: For synchronisation inside an application I agree that we shouldn't impose any particular locking. But this also has to do with synchronisation *between* applications, that don't necessarily have any way of coordinating, other than libxdp. So I think solving this in a race-free way is in scope; it's basically the same issue as Maciej's patch[0] is addressing in the libbpf version of this code.

[0] https://lore.kernel.org/bpf/3a625c8b-6dc0-3933-25e5-f747197ae1f4@intel.com/T/#t

magnus-karlsson:
On Mon, Feb 15, 2021 at 6:02 PM Toke Høiland-Jørgensen <
notifications@github.com> wrote:

> *@tohojo* commented on this pull request.
> ------------------------------
>
> In lib/libxdp/xsk.c
> <https://github.com/xdp-project/xdp-tools/pull/92#discussion_r576327633>:
>
> > +	return ctx;
> +}
> +
> +static void xsk_destroy_xsk_struct(struct xsk_socket *xsk)
> +{
> +	free(xsk->ctx);
> +	free(xsk);
> +}
> +
> +int xsk_socket__update_xskmap(struct xsk_socket *xsk, int fd)
> +{
> +	xsk->ctx->xsks_map_fd = fd;
> +	return xsk_add_map_entry(xsk);
> +}
> +
> +int xsk_setup_xdp_prog(int ifindex, int *xsks_map_fd)
>
> For synchronisation inside an application I agree that we shouldn't impose
> any particular locking. But this also has to do with synchronisation
> *between* applications, that don't necessarily have any way of
> coordinating, other than libxdp. So I think solving this in a race-free way
> is in scope; it's basically the same issue as Maciej's patch[0] is
> addressing in the libbpf version of this code.
>
> [0]
> https://lore.kernel.org/bpf/3a625c8b-6dc0-3933-25e5-f747197ae1f4@intel.com/T/#t
>

This was never in scope for the libbpf implementation, but I agree that we
should provide this with libxdp if it is successfully going to serve as
this coordination point that the original code just did not offer. But I
consider this outside of this patch set. It is something we should add as a
follow up IMHO.

/Magnus

tohojo:
Magnus Karlsson <notifications@github.com> writes:

> This was never in scope for the libbpf implementation, but I agree that we
> should provide this with libxdp if it is successfully going to serve as
> this coordination point that the original code just did not offer. But I
> consider this outside of this patch set. It is something we should add as a
> follow up IMHO.

OK, fair enough, we can defer this to after merge. I opened an issue so
we don't forget:

https://github.com/xdp-project/xdp-tools/issues/93

magnus-karlsson:
Pushed a new version/branch

tohojo:
Magnus Karlsson ***@***.***> writes:

> Pushed a new version/branch

You know you can just force-push to the same source branch and github
will update the PR, right? :)

magnus-karlsson:
On Tue, Mar 23, 2021 at 11:42 AM Toke Høiland-Jørgensen
***@***.***> wrote:
>
> Magnus Karlsson ***@***.***> writes:
>
> > Pushed a new version/branch
>
> You know you can just force-push to the same source branch and github
> will update the PR, right? :)

Yes :-). Just did not know what you preferred. In this case, a force
push would have been warranted as it is largely a rewrite from
scratch. Threw away my changes and started with a clean libbpf, since
I was not happy with the result last time. Will force push new
versions on top of this one.

PR #95, opened 2021-03-01
Run selftests in CI

shoracek:
Introduces selftests into Travis CI.
As a kernel used while testing a pre-built one or one to be build from scratch can be specified in the configuration file.
Currently only x86 arch is supported.


tohojo:
Very cool! Does the KERNEL_VERSION==latest setting work? I'm not sure if it's a good idea to have a test against an always-changing kernel version like that, at least not as a CI item for xdp-tools (as that can break the test because of an unrelated change in the kernel). Is there a way to mark a job as "optional" (so that it will still run, but the test suite will not fail if that job does)?

Also, what about older kernels? Do those work?

shoracek:
Maybe add a variable for commit or just remove the building completely. For optional jobs it looks like there's a "allow_failures" key for that.
I have tried it with a few older Fedora kernels and the tests start passing with version 5.7.0 (excluding test_load_incremental and test_multi_prog). Before this version every test fails.

tohojo:
Right. Well, I'll merge this now, and we can play around with adding more versions on top. Any idea why the tests fail on old kernels? Some stuff won't work, obviously (like xdpdump), but we should be able to at least load programs with the fallback mode... so maybe the tests need to be fixed?

PR #96, opened 2021-03-04
xdp: Prevent double close of dispatcher prog fd

kkdwivedi:
The dispatcher prog is opened using the open_file path, then loaded onto
the interface, where down the call chain xdp_program__load is called.
This function directly invokes the xdp_program__fill_from_fd helper, but
fails to take into account that it consumes the fd that is passed in (as
it is stashed in the xdp_prog structure).

In all other call sites (wrapped using xdp_program__from_fd), the fd
being consumed is acquired right before it is passed in, so the
ownership transfer works as expected.

All in all, this leads to release paths of the xdp_prog and the bpf_prog
both trying to call close on the same fd, where the xdp_program__close
wins due to it being early in the call chain, and the
bpf_object__unload -> bpf_program__unload's close call gets -EBADF.

Fix this by duplicating the descriptor before passing to the helper.

### DIFF
--- a/lib/libxdp/libxdp.c
+++ b/lib/libxdp/libxdp.c
@@ -1168,7 +1168,15 @@ static int xdp_program__load(struct xdp_program *prog)
 	pr_debug("Loaded XDP program %s, got fd %d\n",
 		 xdp_program__name(prog), bpf_program__fd(prog->bpf_prog));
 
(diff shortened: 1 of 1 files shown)


tohojo (approved):
Looks good, thanks! :)

PR #97, opened 2021-03-06
xsk: export inline helpers as symbols in libxdp

kkdwivedi:
This helps people writing language bindings to not have to rewrite C
wrappers for inline functions in the headers. We force inline the
definition from the header for C and C++ consumers, but also export a
symbol in the library for others. This keeps the performance advantages.

Also see
https://lore.kernel.org/bpf/CAJ8uoz0QqR97qEYYK=VVCE9A=V=k2tKnH6wNM48jeak2RAmL0A@mail.gmail.com/
for some context.

extern inline is used as it's slightly better since it warns when an
inline definition is missing.

/cc @dansiemon
Please let me know if this works for the Rust case.


tohojo:
@magnus-karlsson @bjoto  - what do you think about this approach?

magnus-karlsson:
Cool! And this would work with Rust too? If so, let us adopt this.

Thanks: Magnus

tohojo:
Let's see if we can get @dansiemon or @alexforster to weigh in on that :)

dansiemon:
Thanks for looking at this. I'll do quick test in the next couple days.

dansiemon:
The current project structure makes creating a binding a bit more complicated than libbpf (binding two libs etc). I ran into some problems (probably surmountable with more time) so instead I tried manually making the same changes to libbpf and rebuilding the binding. This does not appear to work. The Rust cbindgen tool didn't find those symbols.

I'm away this week so I won't make much progress in the next few days but I will get back to this.

kkdwivedi:
Thanks for testing this.

I just want to point out that you would need to use generate_inline_functions(true) with bindgen in your build.rs file (or --generate-inline-functions with the tool) for it to be able to process the inline definitions. This is supposed to be used in combination with `-fkeep-inline-definitions` and similar so that there is something to link to, which we already fulfill in a different way. I don't know enough about cbindgen to comment though.

dansiemon:
I tried this with the `generate-inline-functions` flag to cbindgen and the manually changed version of `libbpf` and it found the function and generated the binding. So this looks good.

I haven't yet tried using it but it should be fine.

At some point I will create a binding to `libxdp`. Is the plan for this to stay in `xdp-tools` or move out to its own repo?

tohojo:
> I tried this with the `generate-inline-functions` flag to cbindgen and the manually changed version of `libbpf` and it found the function and generated the binding. So this looks good.

Great!

> I haven't yet tried using it but it should be fine.
> 
> At some point I will create a binding to `libxdp`. Is the plan for this to stay in `xdp-tools` or move out to its own repo?

Keeping it in this repo is convenient for packaging the command line tools along with the library, so unless (or until) there's a compelling reason to move it that outweighs those benefits it's staying :)

tohojo:
@magnus-karlsson  could you please fold this into #102 ? Saves a bit of rebasing...

magnus-karlsson:
Will fix.

magnus-karlsson:
On Wed, Apr 21, 2021 at 11:08 PM Toke Høiland-Jørgensen
***@***.***> wrote:
>
> @magnus-karlsson could you please fold this into #102 ? Saves a bit of rebasing...

Pushed.

/Magnus

magnus-karlsson:
On Thu, Apr 22, 2021 at 11:06 AM Magnus Karlsson
***@***.***> wrote:
>
> On Wed, Apr 21, 2021 at 11:08 PM Toke Høiland-Jørgensen
> ***@***.***> wrote:
> >
> > @magnus-karlsson could you please fold this into #102 ? Saves a bit of rebasing...
>
> Pushed.

The Travis CI check failed for some reason. Complains about scapy
errors and some missing kernel modules.

shoracek:
> The Travis CI check failed for some reason. Complains about scapy errors and some missing kernel modules.

It should be fixed now.

tohojo:
Tried retriggering from the Travis UI, which doesn't seem to be possible, so I think you need to do another push, @magnus-karlsson  :)

alexforster:
This will help us out on the Rust side, but only if the changes are applied to libbpf proper. Is that the plan, and if so will there be a new release to follow?

magnus-karlsson:
Štěpán,

Could you apply your solution to libbpf too and submit a patch to the
kernel? I could then ack it.

Thanks: Magnus

magnus-karlsson:
Pushed again.

shoracek:
> Štěpán, Could you apply your solution to libbpf too and submit a patch to the kernel? I could then ack it. Thanks: Magnus

I'm not sure what solution you mean. I fixed the failing pipeline by removing a bug from a library dependency. But I am not sure how that would be applicable to libbpf, as it does not use that library.

magnus-karlsson:
On Mon, Apr 26, 2021 at 8:29 PM Štěpán Horáček ***@***.***>
wrote:

> Štěpán, Could you apply your solution to libbpf too and submit a patch to
> the kernel? I could then ack it. Thanks: Magnus
>
> I'm not sure what solution you mean. I fixed the failing pipeline by
> removing a bug from a library dependency. But I am not sure how that would
> be applicable to libbpf, as it does not use that library.
>

Sorry Štěpán, wrong person. I meant Kumar who made the fix for Rust. Hope I
got the right person this time ;-).

/Magnus

kkdwivedi:
Sent: https://lore.kernel.org/bpf/20210426202240.518961-1-memxor@gmail.com/

glenn-wang:
Is the lipxdp-sys for rust users is completed ?

glenn-wang:
find a crate 

https://github.com/chenhengqi/libxdp-sys is ok ?

tohojo:
glenn-wang ***@***.***> writes:

> find a crate 
>
> https://github.com/chenhengqi/libxdp-sys is ok ?

No idea; first I've heard of it...

PR #98, opened 2021-03-08
Add xdp_clean_references

kkdwivedi:
This PR implements a clean up function to check if a ifindex's dispatcher directory is stale, and removes the pinned directory. Optionally, passing ifindex 0 makes the helper walk the entire XDP bpffs directory, looking for stale entries.

This is now exposed to the user through the `xdp-loader collect [ifname]` command.

### DIFF
--- a/headers/xdp/libxdp.h
+++ b/headers/xdp/libxdp.h
@@ -41,6 +41,7 @@ struct xdp_multiprog;
 
 long libxdp_get_error(const void *ptr);
 int libxdp_strerror(int err, char *buf, size_t size);
+int libxdp_clean_references(int ifindex);
 
 
 struct xdp_program *xdp_program__from_bpf_obj(struct bpf_object *obj,
--- a/lib/libxdp/libxdp.c
+++ b/lib/libxdp/libxdp.c
@@ -16,7 +16,9 @@
 #include <sys/mman.h>
 #include <sys/file.h>
 #include <sys/vfs.h>
+#include <sys/types.h>
 #include <fcntl.h>
+#include <inttypes.h>
 #include <dirent.h>
 
 #include <linux/err.h> /* ERR_PTR */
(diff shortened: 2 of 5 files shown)


tohojo:
Hmm, how is the caller supposed to obtain a struct xdp_multiprog reference to the program if it's not attached to an interface? And also, what is the caller supposed to do with the information returned by xdp_multiprog__defunct()?

Since the objective is garbage collecting the pinned files, I think maybe we just need a `libxdp_clean_references(ifindex)` type interface? Where it's also possible to pass in 0 as an ifindex to walk all the pinned programs in the pin directory and clean them. This could then be tied into an xdp-loader action...

kkdwivedi:
> Since the objective is garbage collecting the pinned files, I think maybe we just need a `libxdp_clean_references(ifindex)` type interface? Where it's also possible to pass in 0 as an ifindex to walk all the pinned programs in the pin directory and clean them. This could then be tied into an xdp-loader action...

Sounds good, I'll push with this change and the `xdp-loader collect` part.

kkdwivedi:
This should be ready for review now.

headers/xdp/libxdp.h
```diff
@@ -99,5 +99,6 @@ enum xdp_attach_mode xdp_multiprog__attach_mode(const struct xdp_multiprog *mp);
 struct xdp_program *xdp_multiprog__main_prog(const struct xdp_multiprog *mp);
 struct xdp_program *xdp_multiprog__hw_prog(const struct xdp_multiprog *mp);
 bool xdp_multiprog__is_legacy(const struct xdp_multiprog *mp);
+int xdp_clean_references(int ifindex);
```
tohojo, line 102: Let's prefix this with libxdp_ instead of xdp_ and put it at the top with libxdp_get_error() and libxdp_strerror()

lib/libxdp/libxdp.c
```diff
@@ -1228,7 +1230,7 @@ static int xdp_multiprog__main_fd(struct xdp_multiprog *mp)
 	return mp->main_prog->prog_fd;
 }
 
-static __u32 xdp_multiprog__main_id(struct xdp_multiprog *mp)
+static __u32 xdp_multiprog__main_id(const struct xdp_multiprog *mp)
```
tohojo, line 1233: Please put these in a separate commit

lib/libxdp/libxdp.c
```diff
@@ -2570,3 +2572,139 @@ bool xdp_multiprog__is_legacy(const struct xdp_multiprog *mp)
 {
 	return !!(mp && mp->is_legacy);
 }
+
+static void xdp_get_ifindex_prog_id(int ifindex, __u32 *prog_id)
```
tohojo, line 2576: Why not just return the prog_id instead of having a point in a void function?

lib/libxdp/libxdp.c
```diff
@@ -2570,3 +2572,139 @@ bool xdp_multiprog__is_legacy(const struct xdp_multiprog *mp)
...
+		default:;
+		}
+	}
+}
+
+static int remove_pin_dir(const char *subdir)
+{
+	char prog_path[PATH_MAX], pin_path[PATH_MAX];
+	int err;
+	DIR *d = NULL;
```
tohojo, line 2602: We generally try to use the 'reverse x-mas tree' variable ordering from the kernel netdev tradition

lib/libxdp/libxdp.c
```diff
@@ -2570,3 +2572,139 @@ bool xdp_multiprog__is_legacy(const struct xdp_multiprog *mp)
...
+	return err;
+}
+
+int xdp_clean_references(int ifindex)
+{
+	int err, lock_fd;
+	__u32 dir_prog_id;
+	char fmt[PATH_MAX];
+	__u32 prog_id = 0;
+	DIR *d = NULL;
```
tohojo, line 2653: Same as above - reverse x-mas tree, please. Also, just put the two __u32s on the same line

lib/libxdp/libxdp.c
```diff
@@ -2570,3 +2572,139 @@ bool xdp_multiprog__is_legacy(const struct xdp_multiprog *mp)
...
+		return lock_fd;
+
+	d = opendir(dir);
+	if (!d) {
+		err = -errno;
+		pr_debug("Failed to open bpffs directory: %s\n", strerror(-err));
+		goto err;
+	}
+
+	if (!ifindex) {
```
tohojo, line 2670: Instead of this...

lib/libxdp/libxdp.c
```diff
@@ -2570,3 +2572,139 @@ bool xdp_multiprog__is_legacy(const struct xdp_multiprog *mp)
...
+		for (struct dirent *dent = readdir(d); dent; dent = readdir(d)) {
+			if (dent->d_type != DT_DIR)
+				continue;
+
+			if (sscanf(dent->d_name, "dispatch-%d-%"PRIu32"", &ifindex, &dir_prog_id) != 2)
+				continue;
+
```
tohojo, line 2677: ...just add a check if the ifindex is set and matches here (and continue if it doesn't)...

lib/libxdp/libxdp.c
```diff
@@ -2570,3 +2572,139 @@ bool xdp_multiprog__is_legacy(const struct xdp_multiprog *mp)
...
+			xdp_get_ifindex_prog_id(ifindex, &prog_id);
+			if (!prog_id || prog_id != dir_prog_id) {
+				err = remove_pin_dir(dent->d_name);
+				if (err)
+					goto err;
+			}
+			prog_id = 0;
+		}
+	} else {
```
tohojo, line 2686: ... and you can avoid the duplication here.

lib/libxdp/libxdp.c
```diff
@@ -2570,3 +2572,139 @@ bool xdp_multiprog__is_legacy(const struct xdp_multiprog *mp)
...
+			if (!prog_id || prog_id != dir_prog_id) {
+				err = remove_pin_dir(dent->d_name);
+				if (err)
+					goto err;
+			}
+		}
+	}
+err:
+	xdp_lock_release(lock_fd);
+	closedir(d);
```
tohojo, line 2708: order these as the reverse of what you did at the top?

xdp-loader/xdp-loader.c
```diff
@@ -294,6 +313,7 @@ int do_help(__unused const void *cfg, __unused const char *pin_root_path)
 		"       load        - load an XDP program on an interface\n"
 		"       unload      - unload an XDP program from an interface\n"
 		"       status      - show current XDP program status\n"
+		"       collect     - clean up dead links in XDP bpffs directory\n"
```
tohojo, line 316: Let's just name the action 'clean' as well - and don't forget to update the man page (README.org) as well

tohojo (changes requested):

kkdwivedi:
Pushed, PTAL.

lib/libxdp/libxdp.c
```diff
@@ -2570,3 +2572,124 @@ bool xdp_multiprog__is_legacy(const struct xdp_multiprog *mp)
...
+			   &path_ifindex, &dir_prog_id) != 2)
+			continue;
+
+		/* If ifindex is set, skip this dir if it doesn't match */
+		if (ifindex && path_ifindex != ifindex)
+			continue;
+
+		prog_id = xdp_get_ifindex_prog_id(path_ifindex);
+		if (!prog_id || prog_id != dir_prog_id) {
+			err = remove_pin_dir(dent->d_name);
```
tohojo, line 2686: A debug line here saying what's going on would be good (including whether the prog ID changed or if nothing is attached now)...

xdp-loader/README.org
```diff
@@ -114,6 +115,20 @@ Enable debug logging. Specify twice for even more verbosity.
 ** -h, --help
 Display a summary of the available options
 
+* The CLEAN command
```
tohojo, line 118: For consistency with the other commands, please add an explicit "the syntax for the =clean= command is" bit that includes the [ifindex] parameter.

xdp-loader/xdp-loader.c
```diff
@@ -285,6 +285,25 @@ int do_status(const void *cfg, __unused const char *pin_root_path)
...
+		      .positional = true, .metavar = "[ifname]",
+		      .help = "Clean up detached program links for [ifname] (default all interfaces)"),
+	END_OPTIONS
+};
+
+int do_clean(const void *cfg, __unused const char *pin_root_path)
+{
+	const struct cleanopt *opt = cfg;
+
+	printf("Cleaning up detached XDP program links...\n");
```
tohojo, line 303: This should include whether we're cleaning all interfaces or just the one. Also, ending with an ellipsis implies there will be more output coming, so just end with a period.

tohojo (changes requested):
Functionality wise this is fine now, just a few more nits on the output and documentation...

kkdwivedi:
Pushed again, PTAL.

tohojo:
Fixed up the log message and merged - thanks! :)

PR #99, opened 2021-03-09
async_close

kkdwivedi:
This PR implements the async_close helper, needed in cases where close may block when closing bpf_link fds.

I have relaxed the barrier on the CQE read side, though it would be nice to get some more eyes on that part of the code to see if it's correct.

### DIFF
--- a/lib/libxdp/libxdp.c
+++ b/lib/libxdp/libxdp.c
@@ -28,6 +28,7 @@
 #include <xdp/libxdp.h>
 #include <xdp/prog_dispatcher.h>
 
+#include "async_close.h"
 #include "compat.h"
 
 #define XDP_RUN_CONFIG_SEC ".xdp_run_config"
@@ -801,7 +802,7 @@ void xdp_program__close(struct xdp_program *xdp_prog)
 		return;
 
 	if (xdp_prog->link_fd >= 0)
-		close(xdp_prog->link_fd);
+		async_close(xdp_prog->link_fd);
 	if (xdp_prog->prog_fd >= 0)
 		close(xdp_prog->prog_fd);
 
@@ -1971,7 +1972,7 @@ static int xdp_multiprog__check_compat(struct xdp_multiprog *mp)
 			 strerror(-err));
 		goto out;
 	}
-	close(lfd);
+	async_close(lfd);
 	mp->checked_compat = true;
 out:
 	xdp_program__close(test_prog);
--- a/lib/util/async_close.h
+++ b/lib/util/async_close.h
@@ -1,0 +1,159 @@
+/* SPDX-License-Identifier: GPL-2.0 */
+
+/* This is a temporary workaround for the blocking close for bpf_link fds. We
+ * try a best effort asynchronous close operation, failing which we fall back to
+ * the normal close(2).
+ *
+ * We only submit SQEs, but don't really consume the result, but still advance
+ * so as to not breach our CQE overflow limits and get -EBUSY back when doing
+ * io_uring_enter.
+ *
+ * At initialization time, the constructor checks if the IORING_OP_CLOSE opcode
+ * is supported, and sets a bool indicating the status. This allows us to avoid
+ * setting up and tearing down the mappings each time async_close is called.
+ */
+
+#ifndef __ASYNC_CLOSE_H
+#define __ASYNC_CLOSE_H
+
+#include <linux/io_uring.h>
+#include <string.h>
+#include <stdbool.h>
+#include <sys/mman.h>
+#include <sys/syscall.h>
+#include <unistd.h>
+#include <errno.h>
+
+#define XDP_URING_OP_CLOSE 19
+
+static bool avail_op_close = false;
+
+static inline int __io_uring_setup(unsigned int entries,
+				   struct io_uring_params *p)
+{
+	return (int)syscall(__NR_io_uring_setup, entries, p);
+}
+
+static inline int __io_uring_enter(int ring_fd, unsigned int to_submit,
+				   unsigned int min_complete,
+				   unsigned int flags)
+{
+	return (int)syscall(__NR_io_uring_enter, ring_fd, to_submit,
+			    min_complete, flags, NULL, 0);
+}
+
+static inline void submit_close_sqe(struct io_uring_sqe *sqe_ptr,
+				    unsigned int *sq_array,
+				    unsigned int *sqe_ring_tail,
+				    unsigned int *sq_ring_mask, int fd)
+{
+	unsigned int tail, index;
+
+	/* There's always a free SQE slot, as they're consumed on enter */
+	tail = __atomic_load_n(sqe_ring_tail, __ATOMIC_RELAXED);
+	index = tail & *sq_ring_mask;
+	sqe_ptr[index].fd = fd;
+	sqe_ptr[index].opcode = XDP_URING_OP_CLOSE;
+	sq_array[index] = index;
+	__atomic_store_n(sqe_ring_tail, tail + 1, __ATOMIC_RELEASE);
+}
+
+static inline void consume_close_cqe(unsigned int *cqe_ring_head)
+{
+	/* We don't really care about the result, so just advance head.  This
+	 * also means the update to head can be a relaxed store, as we don't
+	 * really read anything, so it is fine. Also, a future submit will
+	 * ensure this is visible before we enter the kernel again.
+	 */
+	unsigned int head = __atomic_load_n(cqe_ring_head, __ATOMIC_RELAXED);
+	__atomic_store_n(cqe_ring_head, head + 1, __ATOMIC_RELAXED);
+}
+
+static int __async_close(int close_fd, bool feature)
+{
+	unsigned int *sq_ring_mask, *sq_array, *cqe_ring_head, *sqe_ring_tail;
+	size_t sq_len, sqe_len, cqe_len;
+	struct io_uring_sqe *sqe_ptr;
+	struct io_uring_params p;
+	void *sq_ptr, *cqe_ptr;
+	int fd;
+	int r = -1;
+
+	memset(&p, 0, sizeof(p));
+	fd = __io_uring_setup(2, &p);
+	if (fd < 0)
+		return fd;
+
+	sq_len = p.sq_off.array + p.sq_entries * sizeof(unsigned int);
+	cqe_len = p.cq_off.cqes + p.cq_entries * sizeof(struct io_uring_cqe);
+
+	sq_ptr = mmap(NULL, sq_len, PROT_READ | PROT_WRITE,
+		      MAP_SHARED | MAP_POPULATE, fd, IORING_OFF_SQ_RING);
+	if (sq_ptr == MAP_FAILED)
+		goto close_fd;
+
+	cqe_ptr = mmap(NULL, cqe_len, PROT_READ | PROT_WRITE,
+		       MAP_SHARED | MAP_POPULATE, fd, IORING_OFF_CQ_RING);
+	if (cqe_ptr == MAP_FAILED)
+		goto unmap_sq_ptr;
+
+	cqe_ring_head = cqe_ptr + p.cq_off.head;
+
+	sqe_ring_tail = sq_ptr + p.sq_off.tail;
+	sq_ring_mask = sq_ptr + p.sq_off.ring_mask;
+	sq_array = sq_ptr + p.sq_off.array;
+
+	sqe_len = p.sq_entries * sizeof(struct io_uring_sqe);
+	sqe_ptr = mmap(NULL, sqe_len, PROT_READ | PROT_WRITE,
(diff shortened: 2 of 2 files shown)


tohojo:
You forgot to change the close of lfd in xdp_multiprog__check_compat().

@magnus-karlsson any chance you could give this a spin and see if it works as a
workaround for that issue with close() hanging?

kkdwivedi:
> You forgot to change the close of lfd in xdp_multiprog__check_compat().

Thanks, fixed.

tohojo:
The LGTM check fails with this build error:
```
[2021-03-09 20:11:53] [build] [2021-03-09 20:11:53] [build-err] In file included from libxdp.c:31:
[2021-03-09 20:11:53] [build] [2021-03-09 20:11:53] [build-err] ../util/async_close.h: In function ‘__async_close’:
[2021-03-09 20:11:53] [build] [2021-03-09 20:11:53] [build-err] ../util/async_close.h:89:7: error: ‘struct io_uring_params’ has no member named ‘features’
[2021-03-09 20:11:53] [build] [2021-03-09 20:11:53] [build-err]    89 |  if (p.features & IORING_FEAT_SINGLE_MMAP) {
[2021-03-09 20:11:53] [build] [2021-03-09 20:11:53] [build-err]       |       ^
[2021-03-09 20:11:53] [build] [2021-03-09 20:11:53] [build-err] ../util/async_close.h:89:19: error: ‘IORING_FEAT_SINGLE_MMAP’ undeclared (first use in this function)
[2021-03-09 20:11:53] [build] [2021-03-09 20:11:53] [build-err]    89 |  if (p.features & IORING_FEAT_SINGLE_MMAP) {
[2021-03-09 20:11:53] [build] [2021-03-09 20:11:53] [build-err]       |                   ^~~~~~~~~~~~~~~~~~~~~~~
[2021-03-09 20:11:53] [build] [2021-03-09 20:11:53] [build-err] ../util/async_close.h:89:19: note: each undeclared identifier is reported only once for each function it appears in
[2021-03-09 20:11:53] [build] [2021-03-09 20:11:53] [build-err] ../util/async_close.h:100:7: error: ‘struct io_uring_params’ has no member named ‘features’
[2021-03-09 20:11:53] [build] [2021-03-09 20:11:53] [build-err]   100 |  if (p.features & IORING_FEAT_SINGLE_MMAP) {
[2021-03-09 20:11:53] [build] [2021-03-09 20:11:53] [build-err]       |       ^
[2021-03-09 20:11:53] [build] [2021-03-09 20:11:53] [build-err] make[2]: *** [Makefile:85: staticobjs/libxdp.o] Error 1
[2021-03-09 20:11:53] [build] [2021-03-09 20:11:53] [build-err] make[1]: *** [Makefile:10: all] Error 2
[2021-03-09 20:11:53] [build] [2021-03-09 20:11:53] [build-err] make: *** [Makefile:25: lib] Error 2
```

so this will need a check in 'configure'. However, you're looking at the features just to possibly save an mmap() call, right? So why not just skip that and always do the extra mmap?

kkdwivedi:
> so this will need a check in 'configure'. However, you're looking at the features just to possibly save an mmap() call, right? So why not just skip that and always do the extra mmap?

Yes, just to save on an extra mmap. I dropped it now.

tohojo:
Alright, cool. Guess I need to go read up on how io_uring is supposed to work to sanity check the rest, then :)

In the meantime, let's see if Magnus can test if it actually fixes the issue :)

magnus-karlsson:
Tried it out and it was like I feared: it just postpones the problem. Yes, we can attach the first program fine as we do not wait for close anymore, but the second program freezes under packet load just like before, but not on close of the raw_tracepoint but the actual operation (if I have parsed things correctly, see details below). So I think we have to solve the root cause of this, unfortunately. Or is there some other way around this?

From the second program attached:

mkarlsso@kurt:~/src/xdp-tools/xdp-loader$ sudo ./xdp-loader load ens802f0 ../lib/libxdp/xsk_def_xdp_prog.o
Verified XDP dispatcher version 1 <= 1
DATASEC '.xdp_run_config' not found.
Acquired lock from /sys/fs/bpf/xdp with fd 6
Reading multiprog component programs from pinned directory
Released lock fd 6
Found multiprog with id 37 and 3 component progs
Generating multi-prog dispatcher for 4 programs
Looking for './xdp-dispatcher.o'
Loading XDP program from './xdp-dispatcher.o' section 'xdp/dispatcher'
DATASEC '.xdp_run_config' not found.
Verified XDP dispatcher version 1 <= 1
Loading multiprog dispatcher for 4 programs
Loaded XDP program xdp_dispatcher, got fd 11
Checking dispatcher compatibility
Looking for './xdp-dispatcher.o'
Loading XDP program from './xdp-dispatcher.o' section 'xdp/pass'
DATASEC '.xdp_run_config' not found.
Loaded XDP program xdp_pass, got fd 16
***** And here it freezes until I release the packet load. *****

write(1, "Loaded XDP program xdp_pass, got"..., 39Loaded XDP program xdp_pass, got fd 15
) = 39
bpf(BPF_OBJ_GET_INFO_BY_FD, {info={bpf_fd=15, info_len=208, info=0x7ffdb64ac730}}, 120) = 0
bpf(BPF_RAW_TRACEPOINT_OPEN, {raw_tracepoint={name=NULL, prog_fd=15}}, 120) = 16
io_uring_setup(2, {flags=0, sq_thread_cpu=0, sq_thread_idle=0, sq_entries=2, cq_entries=4, features=IORING_FEAT_SINGLE_MMAP|IORING_FEAT_NODROP|IORING_FEAT_SUBMIT_STABLE|IORING_FEAT_RW_CUR_POS|IORING_FEAT_CUR_PERSONALITY|0x3e0, sq_off={head=0, tail=64, ring_mask=256, ring_entries=264, flags=276, dropped=272, array=384}, cq_off={head=128, tail=192, ring_mask=260, ring_entries=268, overflow=284, cqes=320, resv=[0x118, 0]}}) = 17
mmap(NULL, 392, PROT_READ|PROT_WRITE, MAP_SHARED|MAP_POPULATE, 17, 0) = 0x7f822dbec000
mmap(NULL, 384, PROT_READ|PROT_WRITE, MAP_SHARED|MAP_POPULATE, 17, 0x8000000) = 0x7f822dbeb000
mmap(NULL, 128, PROT_READ|PROT_WRITE, MAP_SHARED|MAP_POPULATE, 17, 0x10000000) = 0x7f822dbea000
io_uring_enter(17, 1, 0, 0, NULL, 0  **** Hangs here until I release packet load. *****

tohojo:
Magnus Karlsson ***@***.***> writes:

> Tried it out and it was like I feared: it just postpones the problem.
> Yes, we can attach the first program fine as we do not wait for close
> anymore, but the second program freezes under packet load just like
> before, but not on close of the raw_tracepoint but the actual
> operation (if I have parsed things correctly, see details below). So I
> think we have to solve the root cause of this, unfortunately. Or is
> there some other way around this?

That trace looks like it's just hanging in the same place, but due to
the io_uring_enter() syscall instead of close(). Which makes sense, that
seems to wait for completion, so we need to make it even more
fire-and-forget :)

However, I do agree we should get the root cause fixed upstream; I
consider this io_uring hack to be a workaround we can use on kernels
that are not fixed. I'll play around with the io_uring options a bit
more to explore this; but we can proceed with getting upstream fixed in
the meantime.

To do that, I'll try to hack up a minimal reproducer for you to check -
any chance you could check if the previously-reverted patch to make link
freeing deferred[0] fixes the issue?

***@***.***/

magnus-karlsson:
On Fri, Mar 19, 2021 at 4:10 PM Toke Høiland-Jørgensen <
***@***.***> wrote:

> Magnus Karlsson ***@***.***> writes:
>
> > Tried it out and it was like I feared: it just postpones the problem.
> > Yes, we can attach the first program fine as we do not wait for close
> > anymore, but the second program freezes under packet load just like
> > before, but not on close of the raw_tracepoint but the actual
> > operation (if I have parsed things correctly, see details below). So I
> > think we have to solve the root cause of this, unfortunately. Or is
> > there some other way around this?
>
> That trace looks like it's just hanging in the same place, but due to
> the io_uring_enter() syscall instead of close(). Which makes sense, that
> seems to wait for completion, so we need to make it even more
> fire-and-forget :)
>
> However, I do agree we should get the root cause fixed upstream; I
> consider this io_uring hack to be a workaround we can use on kernels
> that are not fixed. I'll play around with the io_uring options a bit
> more to explore this; but we can proceed with getting upstream fixed in
> the meantime.
>
> To do that, I'll try to hack up a minimal reproducer for you to check -
> any chance you could check if the previously-reverted patch to make link
> freeing deferred[0] fixes the issue?
>

Forgot about that one. Will do on Monday.

tohojo:
> Forgot about that one. Will do on Monday.

Awesome, thanks!

tohojo:
Here's the link to the patch in a format that Github hopefully won't mangle:

https://patchwork.ozlabs.org/project/netdev/patch/20200917074453.20621-1-songmuchun@bytedance.com/

magnus-karlsson:
Tried the patch and it does not help. It defers work for put_link, but here we are trying to close a raw tracepoint so I guess these are not related. I have waited for 10 minutes, but still the rcu_task grace period has not expired. It really seems it is stuck there forever until load is released. Tried to find some good debug options to turn on for RCU, but could not find any relevant ones. Or do you have any?

tohojo:
Magnus Karlsson ***@***.***> writes:

> Tried the patch and it does not help. It defers work for put_link, but
> here we are trying to close a raw tracepoint so I guess these are not
> related.

bpf_raw_tracepoint_open still ends up calling bpf_tracing_prog_attach()
for freplace programs, which does result in a bpf_link:

https://elixir.bootlin.com/linux/latest/source/kernel/bpf/syscall.c#L2553

> I have waited for 10 minutes, but still the rcu_task grace period has
> not expired. It really seems it is stuck there forever until load is
> released. Tried to find some good debug options to turn on for RCU,
> but could not find any relevant ones. Or do you have any?

Nope, no idea. I guess we'll just have to ask upstream. I'll see if I
can reproduce this and make a minimal example to test from. Were you
able to trigger this with regular XDP programs as well, or is it only
with AF_XDP?

magnus-karlsson:
> Magnus Karlsson ***@***.***> writes:
> Tried the patch and it does not help. It defers work for put_link, but here we are trying to close a raw tracepoint so I guess these are not related.
> bpf_raw_tracepoint_open still ends up calling bpf_tracing_prog_attach() for freplace programs, which does result in a bpf_link: https://elixir.bootlin.com/linux/latest/source/kernel/bpf/syscall.c#L2553

OK, got it.

> I have waited for 10 minutes, but still the rcu_task grace period has not expired. It really seems it is stuck there forever until load is released. Tried to find some good debug options to turn on for RCU, but could not find any relevant ones. Or do you have any?
> Nope, no idea. I guess we'll just have to ask upstream. I'll see if I can reproduce this and make a minimal example to test from. Were you able to trigger this with regular XDP programs as well, or is it only with AF_XDP?

Yes. This hangs in the same way too:

xdp-loader/xdp-loader load ens802f0 lib/libxdp/xdp-dispatcher.o

/Magnus

tohojo:
Magnus Karlsson ***@***.***> writes:

>> Magnus Karlsson ***@***.***> writes:
>> Tried the patch and it does not help. It defers work for put_link, but here we are trying to close a raw tracepoint so I guess these are not related.
>> bpf_raw_tracepoint_open still ends up calling bpf_tracing_prog_attach() for freplace programs, which does result in a bpf_link: https://elixir.bootlin.com/linux/latest/source/kernel/bpf/syscall.c#L2553
>
> OK, got it.
>
>> I have waited for 10 minutes, but still the rcu_task grace period has not expired. It really seems it is stuck there forever until load is released. Tried to find some good debug options to turn on for RCU, but could not find any relevant ones. Or do you have any?
>> Nope, no idea. I guess we'll just have to ask upstream. I'll see if I can reproduce this and make a minimal example to test from. Were you able to trigger this with regular XDP programs as well, or is it only with AF_XDP?
>
> Yes. This hangs in the same way too:
>
> xdp-loader/xdp-loader load ens802f0 lib/libxdp/xdp-dispatcher.o

Right, created a minimal reproducer here:
https://github.com/xdp-project/bpf-examples/tree/master/bpf-link-hang

will run a few more tests tomorrow, then raise this with upstream...

magnus-karlsson:
Perfect! Thanks Toke.

tohojo:
So it turns out that the reason why the kernel workaround (moving the trampoline put to a tasklet) doesn't work, is that `bpf_trampoline_put()` calls `synchronize_rcu_tasks()` while holding the trampoline mutex. So it just moves the problem so that the next time we try to load a program it will block on acquiring the mutex.

This also implies that it's no use trying to work around the issue from userspace, as that would also just end up with subsequent blocks on the mutex. So going to close this PR - but thanks anyway @kkdwivedi for the prototype :)

PR #102, opened 2021-03-23
Xsk migration5

magnus-karlsson:
Just dumping my latest version here so that it is backed up somewhere. Note that the 3 last commits are supposed to be merged into the main libbpf migration commit.

Changes since the previous version:

* Various improvements suggested by Toke
* My own clean ups to make the code less horrible
* Migrated over the latest new features from libbpf
* Rebase on latest libxdp

The only thing I have not managed to solve of Toke's suggestions is to move the attach last in the xsk program setup path. I just do not get the setting of the map to work before I have attached the program.


lib/libxdp/xsk.c
```diff
@@ -0,0 +1,982 @@
...
+	}
+	if (xsks_map_fd)
+		*xsks_map_fd = ctx->xsks_map_fd;
+
+	return 0;
+
+err_lookup:
+	if (!prog)
+		xdp_program__detach(ctx->xdp_prog, ctx->ifindex,
+				    xsk_convert_xdp_flags(xsk->config.xdp_flags), 0);
```
tohojo, line 624: This will detach the program even if it was cloned instead of attached above...

lib/libxdp/xsk.c
```diff
@@ -0,0 +1,982 @@
...
+
+	xsk->ctx = ctx;
+
+	return 0;
+}
+
+static enum xdp_attach_mode xsk_convert_xdp_flags(__u32 xdp_flags)
+{
+	if (xdp_flags & XDP_FLAGS_REPLACE)
+		pr_warn("XDP_FLAGS_REPLACE not supported by libxdp.\n");
```
tohojo, line 361: Shouldn't this warn for any flag that's not a mode?

lib/libxdp/xsk.c
```diff
@@ -0,0 +1,982 @@
...
+
+	if (xdp_flags & XDP_FLAGS_SKB_MODE)
+		return XDP_MODE_SKB;
+	if (xdp_flags & XDP_FLAGS_DRV_MODE)
+		return XDP_MODE_NATIVE;
+	if (xdp_flags & XDP_FLAGS_HW_MODE)
+		return XDP_MODE_HW;
+
+	return XDP_MODE_UNSPEC;
```
tohojo, line 370: We changed to DRV_MODE as the default in the rest of libxdp, since the "let the kernel pick" behaviour leads to unpredictability. There's not really any other way to request the automatic behaviour if the interface is just a flags field, though; so not sure what the right thing here is? Could we just disallow the automatic mode entirely?

lib/libxdp/xsk_def_xdp_prog_5.3.c
```diff
@@ -0,0 +1,39 @@
...
+} xsks_map SEC(".maps");
+
+struct {
+        __uint(priority, 20);
+	__uint(XDP_PASS, 1);
+} XDP_RUN_CONFIG(xsk_def_prog);
+
+/* This is the program for 5.3 kernels and older. */
+SEC("xdp/xsk_def_prog")
+int xsk_def_prog(struct xdp_md *ctx)
```
tohojo, line 25: Any reason these two progs couldn't just be kept in the same object (but in different sections)?

tohojo (changes requested):
First off, I think you forgot to commit `xsk_def_xdp_prog.h`? Compilation is broken without it (as you can also see from the CI results) - although the error message is a bit obscure...

And speaking of CI, I guess we should also figure out how we can add test cases for this. I suppose we need a separate set of cases for libxdp that is not just running the tools? Or alternatively we could port over xdpsock from samples as a 'tool' (maybe not installed in $PATH, but perhaps somewhere in /usr/libexec?).

As for moving the program attach last: Am I understanding you correctly that the issue is that when you first attach the program, the map won't be created until the program is loaded into the kernel, and that only happens with `xdp_program__attach()`, right? This happens because we need a target at load to attach an freplace program, so we need a dispatcher as the target. But now that we can reattach freplace programs, we could provide a separate `xdp_program__load()` that would load with a dummy dispatcher as the target, and then just do the attach later. I think this would solve your problem with the map not being created?

As few other detailed comments below...

magnus-karlsson:
> First off, I think you forgot to commit `xsk_def_xdp_prog.h`? Compilation is broken without it (as you can also see from the CI results) - although the error message is a bit obscure...

Thanks. How did you see that? I just got a compilation error of the static libxdp library.

> And speaking of CI, I guess we should also figure out how we can add test cases for this. I suppose we need a separate set of cases for libxdp that is not just running the tools? Or alternatively we could port over xdpsock from samples as a 'tool' (maybe not installed in $PATH, but perhaps somewhere in /usr/libexec?).

I think we need a simple xdpsock for a test. xdpsock itself has grown way too large and complicated as it basically used to demonstrate every single feature of AF_XDP.

> As for moving the program attach last: Am I understanding you correctly that the issue is that when you first attach the program, the map won't be created until the program is loaded into the kernel, and that only happens with `xdp_program__attach()`, right? This happens because we need a target at load to attach an freplace program, so we need a dispatcher as the target. But now that we can reattach freplace programs, we could provide a separate `xdp_program__load()` that would load with a dummy dispatcher as the target, and then just do the attach later. I think this would solve your problem with the map not being created?

Correct. I think that would solve the problem. But what are the side effects of doing an attach in the first place and that necessitates for it to be performed last? What is it that we cannot back out?

> As few other detailed comments below...

lib/libxdp/xsk.c
magnus-karlsson: Will fix.

lib/libxdp/xsk.c
magnus-karlsson: I though we specifically did not support this flag, but that might be wrong. At least I do not see an equivalent one in enum xdp_attach_mode. What are the XDP flags not supported by libxdp, if any?

lib/libxdp/xsk.c
magnus-karlsson: I think that would break compatibility in the case of this code from libbpf. But on the other hand, we should be allowed to change some things in the move, so if we just document it this would be a good candidate.

lib/libxdp/xsk_def_xdp_prog_5.3.c
magnus-karlsson: The code in xsk.c turned out simpler if I kept them in two separate files. Actually had them in the same first.

lib/libxdp/xsk.c
tohojo: Both XDP_FLAGS_REPLACE and XDP_FLAGS_UPDATE_IF_NOEXIST has to do with how the XDP program is attached to the interface. And since libxdp takes over the attachment entirely and does multiprog, none of those flags really make sense to set (libxdp does use them when attaching the dispatcher to the interface, but that is part of the dispatcher logic, so not exposed to API users)

lib/libxdp/xsk.c
tohojo: Yes, I think we should at least change the default (do you have any examples of anyone actually relying on the automatic fallback behaviour?)

I'm OK with allowing the "auto" mode, if we can find a good way to expose that to users. Ideas?

lib/libxdp/xsk_def_xdp_prog_5.3.c
tohojo: Alright, not a huge deal, just thought it might be simpler if we ever needed to update the map definition or something. But that would require other changes as well, so fine with just keeping it like this.

tohojo:
> > First off, I think you forgot to commit `xsk_def_xdp_prog.h`? Compilation is broken without it (as you can also see from the CI results) - although the error message is a bit obscure...
> 
> Thanks. How did you see that? I just got a compilation error of the static libxdp library.

I happened to have a checkout that had already compiled the library the first time I tried your branch. So I got another error that directly referred the missing header file the first time I typed `make`, and it wasn't until I did a `make clean` that I got that compilation error for the static library. So that helped with finding the culprit, but yeah, it's a bit of a cryptic error from make...

> > And speaking of CI, I guess we should also figure out how we can add test cases for this. I suppose we need a separate set of cases for libxdp that is not just running the tools? Or alternatively we could port over xdpsock from samples as a 'tool' (maybe not installed in $PATH, but perhaps somewhere in /usr/libexec?).
> 
> I think we need a simple xdpsock for a test. xdpsock itself has grown way too large and complicated as it basically used to demonstrate every single feature of AF_XDP.

OK, that's fine with me as well :)

> > As for moving the program attach last: Am I understanding you correctly that the issue is that when you first attach the program, the map won't be created until the program is loaded into the kernel, and that only happens with `xdp_program__attach()`, right? This happens because we need a target at load to attach an freplace program, so we need a dispatcher as the target. But now that we can reattach freplace programs, we could provide a separate `xdp_program__load()` that would load with a dummy dispatcher as the target, and then just do the attach later. I think this would solve your problem with the map not being created?
> 
> Correct. I think that would solve the problem. But what are the side effects of doing an attach in the first place and that necessitates for it to be performed last? What is it that we cannot back out?

Well, my thinking is that attaching the XDP program is the point at which things become visible to the world. I.e., if we back out the program it may have processed some packets in the meantime, and also attaching the program can cause the interface to reconfigure itself. Both of those are side effects I'd rather avoid in the error case if possible.

lib/libxdp/xsk.c
magnus-karlsson: Then let us print a warning to the user for both these, so they can change there program. Or would it be better to just fail if one of these are set? Do we want to be nice, or force people to change?

lib/libxdp/xsk.c
magnus-karlsson: I do not have such an example. Let us try to default to DRV mode and see if anyone complains.

magnus-karlsson:
> > > First off, I think you forgot to commit `xsk_def_xdp_prog.h`? Compilation is broken without it (as you can also see from the CI results) - although the error message is a bit obscure...
> > 
> > 
> > Thanks. How did you see that? I just got a compilation error of the static libxdp library.
> 
> I happened to have a checkout that had already compiled the library the first time I tried your branch. So I got another error that directly referred the missing header file the first time I typed `make`, and it wasn't until I did a `make clean` that I got that compilation error for the static library. So that helped with finding the culprit, but yeah, it's a bit of a cryptic error from make...
> 
> > > And speaking of CI, I guess we should also figure out how we can add test cases for this. I suppose we need a separate set of cases for libxdp that is not just running the tools? Or alternatively we could port over xdpsock from samples as a 'tool' (maybe not installed in $PATH, but perhaps somewhere in /usr/libexec?).
> > 
> > 
> > I think we need a simple xdpsock for a test. xdpsock itself has grown way too large and complicated as it basically used to demonstrate every single feature of AF_XDP.
> 
> OK, that's fine with me as well :)
> 
> > > As for moving the program attach last: Am I understanding you correctly that the issue is that when you first attach the program, the map won't be created until the program is loaded into the kernel, and that only happens with `xdp_program__attach()`, right? This happens because we need a target at load to attach an freplace program, so we need a dispatcher as the target. But now that we can reattach freplace programs, we could provide a separate `xdp_program__load()` that would load with a dummy dispatcher as the target, and then just do the attach later. I think this would solve your problem with the map not being created?
> > 
> > 
> > Correct. I think that would solve the problem. But what are the side effects of doing an attach in the first place and that necessitates for it to be performed last? What is it that we cannot back out?
> 
> Well, my thinking is that attaching the XDP program is the point at which things become visible to the world. I.e., if we back out the program it may have processed some packets in the meantime, and also attaching the program can cause the interface to reconfigure itself. Both of those are side effects I'd rather avoid in the error case if possible.

Got it. Then let us try to do the attach last.

lib/libxdp/xsk.c
tohojo: SGTM

lib/libxdp/xsk.c
tohojo: Heh, always the question, eh? To be nice or to be correct... I think a warning is enough here, since we'll do something reasonable regardless of the flag values...

lib/libxdp/xsk.c
magnus-karlsson: I believe it is correct as it is. The clone path is only taken if the local variable prog is set and prog is only set at the beginning before cloning or attaching. Note that all other calls use ctx->xdp_prog.

tohojo:
Right, so I think this is a decent base; outstanding issues AFAICT:
- Reordering operations so attach is last (requires some changes in libxdp.c)
- Test(s)
- A scheme for sharing the XDP prog on an interface (like what Maciej implemented in libbpf with bpf_link)

Do you have any preference for whether you want to implement these as part of this PR? Or should we merge this and add the rest on top?

magnus-karlsson:
I think we should merge this as is just so that users can try it out. But
please hold on on the merge since I need to get permission internally first
since libxdp is something that nobody from Intel has contributed to yet.
Might take a week or two, just so you know. In the meantime, I will start
on the tests. Item #1 and #3, we can develop and merge later on top. Have
some other things I would like to add later too.

tohojo:
Alright, just let me know when you have finished the lawyer appeasement ritual ;)

headers/xdp/libxdp_util.h
```diff
@@ -0,0 +1,79 @@
...
+
+#ifndef libbpf_smp_load_acquire
+#define libbpf_smp_load_acquire(p)                                     \
+        ({                                                             \
+		typeof(*p) ___p1 = __XSK_READ_ONCE(*p);                \
+		__sync_synchronize();                                  \
+		___p1;                                                 \
+	})
+#endif
+
```
kkdwivedi, line 72: Just a suggestion: I think instead of falling back to a full barrier on platforms we don't have a specialized implementation for, we should just fall back to using the __atomic builtins GCC/clang provide \[1\] with ATOMIC_ACQUIRE/ATOMIC_RELEASE ordering on all platforms, which should also do the right thing for x86, ARM and RISCV.

So libbpf_atomic_acquire simply maps to __atomic_load_n(ptr, __ATOMIC_ACQUIRE) and likewise release would be __atomic_store_n(ptr, val, __ATOMIC_RELEASE). So you can also drop READ_ONCE/WRITE_ONCE as the load/store itself is tagged with acquire/release (so this becomes a one way barrier instead of a two way acquire/release fence, which should be good enough IIUC).

e.g. codegen on ARM is the same: https://godbolt.org/z/e5Gc3WY6n

\[1\]: https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html

magnus-karlsson:
On Thu, May 13, 2021 at 12:41 AM Kumar Kartikeya Dwivedi <
***@***.***> wrote:

> ***@***.**** commented on this pull request.
> ------------------------------
>
> In headers/xdp/libxdp_util.h
> <https://github.com/xdp-project/xdp-tools/pull/102#discussion_r631446497>:
>
> > +#define libbpf_smp_store_release(p, v)                                 \
> +        do {                                                           \
> +		__sync_synchronize();                                  \
> +		__XSK_WRITE_ONCE(*p, v);                               \
> +        } while (0)
> +#endif
> +
> +#ifndef libbpf_smp_load_acquire
> +#define libbpf_smp_load_acquire(p)                                     \
> +        ({                                                             \
> +		typeof(*p) ___p1 = __XSK_READ_ONCE(*p);                \
> +		__sync_synchronize();                                  \
> +		___p1;                                                 \
> +	})
> +#endif
> +
>
> Just a suggestion: I think instead of falling back to a full barrier on
> platforms we don't have a specialized implementation for, we should just
> fall back to using the __atomic builtins GCC/clang provide [1] with
> ATOMIC_ACQUIRE/ATOMIC_RELEASE ordering on all platforms, which should also
> do the right thing for x86, ARM and RISCV.
>
> So libbpf_atomic_acquire simply maps to __atomic_load_n(ptr,
> __ATOMIC_ACQUIRE) and likewise release would be __atomic_store_n(ptr, val,
> __ATOMIC_RELEASE). So you can also drop READ_ONCE/WRITE_ONCE as the
> load/store itself is tagged with acquire/release (so this becomes a one way
> barrier instead of a two way acquire/release fence, which should be good
> enough IIUC).
>

Good idea. Will fix.

e.g. codegen on ARM is the same: https://godbolt.org/z/e5Gc3WY6n

tohojo:
@magnus-karlsson could you try pushing to the branch again, please? seems the previous CI issue was transient...

magnus-karlsson:
Pushed.

/Magnus

tohojo:
Alright, awesome, let's merge this then :)

Issue #103, opened 2021-03-24
Do not load dispatcher when only xsk program is active


magnus-karlsson:
The XDP dispatcher has a significant performance impact (up to 11% decrease) on the performance of an AF_XDP socket. One way to get around this is to not load the dispatcher when the default xsk program is the only one active. The dispatcher is not needed in this case.

Toke suggested that we add some special handling of the xsk program and just load it when there is no other XDP program there. If more programs are added, we simply add the dispatcher and in that case the performance drop should be justified, since you are getting added functionality. The question now is: why do we not do this for any program, i.e. only load the dispatcher only if there are two or more XDP programs that want to share an interface? The only reason I can come up with is that it is much faster to add or remove a program if the dispatcher is already there. Adding the dispatcher plus two programs might result in dropped packets due to this latency, but I do not know. Other reasons?

Plan from Toke:

- When attaching XSK program, if nothing is attached already, just
  attach it directly to the interface.

- When attaching XSK program and another program is already installed,
  do the regular thing and attach it as a secondary program with a
  dispatcher.

- When attaching *another* program and an XSK program is already
  attached, generate a new dispatcher with a new copy of the XSK program
  as one of the sub-programs, but reusing the map from the program
  already attached to the interface; then replace the existing program
  with that dispatcher.

tohojo:
First off, @shoracek had an idea that may obviate the need for that whole dance: Simply attach the freplace program to the main function of the dispatcher if there's only a single component program; that way there won't be an additional function call, and I'm hoping the overhead will (mostly? entirely?) go away then. Stepan, do you want to take a crack at implementing this? Otherwise I'll take a look once I've cleared a few other things off my plate...

But to explain further, the reason we can't just attach other programs like regular XDP programs and then install the dispatcher once a second program shows up, is that we can't change the program type once it's loaded. And when we're attaching a program with freplace it needs to be loaded with `BPF_PROG_TYPE_EXT` as its program type at load time.

With the AF_XDP program we can just load another copy of the program with the right type, because we have the object file handy as part of the library. But for user-loaded XDP programs we can't, unless we cache the object files somewhere, and that I'd rather not do...

shoracek:
Sure thing, I'll try to figure something out.

tohojo:
Štěpán Horáček ***@***.***> writes:

> Sure thing, I'll try to figure something out.

Great, thanks! :)

Should be fairly straight-forward: I think we can keep the pinning and
[...]

PR #105, opened 2021-04-06
Freplace dispatcher if there is only one program attached

shoracek:
One commit introduces this. Since the default behaviour is changed (causing e.g. xdpdump to not be able to be attached), the other commit disables it by default. This might fix issue #103, but while testing I did not see any noticeable difference in performance even between single legacy program and a program with non-freplaced dispatcher, but that might be just a bad testing setup.

### DIFF
--- a/lib/libxdp/libxdp.c
+++ b/lib/libxdp/libxdp.c
@@ -2614,6 +2614,14 @@ bool xdp_multiprog__is_legacy(const struct xdp_multiprog *mp)
 	return !!(mp && mp->is_legacy);
 }
 
+int xdp_multiprog__program_count(const struct xdp_multiprog *mp)
+{
+	if (!mp)
+		return -EINVAL;
+
+	return mp->num_links;
+}
+
 static __u32 xdp_get_ifindex_prog_id(int ifindex)
 {
 	struct xdp_link_info xinfo = {};
--- a/lib/libxdp/libxdp.c
+++ b/lib/libxdp/libxdp.c
@@ -2062,6 +2062,7 @@ static int xdp_multiprog__link_prog(struct xdp_multiprog *mp,
 	bool was_loaded = false;
 	char buf[PATH_MAX];
 	int err, lfd = -1;
+	char *attach_func;
 	__s32 btf_id;
 
 	if (!mp || !prog || !mp->is_loaded ||
@@ -2085,17 +2086,23 @@ static int xdp_multiprog__link_prog(struct xdp_multiprog *mp,
 	if (err)
 		goto err;
 
-	btf_id = find_prog_btf_id(buf, mp->main_prog->prog_fd);
+
+	if (mp->config.num_progs_enabled == 1)
+		attach_func = "xdp_dispatcher";
+	else
+		attach_func = buf;
+
+	btf_id = find_prog_btf_id(attach_func, mp->main_prog->prog_fd);
 	if (btf_id <= 0) {
 		err = btf_id;
-		pr_debug("Couldn't find BTF ID for %s: %d\n", buf, err);
+		pr_debug("Couldn't find BTF ID for %s: %d\n", attach_func, err);
 		goto err;
 	}
 
 	if (prog->prog_fd < 0) {
 		err = bpf_program__set_attach_target(prog->bpf_prog,
 						     mp->main_prog->prog_fd,
-						     buf);
+						     attach_func);
 		if (err) {
 			pr_debug("Failed to set attach target: %s\n", strerror(-err));
 			goto err;
--- a/xdp-dump/xdpdump.c
+++ b/xdp-dump/xdpdump.c
@@ -951,6 +951,13 @@ static int find_target(struct dumpopt *cfg, struct xdp_multiprog *mp,
 
 	/* First take care of the default case, i.e. no function supplied */
 	if (!program_names) {
+		/* The libxdp code optimization where it skips the dispatcher
+		 * if only one program is loaded. If this is the case, we need
+		 * to attach to the actual first program, not the dispatcher.
+		 */
+		if (xdp_multiprog__program_count(mp) == 1)
+			prog = xdp_multiprog__next_prog(NULL, mp);
+
 		matches = find_func_matches(xdp_program__btf(prog),
 					    xdp_program__name(prog),
 					    &func, false, -1, false);
@@ -970,7 +977,6 @@ static int find_target(struct dumpopt *cfg, struct xdp_multiprog *mp,
 		pr_warn("ERROR: Can't identify the full XDP main function!\n"
 			"The following is a list of candidates:\n");
 
-		prog = xdp_multiprog__main_prog(mp);
 		find_func_matches(xdp_program__btf(prog),
 				  xdp_program__name(prog),
 				  &func, true, -1, false);
--- a/headers/xdp/libxdp.h
+++ b/headers/xdp/libxdp.h
@@ -100,5 +100,6 @@ enum xdp_attach_mode xdp_multiprog__attach_mode(const struct xdp_multiprog *mp);
 struct xdp_program *xdp_multiprog__main_prog(const struct xdp_multiprog *mp);
 struct xdp_program *xdp_multiprog__hw_prog(const struct xdp_multiprog *mp);
 bool xdp_multiprog__is_legacy(const struct xdp_multiprog *mp);
+int xdp_multiprog__program_count(const struct xdp_multiprog *mp);
 
 #endif
--- a/lib/libxdp/libxdp.map
+++ b/lib/libxdp/libxdp.map
@@ -40,4 +40,5 @@ LIBXDP_1.0.0 {
 
 LIBXDP_1.2.0 {
 		libxdp_clean_references;
+		xdp_multiprog__program_count;
 } LIBXDP_1.0.0;
--- a/xdp-dump/tests/test-xdpdump.sh
+++ b/xdp-dump/tests/test-xdpdump.sh
@@ -105,7 +105,7 @@ test_capt_pcap()
     $PING6 -W 2 -c 1 "$INSIDE_IP6" || return 1
     $XDP_LOADER load "$NS" "$TEST_PROG_DIR/test_long_func_name.o" || return 1
 
-    PID=$(start_background "$XDPDUMP -i $NS --use-pcap -w - 2> /dev/null | tcpdump -r - -n")
+    PID=$(start_background "$XDPDUMP -i $NS -p xdp_test_prog_with_a_long_name --use-pcap -w - 2> /dev/null | tcpdump -r - -n")
     $PING6 -W 2 -c 1 "$INSIDE_IP6" || return 1
     sleep 1
     RESULT=$(stop_background "$PID")
@@ -145,13 +145,13 @@ test_capt_pcapng()
     INFOS_REGEX+="Capture application: xdpdump v[0-9]+\.[0-9]+\.[0-9]+.*"
     INFOS_REGEX+="Capture comment:     Capture was taken on interface xdptest, with the following XDP programs loaded:   xdp_dispatcher\(\)     xdp_test_prog_w.*"
     INFOS_REGEX+="Interface #0 info:.*"
-    INFOS_REGEX+="Name = ${NS}:xdp_dispatcher\(\)@fentry.*"
+    INFOS_REGEX+="Name = ${NS}:xdp_test_prog_with_a_long_name\(\)@fentry.*"
     if [ $OLD_CAPINFOS -eq 0 ]; then
         INFOS_REGEX+="Hardware = driver: \"veth\", version: \"1\.0\", fw-version: \"\", rom-version: \"\", bus-info: \"\".*"
     fi
     INFOS_REGEX+="Time precision = nanoseconds \(9\).*"
     INFOS_REGEX+="Interface #1 info:.*"
-    INFOS_REGEX+="Name = ${NS}:xdp_dispatcher\(\)@fexit.*"
+    INFOS_REGEX+="Name = ${NS}:xdp_test_prog_with_a_long_name\(\)@fexit.*"
     if [ $OLD_CAPINFOS -eq 0 ]; then
         INFOS_REGEX+="Hardware = driver: \"veth\", version: \"1\.0\", fw-version: \"\", rom-version: \"\", bus-info: \"\".*"
     fi
@@ -160,7 +160,7 @@ test_capt_pcapng()
     $PING6 -W 2 -c 1 "$INSIDE_IP6" || return 1
     $XDP_LOADER load "$NS" "$TEST_PROG_DIR/test_long_func_name.o" || return 1
 
-    PID=$(start_background "$XDPDUMP -i $NS -w - 2> /dev/null | tcpdump -r - -n")
+    PID=$(start_background "$XDPDUMP -i $NS -p xdp_test_prog_with_a_long_name -w - 2> /dev/null | tcpdump -r - -n")
     $PING6 -W 2 -c 1 "$INSIDE_IP6" || return 1
     RESULT=$(stop_background "$PID")
 
@@ -169,7 +169,7 @@ test_capt_pcapng()
         return 1
     fi
 
-    PID=$(start_background "$XDPDUMP -i $NS -w $PCAP_FILE --rx-capture=entry,exit")
+    PID=$(start_background "$XDPDUMP -i $NS -p xdp_test_prog_with_a_long_name -w $PCAP_FILE --rx-capture=entry,exit")
     $PING6 -W 2 -c 1 "$INSIDE_IP6" || return 1
     RESULT=$(stop_background "$PID") || (print_result "xdpdump failed"; return 1)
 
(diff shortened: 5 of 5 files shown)


tohojo:
I don't think we should make this a choice for the caller. I'd much rather if we could just always do this as a performance optimisation. Why does it break xdpdump? And can't we just fix that?

Also, @magnus-karlsson could you please test if this fixes the performance regression for AF_XDP when using the dispatcher? :)

magnus-karlsson:
> I don't think we should make this a choice for the caller. I'd much rather if we could just always do this as a performance optimisation. Why does it break xdpdump? And can't we just fix that?
> 
> Also, @magnus-karlsson could you please test if this fixes the performance regression for AF_XDP when using the dispatcher? :)

As soon as I get access to my lab again I can test this. It has been down for the past week, sigh, but it seems that we are closer to root causing the problem today. Will let you know when it is fixed.

shoracek:
> I don't think we should make this a choice for the caller. I'd much rather if we could just always do this as a performance optimisation. Why does it break xdpdump? And can't we just fix that?

Tried to fix xdpdump. Hopefully I didn't break anything, but the tests are passing.

xdp-dump/xdpdump.c
```diff
@@ -947,7 +947,14 @@ static int find_target(struct dumpopt *cfg, struct xdp_multiprog *mp,
...
 
-	prog = xdp_multiprog__main_prog(mp);
+	struct xdp_program *first_prog = xdp_multiprog__next_prog(NULL, mp);
+	struct xdp_program *second_prog = xdp_multiprog__next_prog(first_prog, mp);
+	bool freplaced_dispatcher = first_prog != NULL && second_prog == NULL;
+	if (freplaced_dispatcher) {
+		prog = first_prog;
+	} else {
+		prog = xdp_multiprog__main_prog(mp);
+	}
```
tohojo, line 957: Hmm, I wonder if  instead of detecting this case, we could just make xdpdump always attach to the first component program if there is only one? Shouldn't make any difference for the dumped packets, and it makes things work regardless of whether the freplace target was changed or not (so it's backwards compatible).

@chaudron any objections to that?

chaudron: Trying to understand why we need this change at all? If there is a dispatcher XDP dump should attach to the dispatcher, if there is none it will be the first program. Or do I miss something?

tohojo: See the previous commit: the idea is to change the libxdp logic so that if there's only a single program, we'll just freplace the main function of the dispatcher instead of the prog0 function that it calls. This saves the overhead of calling multiple functions, which hopefully makes the performance overhead of the dispatcher go away for a single program; but it means that the main dispatcher function never gets called at all, so attaching to it from xdpdump no longer works...

xdp-dump/xdpdump.c
```diff
@@ -947,7 +947,14 @@ static int find_target(struct dumpopt *cfg, struct xdp_multiprog *mp,
 	char                    *prog_safe_ptr;
 	char                    *program_names = cfg->program_names;
 
-	prog = xdp_multiprog__main_prog(mp);
+	struct xdp_program *first_prog = xdp_multiprog__next_prog(NULL, mp);
+	struct xdp_program *second_prog = xdp_multiprog__next_prog(first_prog, mp);
+	bool freplaced_dispatcher = first_prog != NULL && second_prog == NULL;
+	if (freplaced_dispatcher) {
```
chaudron, line 953: No {} needed for single statements, like the Linux kernel coding style.

xdp-dump/xdpdump.c
chaudron: Doesn't this change mess up the entire API? Because now we think two programs are executed/loaded, but we only have one? It might also confuse users, for example for a user displaying the iface_print_status() output makes him feel like we still use the dispatcher?

Can we not have xdp_multiprog__main_prog() return the first (and only) attached program in this case? It will solve the problem for xdpdump, and would not require any changes.

tohojo: Well, my feeling is that it's sort of an under-the-hood detail that I'm not sure if we should be exposing at all? I mean from a semantic PoV, there's still a dispatcher that dispatches to a single program, it's just that we changed the details of how the call is made to avoid some overhead?

So from an API PoV, you still have main_prog() is the dispatcher which is the ID that is attached to the interface; and the actual user progs are returned by __next_prog() and those are what you should be looking at to find what is actually processing the packets. So really, making assumptions about the main program is a bit of a niche application that we're only doing in xdpdump, so it's easier to just fix that?

Changing main_prog() to return the first program has the problem that then there would be no way to get at the actual dispatcher program; which is still the program ID attached to the interface, even if it is never executed. So if someone wanted to, say, check if the main prog was attached to an interface, they'd need a reference to the dispatcher.

Also, there's an additional problem in that we can't actually detect if this optimisation has been applied: the kernel has no interface to discover the target of an attached freplace program. I'm going to submit a patch to fix that, but for older kernels that obviously doesn't help. So if we are going to track that in a backwards-compatible way, we'll need to keep that information somewhere like bpffs; or we could bump the dispatcher version and deal with it that way (but thus losing forward-compatibility for old versions of the library)...

chaudron: Let me take another look tomorrow, the first commit has no comments, and I'm having a problem trying to grasp it (guess I need to sleep or get coffee...)

tohojo: Heh, yeah, I see what you mean, it's missing a bit of context. @shoracek I guess you should take that as feedback: Please write a commit message for the first commit explaining what it's doing and why.

@chaudron see #103 for a bit of that missing context :)

chaudron: @shoracek I looked at your change, and I think it can be more clean doing Toke's suggestion. Here is a diff you could apply that will do all that is needed from the tcpdump side:

```
diff --git a/xdp-dump/xdpdump.c b/xdp-dump/xdpdump.c
index 69a703d..c398b19 100644
--- a/xdp-dump/xdpdump.c
+++ b/xdp-dump/xdpdump.c
@@ -947,17 +947,21 @@ static int find_target(struct dumpopt *cfg, struct xdp_multiprog *mp,
        char                    *prog_safe_ptr;
        char                    *program_names = cfg->program_names;
 
-       struct xdp_program *first_prog = xdp_multiprog__next_prog(NULL, mp);
-       struct xdp_program *second_prog = xdp_multiprog__next_prog(first_prog, mp);
-       bool freplaced_dispatcher = first_prog != NULL && second_prog == NULL;
-       if (freplaced_dispatcher) {
-               prog = first_prog;
-       } else {
-               prog = xdp_multiprog__main_prog(mp);
-       }
+       prog = xdp_multiprog__main_prog(mp);
 
        /* First take care of the default case, i.e. no function supplied */
        if (!program_names) {
+               struct xdp_program *first_prog;
+
+               /* The libxdp code optimization where it skips the dispatcher
+                * if only one program is loaded. If this is the case, we need
+                * to attach to the actual first program, not the dispatcher.
+                */
+               first_prog = xdp_multiprog__next_prog(NULL, mp);
+               if (first_prog != NULL &&
+                   xdp_multiprog__next_prog(first_prog, mp) == NULL)
+                       prog = first_prog;
+
                matches = find_func_matches(xdp_program__btf(prog),
                                            xdp_program__name(prog),
                                            &func, false, -1, false);
@@ -1006,23 +1010,21 @@ static int find_target(struct dumpopt *cfg, struct xdp_multiprog *mp,
                progs.progs[progs.nr_of_progs].func = NULL;
        progs.nr_of_progs++;
 
-       if (!freplaced_dispatcher) {
-               for (p = xdp_multiprog__next_prog(NULL, mp);
-                    p;
-                    p = xdp_multiprog__next_prog(p, mp)) {
-
-                       progs.progs[progs.nr_of_progs].prog = p;
-                       matches = find_func_matches(xdp_program__btf(p),
-                                                   xdp_program__name(p),
-                                                   &progs.progs[progs.nr_of_progs].func,
-                                                   false, -1, false);
-                       if (matches != 1)
-                               progs.progs[progs.nr_of_progs].func = NULL;
-                       progs.nr_of_progs++;
-
-                       if (progs.nr_of_progs >= MAX_LOADED_XDP_PROGRAMS)
-                               break;
-               }
+       for (p = xdp_multiprog__next_prog(NULL, mp);
+            p;
+            p = xdp_multiprog__next_prog(p, mp)) {
+
+               progs.progs[progs.nr_of_progs].prog = p;
+               matches = find_func_matches(xdp_program__btf(p),
+                                           xdp_program__name(p),
+                                           &progs.progs[progs.nr_of_progs].func,
+                                           false, -1, false);
+               if (matches != 1)
+                       progs.progs[progs.nr_of_progs].func = NULL;
+               progs.nr_of_progs++;
+
+               if (progs.nr_of_progs >= MAX_LOADED_XDP_PROGRAMS)
+                       break;
        }
 
        /* If "all" option is specified create temp program names */
```

The changes in the test cases are fine, as there are still some cases having multiple programs loaded I think no new test case is needed for this new behavior.

chaudron: To maybe make this cleaner:

```
+ first_prog = xdp_multiprog__next_prog(NULL, mp);
+ if (first_prog != NULL && + xdp_multiprog__next_prog(first_prog, mp) == NULL) 
+     prog = first_prog; 
```

We could add an API to return the number of programs loaded? i.e. xdp_multiprog__nr_loaded_program().

So it will become:
```
+ if (dp_multiprog__nr_loaded_program(mp) == 1) 
+     prog = first_prog; 
```

tohojo: Yeah, fine with adding that to the API... bikeshedding the name a bit, maybe `xdp_multiprog__program_count()` ?

chaudron: Ok, so the request to @shoracek would be to add the xdp_multiprog__program_count() API to the first commit and take my chances with using the xdp_multiprog__program_count() for the second commit. And add some more details to the commit message for both. @tohojo did I miss anything?

tohojo: No, that sounds about right! AFAICT :)

shoracek:
Added the requested changes, but with separate `xdp_multiprog__program_count()` commit.

lib/libxdp/libxdp.c
```diff
@@ -2614,6 +2614,20 @@ bool xdp_multiprog__is_legacy(const struct xdp_multiprog *mp)
...
 
+int xdp_multiprog__program_count(const struct xdp_multiprog *mp)
+{
+	struct xdp_program *prog = NULL;
+	int count = 0;
+
+	if (!mp)
+		return -EINVAL;
+
+	while ((prog = xdp_multiprog__next_prog(prog, mp)) != NULL)
```
tohojo, line 2625: Erm, just return mp->num_links? :)

tohojo (changes requested):
You're doing a bit too much work in that program_count() function. Also, the commit needs a commit message :)

tohojo (approved):
LGTM - @chaudron please take a look as well :)

tohojo:
And let's wait for @magnus-karlsson to test the performance before we merge this :)

magnus-karlsson:
I get a verifier error when trying this out. xdp-loader does not say much:

mkarlsso@kurt:~/src/xdp-tools$ sudo xdp-loader/xdp-loader load ens785f0 lib/libxdp/xsk_def_xdp_prog.o
libxdp: Got verifier error while loading dispatcher.
libxdp: Falling back to loading single prog without dispatcher

But if I run my stuff on top of it, I dump some more:

root@kurt:~/src/dna-linux# taskset -c 2 samples/bpf/xdpsock -i ens785f0 -q 2 -n 3 -r -N -B -b 256
libbpf: elf: skipping unrecognized data section(7) .xdp_run_config
libbpf: elf: skipping unrecognized data section(8) xdp_metadata
libbpf: elf: skipping unrecognized data section(8) xdp_metadata
libbpf: load bpf program failed: Invalid argument
libbpf: -- BEGIN DUMP LOG ---
libbpf: 
Func#11 is safe for any args that match its prototype
btf_vmlinux is malformed
R1 type=ctx expected=fp
; int xdp_dispatcher(struct xdp_md *ctx)
0: (bf) r6 = r1
1: (b7) r0 = 2
; __u8 num_progs_enabled = conf.num_progs_enabled;
2: (18) r8 = 0xffffc9000cd8c000
4: (71) r7 = *(u8 *)(r8 +0)
 R0_w=invP2 R1=ctx(id=0,off=0,imm=0) R6_w=ctx(id=0,off=0,imm=0) R8_w=map_value(id=0,off=0,ks=4,vs=84,imm=0) R10=fp0
; if (num_progs_enabled < 1)
5: (15) if r7 == 0x0 goto pc+141
; ret = prog0(ctx);
6: (bf) r1 = r6
7: (85) call pc+140
btf_vmlinux is malformed
R1 type=ctx expected=fp
Caller passes invalid args into func#1
processed 84 insns (limit 1000000) max_states_per_insn 0 total_states 9 peak_states 9 mark_read 1

libbpf: -- END LOG --
libbpf: failed to load program 'xdp_dispatcher'
libbpf: failed to load object '/home/mkarlsso/src/xdp-tools/lib/libxdp//xdp-dispatcher.o'
libxdp: Got verifier error while loading dispatcher.
libxdp: Falling back to loading single prog without dispatcher

Any idea what this is? Trying this with latest bpf-next. What am I missing?

magnus-karlsson:
> I get a verifier error when trying this out. xdp-loader does not say much:
> 
> mkarlsso@kurt:~/src/xdp-tools$ sudo xdp-loader/xdp-loader load ens785f0 lib/libxdp/xsk_def_xdp_prog.o
> libxdp: Got verifier error while loading dispatcher.
> libxdp: Falling back to loading single prog without dispatcher
> 
> But if I run my stuff on top of it, I dump some more:
> 
> root@kurt:~/src/dna-linux# taskset -c 2 samples/bpf/xdpsock -i ens785f0 -q 2 -n 3 -r -N -B -b 256
> libbpf: elf: skipping unrecognized data section(7) .xdp_run_config
> libbpf: elf: skipping unrecognized data section(8) xdp_metadata
> libbpf: elf: skipping unrecognized data section(8) xdp_metadata
> libbpf: load bpf program failed: Invalid argument
> libbpf: -- BEGIN DUMP LOG ---
> libbpf:
> Func#11 is safe for any args that match its prototype
> btf_vmlinux is malformed
> R1 type=ctx expected=fp
> ; int xdp_dispatcher(struct xdp_md *ctx)
> 0: (bf) r6 = r1
> 1: (b7) r0 = 2
> ; __u8 num_progs_enabled = conf.num_progs_enabled;
> 2: (18) r8 = 0xffffc9000cd8c000
> 4: (71) r7 = *(u8 *)(r8 +0)
> R0_w=invP2 R1=ctx(id=0,off=0,imm=0) R6_w=ctx(id=0,off=0,imm=0) R8_w=map_value(id=0,off=0,ks=4,vs=84,imm=0) R10=fp0
> ; if (num_progs_enabled < 1)
> 5: (15) if r7 == 0x0 goto pc+141
> ; ret = prog0(ctx);
> 6: (bf) r1 = r6
> 7: (85) call pc+140
> btf_vmlinux is malformed
> R1 type=ctx expected=fp
> Caller passes invalid args into func#1
> processed 84 insns (limit 1000000) max_states_per_insn 0 total_states 9 peak_states 9 mark_read 1
> 
> libbpf: -- END LOG --
> libbpf: failed to load program 'xdp_dispatcher'
> libbpf: failed to load object '/home/mkarlsso/src/xdp-tools/lib/libxdp//xdp-dispatcher.o'
> libxdp: Got verifier error while loading dispatcher.
> libxdp: Falling back to loading single prog without dispatcher
> 
> Any idea what this is? Trying this with latest bpf-next. What am I missing?

I am likely missing BTF for Linux since it does not compile with it. Will try to fix that.

magnus-karlsson:
This is good! Only a 200 kpps drop in performance which is around 1% on my server and well within my tolerance. Let us merge it. Thanks!

tohojo:
Awesome! @shoracek could you please put a "Fixes #103" tag into the commit message of the first commit, so we can record the fact? :)

lib/libxdp/libxdp.c
```diff
@@ -2607,6 +2614,14 @@ bool xdp_multiprog__is_legacy(const struct xdp_multiprog *mp)
 	return !!(mp && mp->is_legacy);
 }
 
+int xdp_multiprog__program_count(const struct xdp_multiprog *mp)
+{
+	if (!mp)
+		return -EINVAL;
+
+	return mp->num_links;
```
chaudron, line 2622: Any reason why using mp->num_links here, and above mp->config.num_progs_enabled?

xdp-dump/xdpdump.c
chaudron: xdpdump changes look good to me!

lib/libxdp/libxdp.c
chaudron: If this is ok with @tohojo I have no further comments and you can go ahead with the merge.

shoracek: I used `mp->config.num_progs_enabled` because it returns the total amount of programs to be loaded, compared to `mp->num_links` returning the amount of currently loaded programs. So since `xdp_multiprog__link_prog` is called while programs are still being loaded `mp->num_links` would return just the number of already loaded programs, which would always be 1 once.

chaudron: Thanks this is now clear.

chaudron (approved):

PR #113, opened 2021-05-28
CI experiments

tohojo:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -71,3 +71,5 @@ env:
     - CLANG=clang-10
     - LLC=llc-10
     - VERBOSE=1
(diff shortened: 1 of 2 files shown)


tohojo:
This PR is just for testing out the CI runs, as @magnus-karlsson reported some issues with that.

Also, it seems we will need to move away from Travis CI as they are finally closing down the .org version and the "free open source credits" option does not look appealing. Going to test out github actions and see if we can get QEMU to run in software-only emulation mode there (since there doesn't seem to be nested virtualisation support in github actions)

tohojo:
Superseded by #116

PR #115, opened 2021-06-09
Dev/xdpdump missing btf

### LINKED ISSUE #112: xdp-dump: add more clear error message when BTF information is not found.
See output below, i.e. fix `ERROR: Can't find function '' on interface!` to be more specific as `-vv` is not used by default.

```
# ./xdpdump -i enp3s0f1 --rx-capture entry,exit -w capture.pcap -vv 
Current rlimit 67108864 already >= minimum 1048576
libxdp: No BTF for prog ID 11
libxdp: No BTF found for program
libxdp: Found legacy program with id 11 and 0 component progs
No BTF found for program
ERROR: Can't find function '' on interface!
```

chaudron:
Fixes #112 and a crash found during playing with command line options.

### DIFF
--- a/xdp-dump/xdpdump.c
+++ b/xdp-dump/xdpdump.c
@@ -924,6 +924,20 @@ static int match_target_function(struct dumpopt *cfg,
 	return -EAGAIN;
 }
 
+
+/*****************************************************************************
+ * check_btf()
+ *****************************************************************************/
+static bool check_btf(struct xdp_program *prog)
+{
+	if (xdp_program__btf(prog))
+		return true;
+
+	pr_warn("ERROR: xdpdump requires BTF information, but that is missing "
+		"from the loaded XDP program!\n");
+	return false;
+}
+
 /*****************************************************************************
  * find_target()
  *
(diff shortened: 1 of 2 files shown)


xdp-dump/xdpdump.c
```diff
@@ -924,6 +924,20 @@ static int match_target_function(struct dumpopt *cfg,
...
+/*****************************************************************************
+ * check_btf()
+ *****************************************************************************/
+static bool check_btf(struct xdp_program *prog)
+{
+	if (xdp_program__btf(prog))
+		return true;
+
+	pr_warn("ERROR: The loaded XDP program has no BTF information, "
+		"which xdpdump requires!\n");
```
tohojo, line 937: Small nit: this could be interpreted to mean "xdpdump requires that there is no BTF information". So maybe change it to something like "xdpdump requires BTF information, but that is missing from the loaded XDP program"?

chaudron: When reading again you are right will update pull req.

Issue #114, opened 2021-06-02
Move CI to Github Actions


tohojo:
Travis is moving to a paid tier, so can't use that anymore. Github Actions seems viable, if we can solve the lack of nested virtualisation (maybe QEMU can run in software more with TCG acceleration there, though?)

tohojo:
See also #113 - seems it's possible to get the tests to run on Travis with TCQ accel at least, although they're suffering from some other failures now...

tohojo:
Fixed by #116

PR #116, opened 2021-06-22
Migrate to GitHub Actions

shoracek:
Addresses issue #114.


tohojo:
Awesome, thanks! Do you know if there is some repository configuration that needs to happen to enable this, or is it enough to just merge the PR?

xdp-dump/tests/test-xdpdump.sh
```diff
@@ -341,7 +341,7 @@ test_multi_pkt()
     for PKT_SIZE in "${PKT_SIZES[@]}" ; do
 
         PID=$(start_background_no_stderr "$XDPDUMP -i $NS -p xdp_test_prog_with_a_long_name --rx-capture=entry,exit")
-        timeout 4 $PING6 -W 2 -s "$PKT_SIZE" -c 20000 -f "$INSIDE_IP6" || return 1
+        $PING6 -W 2 -s "$PKT_SIZE" -c 20000 -f "$INSIDE_IP6" || return 1
```
chaudron, line 344: The timeout should not be deleted, as it could cause the test to run for an extremely long time on failure.

tohojo: We could increase it to something that should never be hit in practice but still won't leave the test running forever? Like 30 seconds or so?

chaudron: Guess the question was, why was it removed? Was it causing a timeout before the test was actually completed?

If so, yes then changing it to "timeout 30 ..." would be better than not having it.

shoracek: While initially testing it, even after increasing the time, it sometimes caused a timeout. But now, I cannot reproduce that anymore, so I will just use the longer timeouts.

tohojo:
Ah, awesome, and now the tests also ran on the PR; looks good! Thanks a bunch for fixing this! :)

tohojo:
Right, so the first test-run after merging failed...
https://github.com/xdp-project/xdp-tools/runs/2904042782?check_suite_focus=true

@chaudron any thoughts on how to make that test more robust?

chaudron:
This is not the test with the timeout. Not sure why it's not receiving the packet. Maybe it takes longer for xdpdump to load, and hence the packets arrive before it's ready? Maybe the sleep 1, needs to be sleep 2 in start_background()?

tohojo:
Eelco Chaudron ***@***.***> writes:

> This is not the test with the timeout. Not sure why it's not receiving
> the packet. Maybe it takes longer for xdpdump to load, and hence the
> packets arrive before it's ready? Maybe the sleep 1, needs to be sleep
> 2 in start_background()?

Hmm, would be great if we could rely less on random sleeps and more on
proper serialisation; but that would require xdpdump to signal readiness
somehow, I suppose?

shoracek:
Doesn't xdpdump print a string after it starts listening? Maybe just waiting until the output file contains that string could be enough.

chaudron:
There is the "listening on...." message which is printed when done setting up all stuff, but it might be buffered by the redirected output. We could use bpftool to see if the tracepoint is attached. All this assuming this is causing the failure ;)

tohojo:
Seems just increasing the sleep time at least fixed it for now :)

Issue #118, opened 2021-06-28
Test failure on new libbpf


tohojo:
Seems updating libbpf made the tests fail for xdpdump. @chaudron care to take a look?

https://github.com/xdp-project/xdp-tools/runs/2932140156?check_suite_focus=true

chaudron:
@tohojo submitted mr #119 , lets see if tests pass ;)

PR #119, opened 2021-06-29
xdpdump: Fix perf buffer API fixes to comply with new error checking.

chaudron:
libbpf v1 no longer overloads pointers with an error value instead,
it uses errno. This fix will change the perf_buffer API to take this
into account.

### DIFF
--- a/xdp-dump/xdpdump.c
+++ b/xdp-dump/xdpdump.c
@@ -1611,7 +1611,7 @@ static void unload_xdp_trace_program(struct dumpopt *cfg,
  *****************************************************************************/
 static bool capture_on_interface(struct dumpopt *cfg)
 {
-	int                          err;
+	int                          err, cnt;
 	bool                         rc = false;
 	bool                         load_xdp = false;
 	bool                         promiscuous = false;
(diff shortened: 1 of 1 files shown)


chaudron:
This commit should fix #118

tohojo:
This will break stuff on old libbpf, though, no? Weren't the libbpf API change supposed to be opt-in until 1.0?

chaudron:
Yes, this will break the old libbpf version, guess the update you did includes libbpf v1.0 changes?

chaudron:
BTW this was the commit causing the problem: https://github.com/libbpf/libbpf/commit/8e3a63ea48d2588d5f3626b45cb2a4eb4ae0ae83

tohojo:
But that commit is only supposed to change the visible behaviour if we set the forward-compatible mode. Which we don't, so why is it breaking stuff?

chaudron:
> But that commit is only supposed to change the visible behaviour if we set the forward-compatible mode. Which we don't, so why is it breaking stuff?

Not sure what you mean with " forward-compatible" but this commit directly changes the APIs (and you ACKed it ;).

tohojo:
Okay, so looking at this a bit closer, it seems the idea of the commit is to
change every function so that they can potentially return the error code
directly in the return value instead of setting errno, controlled by a new
setting libbpf_set_strict_mode(LIBBPF_STRICT_DIRECT_ERRS).

However, perf_buffer__poll() happened to already return the error code directly,
and because that is now hidden behind the switch, that patch unintentionally
changed the behaviour of that function.

I'm wondering if a better fix for this on the xdp-tools side is not to detect
the presence of libbpf_set_strict_mode() in './configure' and just set
LIBBPF_STRICT_DIRECT_ERRS if it's available? I think that should just make
everything work...

tohojo:
I sent a revert for the changed return code. However, the fix in this PR doesn't actually break anything with old versions of libbpf (because errno will still be set even if the return value is different), so let's just merge it as-is :)

Issue #123, opened 2021-09-09
libxdp: Get queue num from sys in xsk while driver not implement get_channels?


arikachen:
I found some driver has fixed queue num like ixgbevf and it not implement get_channels interface. 
Does we need to adapt this situation?

```
static int get_queue(const char* ifname) {
        DIR *dir;
        struct dirent *entry;
        char buf[64];
        int rx, tx;

        sprintf(buf, "/sys/class/net/%s/queues/", ifname);
        dir = opendir(buf);
        if(dir == NULL)
        {
                return 1;
        }
        while((entry=readdir(dir)))
        {
                if (0 == strncmp(entry->d_name, "rx", 2)) {
                        rx++;
                }
                if (0 == strncmp(entry->d_name, "tx", 2)) {
                        tx++;
                }
        }
        closedir(dir);
        // rx tx is zero ?
        return rx > tx ? rx : tx;
}

static int xsk_get_max_queues(struct xsk_socket *xsk)
...
if (err) {
        ret = get_queue(ctx->ifname) 
}
...
```

magnus-karlsson:
That is a good idea. Did not know there was support for ZC on a device without get_channels. Question is if it would not be easier to implement get_channels (but not set_channels) on that device. On the other hand, your approach would support users with older kernels/drivers which is good. So I say, submit a patch please. But use snprintf() to avoid buffer overruns. And "/* */" as comments :-).

Thanks: Magnus

PR #124, opened 2021-09-13
libxdp: get queue num from sysfs in xsk

arikachen:
Fixes: #123

Signed-off-by: arikachen <chenqijun@corp.netease.com>

### DIFF
--- a/lib/libxdp/xsk.c
+++ b/lib/libxdp/xsk.c
@@ -15,6 +15,7 @@
 #include <arpa/inet.h>
 #include <bpf/bpf.h>
 #include <bpf/libbpf.h>
+#include <dirent.h>
 #include <linux/err.h>
 #include <linux/ethtool.h>
 #include <linux/filter.h>
@@ -398,6 +399,36 @@ static enum xdp_attach_mode xsk_convert_xdp_flags(__u32 xdp_flags)
 	return XDP_MODE_NATIVE;
 }
 
+#define MAX_DEV_QUEUE_PATH_LEN 64
+
+static void xsk_get_queues_from_sysfs(const char* ifname, __u32 *rx, __u32 *tx) {
+	char buf[MAX_DEV_QUEUE_PATH_LEN];
+	struct dirent *entry;
+	DIR *dir;
+	int err;
+
+	*rx = *tx = 0;
+
+	err = try_snprintf(buf, MAX_DEV_QUEUE_PATH_LEN,
+			"/sys/class/net/%s/queues/", ifname);
+	if (err)
+		return;
+
+	dir = opendir(buf);
+	if(dir == NULL)
+		return;
+
+	while((entry = readdir(dir))) {
+		if (0 == strncmp(entry->d_name, "rx", 2))
+			++*rx;
+
+		if (0 == strncmp(entry->d_name, "tx", 2))
+			++*tx;
+	}
+
+	closedir(dir);
+}
+
 static int xsk_get_max_queues(char *ifname)
 {
 	struct ethtool_channels channels = { .cmd = ETHTOOL_GCHANNELS };
@@ -418,10 +449,13 @@ static int xsk_get_max_queues(char *ifname)
 	}
 
 	if (err) {
-		/* If the device says it has no channels, then all traffic
+		/* If the device says it has no channels, 
+		 * try to get rx tx from sysfs, otherwise all traffic
 		 * is sent to a single stream, so max queues = 1.
 		 */
-		ret = 1;
+		__u32 rx, tx;
+		xsk_get_queues_from_sysfs(ifr.ifr_name, &rx, &tx);
+		ret = max(max(rx, tx), 1);
 	} else {
 		/* Take the max of rx, tx, combined. Drivers return
 		 * the number of channels in different ways.


tohojo:
Please add the full description of the problem, and your solution, to the commit message as well (instead of just referencing the issue) :)

arikachen:
> Please add the full description of the problem, and your solution, to the commit message as well (instead of just referencing the issue) :)

done

lib/libxdp/xsk.c
```diff
@@ -398,6 +399,32 @@ static enum xdp_attach_mode xsk_convert_xdp_flags(__u32 xdp_flags)
 	return XDP_MODE_NATIVE;
 }
 
+#define MAX_DEV_QUEUE_PATH_LEN 64
+
+static int xsk_get_queues_from_sysfs(const char* ifname) {
+	DIR *dir;
+	struct dirent *entry;
+	char buf[MAX_DEV_QUEUE_PATH_LEN];
+	int rx = 0, tx = 0;
```
magnus-karlsson, line 408: Make these __u32 since negative values are not used.

lib/libxdp/xsk.c
```diff
@@ -398,6 +399,32 @@ static enum xdp_attach_mode xsk_convert_xdp_flags(__u32 xdp_flags)
 	return XDP_MODE_NATIVE;
 }
 
+#define MAX_DEV_QUEUE_PATH_LEN 64
+
+static int xsk_get_queues_from_sysfs(const char* ifname) {
+	DIR *dir;
```
magnus-karlsson, line 405: This is a comment for Toke I guess: are we using reverse Christmas tree in libxdp or not? If yes, then this is in the wrong order. If not, just leave it as is.

lib/libxdp/xsk.c
```diff
@@ -398,6 +399,32 @@ static enum xdp_attach_mode xsk_convert_xdp_flags(__u32 xdp_flags)
...
+		if (0 == strncmp(entry->d_name, "rx", 2))
+			rx++;
+
+		if (0 == strncmp(entry->d_name, "tx", 2))
+			tx++;
+	}
+
+	closedir(dir);
+
+	return max(max(rx, tx), 1);
```
magnus-karlsson, line 425: I think that this logic belongs to the calling function. This function should only do what it states: get_queues_from_sysfs (good name BTW) and if it cannot, return an error. Let the logic above set channels to 1 if all methods fail.

lib/libxdp/xsk.c
```diff
@@ -398,6 +399,32 @@ static enum xdp_attach_mode xsk_convert_xdp_flags(__u32 xdp_flags)
...
+static int xsk_get_queues_from_sysfs(const char* ifname) {
+	DIR *dir;
+	struct dirent *entry;
+	char buf[MAX_DEV_QUEUE_PATH_LEN];
+	int rx = 0, tx = 0;
+
+	snprintf(buf, MAX_DEV_QUEUE_PATH_LEN, "/sys/class/net/%s/queues/", ifname);
+	dir = opendir(buf);
+	if(dir == NULL)
+		return 1;
```
magnus-karlsson, line 413: Please see below.

magnus-karlsson (commented):
Looks good. Just a few small comments to make the code clearer. Thanks for detecting and fixing this!

lib/libxdp/xsk.c
tohojo: Yeah, please try to keep reverse x-mas tree :)

arikachen:
@magnus-karlsson @tohojo thanks for the reviews! I have update the pr, PTAL.

lib/libxdp/xsk.c
```diff
@@ -401,28 +401,29 @@ static enum xdp_attach_mode xsk_convert_xdp_flags(__u32 xdp_flags)
...
+static int xsk_get_queues_from_sysfs(const char* ifname, __u32 *rx, __u32 *tx) {
 	char buf[MAX_DEV_QUEUE_PATH_LEN];
-	int rx = 0, tx = 0;
+	struct dirent *entry;
+	DIR *dir;
 
-	snprintf(buf, MAX_DEV_QUEUE_PATH_LEN, "/sys/class/net/%s/queues/", ifname);
+	*rx = *tx = 0;
+
+	try_snprintf(buf, MAX_DEV_QUEUE_PATH_LEN, "/sys/class/net/%s/queues/", ifname);
```
tohojo, line 411: You still need to check the return code, it's not a magical function ;)

tohojo (changes requested):
Fixing things in a separate commit is fine for reviewing, but you'll need to squash everything into a single commit before merging... Also, one other comment below:

arikachen:
> Fixing things in a separate commit is fine for reviewing, but you'll need to squash everything into a single commit before merging... 

Thanks. I think [squash merge](https://docs.github.com/en/github/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/about-pull-request-merges) could do this.

tohojo:
On 14 September 2021 04:06:43 CEST, arikachen ***@***.***> wrote:
>> Fixing things in a separate commit is fine for reviewing, but you'll need to squash everything into a single commit before merging... 
>
>Thanks. I think [squash merge](https://docs.github.com/en/github/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/about-pull-request-merges) could do this.

Yes, but badly. We'd like everything in a single commit with a nicely formatted commit message and proper signoff. This requires manual squashing...

arikachen:
@tohojo done

tohojo:
Yeah, that's better! One more nit, though: please use your real name as commit author and in the signed-off-by tag :)

lib/libxdp/xsk.c
```diff
@@ -418,10 +451,13 @@ static int xsk_get_max_queues(char *ifname)
 	}
 
 	if (err) {
-		/* If the device says it has no channels, then all traffic
+		/* If the device says it has no channels, 
+		 * try to get rx tx from sysfs, otherwise all traffic
 		 * is sent to a single stream, so max queues = 1.
 		 */
-		ret = 1;
+		__u32 rx = 0, tx = 0;
```
magnus-karlsson, line 458: Not necessary to initialize these since you do this in xsk_get_queues_from_sysfs(). __u32 rx, tx; is enough.

lib/libxdp/xsk.c
```diff
@@ -398,6 +399,38 @@ static enum xdp_attach_mode xsk_convert_xdp_flags(__u32 xdp_flags)
 	return XDP_MODE_NATIVE;
 }
 
+#define MAX_DEV_QUEUE_PATH_LEN 64
+
+static int xsk_get_queues_from_sysfs(const char* ifname, __u32 *rx, __u32 *tx) {
```
magnus-karlsson, line 404: This should return void since you do not check for the return value anyway. Instead of return err below, just return. rx = 0 or tx = 0 tells the user that this function did not manage to do what it was supposed to do.

magnus-karlsson (changes requested):
I think these are the last comments from me. Once you fix these, I am ready to approve your contribution.

Thanks: Magnus

arikachen:
@tohojo @magnus-karlsson thanks.

tohojo (approved):
Alright, looks good, thanks! :)

PR #126, opened 2021-10-07
libxdp: use 3 as the minimum fd number when cloning fds

tohojo:
It seems[0] that there are some systems that will force-close fds 0, 1 and
2 if they don't refer to the regular std{in,out,err} streams. To make sure
we don't run afoul of this, let's use 3 as the minimum fd when cloning file
descriptors.

[0] https://lore.kernel.org/r/CAADnVQKVKY8o_3aU8Gzke443+uHa-eGoM0h7W4srChMXU1S4Bg@mail.gmail.com

### DIFF
--- a/lib/libxdp/libxdp.c
+++ b/lib/libxdp/libxdp.c
@@ -37,6 +37,14 @@
 
 #define ARRAY_SIZE(x) (sizeof(x) / sizeof(*(x)))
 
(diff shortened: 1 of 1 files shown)


rst0git (approved):
LGTM

tohojo:
Thanks for taking a look!

PR #131, opened 2021-10-28
xdpdump: only buffer in perf ring when using pcap_file

### LINKED ISSUE #130: xdpdump buffers packets when printing to stdout
It is confusing that xdpdump buffers packets (via `perf_attr.wakeup_events`) when printing to stdout.

When dumping packets into a PCAP file it is great for performance that we don't need to wakeup (perf-ring) for every packet, as I pointed out in #28, but this also seems to happen for the default case when xdpdump prints to stdout.

netoptimizer:
It is confusing that xdpdump buffers packets (via perf_attr.wakeup_events) when printing to stdout, as described in #130.

This fixes #130 by only buffering packet in the perf ring when cmdline option --write pcap_file have been specified.

Fixes: cbc47d53b275 ("xdp-dump: add support for multiple buffer wakeup")

### DIFF
--- a/xdp-dump/xdpdump.c
+++ b/xdp-dump/xdpdump.c
@@ -1747,11 +1747,8 @@ static bool capture_on_interface(struct dumpopt *cfg)
 
 	/* Determine the perf wakeup_events value to use */
 #ifdef HAVE_LIBBPF_PERF_BUFFER__CONSUME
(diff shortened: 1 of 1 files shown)


chaudron:
Thanks for fixing looks good to me!

PR #136, opened 2021-11-02
Fix libxdp build and linking errors

### LINKED ISSUE #135: Error when linking with libxdp
When linking with libxdp, two errors are generated:
```
cc -Wl,--no-as-needed -lxdp
/usr/bin/ld: /usr/local/lib/libxdp.so: undefined reference to `try_snprintf'
/usr/bin/ld: /usr/local/lib/libxdp.so: undefined reference to `silence_libbpf_logging'
```

objdump confirms they are undefined:
```
objdump -t /usr/local/lib/libxdp.so | grep try_snprintf
000000000000a710 l F .text 00000000000000da try_snprintf
0000000000000000 *UND* 0000000000000000 try_snprintf

objdump -t /usr/local/lib/libxdp.so | grep silence_libbpf
0000000000000000 *UND* 0000000000000000 silence_libbpf_logging
```

The call to try_snprintf was introduced to xsk.c in:
fdd1566dcd1f ("libxdp: get queue num from sysfs in xsk")

The call to silence_libbpf_logging in xsk.c was introduced in:
ea481a18ed07 ("libxdp: move AF_XDP functionality from libbpf")

tohojo:
Fix a couple of build and link errors for libxdp, and add a selftest to make
sure they don't resurface.

Fixes #135.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,7 +72,7 @@ install: all
 	do $(MAKE) -C $$i install; done
 
 test: all
-	@for i in $(UTILS); do \
+	@for i in lib/libxdp $(UTILS); do \
 		echo; echo test $$i; $(MAKE) -C $$i test; \
 		if [ $$? -ne 0 ]; then failed="y"; fi; \
 	done; \
--- a/headers/xdp/libxdp.h
+++ b/headers/xdp/libxdp.h
@@ -11,6 +11,7 @@
 
 #include <stdio.h>
 #include <linux/bpf.h>
+#include <bpf/libbpf.h>
 #include <bpf/bpf.h>
 #include "xdp_helpers.h"
 
--- a/lib/libxdp/Makefile
+++ b/lib/libxdp/Makefile
@@ -16,6 +16,10 @@ STATIC_OBJS := $(addprefix $(STATIC_OBJDIR)/,$(OBJS))
 STATIC_LIBS := $(OBJDIR)/libxdp.a
 MAN_PAGE := libxdp.3
 MAN_OBJ := ${MAN_PAGE:.3=.man}
+TEST_DIR := tests
+TEST_FILE := $(TEST_DIR)/test-libxdp.sh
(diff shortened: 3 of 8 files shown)


magnus-karlsson (approved):
Thanks for the fixing Toke. Removing the silence_libbpf_logging() will likely make libbpf barf out confusing messages to the user. I believe that is why I put it there in the first place. But I agree that we should try to use default libbpf settings. Let us see what Ciara says for her DPDK use case.

tohojo:
Right, we may want to add a way for users of the library to opt-in to libxdp taking over (and silencing) the libbpf logging. But doing it unconditionally is a bit rude I think ;)

cloftus:
Thanks for fixing @tohojo 

Some new libbpf logs pop up when using libxdp with DPDK:
libbpf: elf: skipping unrecognized data section(7) .xdp_run_config
libbpf: elf: skipping unrecognized data section(8) xdp_metadata
libbpf: elf: skipping unrecognized data section(8) xdp_metadata
libbpf: elf: skipping unrecognized data section(8) xdp_metadata

It would be nice to have the option to suppress those eventually.

tohojo:
Ugh, yeah, those are annoying... will see if we can figure out a way to silence those that are not too intrusive...

tohojo:
Hmm, we could add a new flag `XSK_LIBXDP_FLAGS__VERBOSE_LOGGING` and silence the logging unless this is set? Or we could try to install a filter function and filter out those particular messages but keep everything else intact? Is libbpf logs something people who use XSKs are generally interested in (for debugging)? And if so how do they get at them?

magnus-karlsson:
In the libbpf version, the logging is not silenced as it does not print the
warnings that we get with libxdp. Just silencing those warnings would be
helpful.

tohojo:
Actually for that particular error message I think the right thing to do is fix libbpf to not emit it. I seem to recall having such a patch rejected before, but with a more concrete reason for it maybe it'll fly; will give it a shot and otherwise look into filtering...

tohojo:
Submitted here: https://lore.kernel.org/r/20211104122911.779034-1-toke@redhat.com

PR #140, opened 2021-11-05
Include man pages in source code repository

### LINKED ISSUE #132: Include pre-generated man pages in the source code
As xdp-tools get packaged for more distributions, having emacs as a builddep turns out to be less than convenient, especially for things like Gentoo. So let's include pre-generated versions of the man pages in the source repo.

tohojo:
Fixes #132.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -54,7 +54,6 @@ dkms.conf
 config.mk
 xdp-dispatcher.c
 
-*.8
 *.man
 *.rpm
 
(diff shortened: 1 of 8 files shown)


zx2c4:
Does the RPM spec need to be updated too?

tohojo:
It doesn't have to: this change makes it possible to build without having emacs available, but if emacs *is* available, the man pages will still be generated on build...

shunghsiyu:
Thanks for this, makes packaging much easier :grinning:

PR #141, opened 2021-11-11
Build with clang

tstellar:
These patches make it possible to build xdp-tools with clang.


configure
```diff
@@ -29,7 +29,7 @@ check_toolchain()
     local clang_version
 
     : ${PKG_CONFIG:=pkg-config}
-    : ${CC=gcc}
+    : ${CC:=gcc}
```
tohojo, line 32: The inconsistency between `:=` and `=` is certainly a bit weird, but setting CC already works AFAICT?

```sh
CC=clang ./configure ; grep CC: config.mk     
CC:=clang
```

tstellar: Yes, you are right, this isn't needed.  I've dropped this patch.

tohojo:
Great, thanks for the fix!

PR #147, opened 2022-01-11
libxdp: fix program lookup for legacy

cloftus:
If the legacy loading prodedures are used, xsk_lookup_program will
fail even if there is a program loaded on the interface because
xdp_multiprog__next_prog fails. To rectify this introduce a check for
legacy mode which if true, will skip the xdp_multiprog__next_prog call
and ensure the legacy program is returned.

Fixes: ea481a18ed07 ("libxdp: move AF_XDP functionality from libbpf")

Signed-off-by: Ciara Loftus <ciara.loftus@intel.com>

### DIFF
--- a/.github/scripts/run_tests.sh
+++ b/.github/scripts/run_tests.sh
@@ -1,4 +1,4 @@
 #!/bin/bash
 
-make test >> TEST_OUTPUT 2>&1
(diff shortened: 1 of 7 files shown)


tohojo:
Thanks! I'll merge this after I unbork the selftests :)

PR #148, opened 2022-01-11
Libbpf compatibility fixes

tohojo:
With libbpf v0.7 a bunch of functions were deprecated. This branch updates the
submodule version of libbpf to the newest upstream and adds a bunch of
compatibility checks so we can switch to the new versions of the functions but
still keep compatibility with older versions of libbpf.


tohojo:
@magnus-karlsson and @chaudron could you please take a look at the XSK and xdpdump bits (and the rest, if you feel like it)? Thanks! :)

magnus-karlsson (approved):
Thanks for this fix Toke. Reviewed the changes to the xsk part and they look good to me.

lib/util/compat.h
```diff
@@ -0,0 +1,12 @@
+#ifndef __COMPAT_H
+#define __COMPAT_H
+
+#ifndef HAVE_LIBBPF_BTF__TYPE_CNT
+static __u32 btf__type_cnt(const struct btf *btf)
```
chaudron, line 5: I was wondering if there is a specific reason why you decided to inline them for libxdp, and not add them here?

xdp-dump/xdpdump.c
```diff
@@ -1797,13 +1796,22 @@ static bool capture_on_interface(struct dumpopt *cfg)
...
 
+#ifdef HAVE_LIBBPF_PERF_BUFFER__NEW_RAW
+	/* the configure check looks for the 6-argument variant of the function */
+	perf_buf = perf_buffer__new_raw(tgt_progs.progs[0].perf_map_fd,
+					PERF_MMAP_PAGE_COUNT,
+					&perf_attr, handle_perf_event,
+					&perf_ctx, NULL);
+#else
+	struct perf_buffer_raw_opts  perf_opts = {};
 	/* Setup perf ring buffers */
```
chaudron, line 1807: NIT: empty line after variable decliration.

chaudron (approved):
Changes look good to me. Will also look at #150 later.

lib/util/compat.h
tohojo: Because the code in `lib/libxdp` is supposed to be self-contained, so we can't include anything from `util/` in there...

xdp-dump/xdpdump.c
tohojo: Right, will fix.

tohojo:
Thanks for the speedy review both of you! :)

PR #152, opened 2022-01-12
Embed BPF programs into libxdp when compiling as a static library

tohojo:
When libxdp is being compiled as a static library, applications should be able
to link against it and work without any supporting files. However, the library
also includes XDP programs for the multiprog dispatcher as well as the XSK
stubs. And right now these are looked up in the file system, so if a static
binary is moved it will cease to work.

To fix this, this series embeds the BPF programs into separate sections of the
the .a file and loads them from there in the static version of the library. This
is similar to the 'bpf skeleton' generator in bpftool, but we use a different
approach to be able to keep the loading from the file system for the dynamic
library with as little additional code as possible, and to not rely on bpftool
in the build.

Unfortunately, the LLVM `ldd` linker doesn't support the mode we're using for
`ld` here, so this breaks building on clang. But maybe @tstellar has an idea for
how to restore that?

@magnus-karlsson or @cloftus could I trouble one of you to test that this
also works correctly for the AF_XDP code?


cloftus:
In DPDK we link with the .so. I tried a smoke test linking with the .a and it failed, even before your patches (likely an issue on my end and not libxdp itself). So I can't test your change immediately. But I think it's something worthwhile getting to work, so I'll hack away at it. It might be a little while before I have an answer for you.

cloftus:
I just needed to add -fPIC to libxdp. With that, linking the .a with DPDK works.
I verified that xdp_program__find_embedded succeeds and the embedded program is found and used.

tohojo:
Awesome! Thank you for testing! :)

PR #154, opened 2022-01-17
Release 1.2.1

tohojo:
Fixes and version bump for release 1.2.1

Cc @cloftus

### DIFF
--- a/configure
+++ b/configure
@@ -191,7 +191,6 @@ EOF
         echo "HAVE_FEATURES+=${config_var}" >>"$CONFIG"
         echo "yes"
     else
-        echo $libbpf_err
         echo "no"
     fi
 }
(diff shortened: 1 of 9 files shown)


cloftus:
Thanks Toke.
When I install libxdp the version reported by pkg-config is still v1.2.0.
I see in the Makefile that the libxdp version relies on libxdp.map which hasn't been bumped to v1.2.1. Should that be bumped too?

tohojo:
Ah, right. Hmm, we normally only add stuff to the .map file when we
actually introduce new symbols. But that is of course a problem for the
.pc file on bugfix releases like this one.

I don't think there's any great harm in keeping the .so filename as
1.2.0, so one way to fix this is to just use $TOOLS_VERSION in the .pc
file instead - WDYT?

cloftus:
Got it. That solution sounds good to me!

PR #156, opened 2022-01-20
Release v1.2.2

tohojo:
Another bugfix release to get the right version in the libxdp .pc file

Cc @cloftus


cloftus:
Thanks for creating another tag. LGTM.

PR #157, opened 2022-01-21
Auto release

tohojo:


.github/workflows/release.yml
```diff
@@ -0,0 +1,27 @@
...
+      - "v*"
+
+jobs:
+  tagged-release:
+    name: "Tagged Release"
+    runs-on: "ubuntu-latest"
+
+    steps:
+      - name: Check out repository code
+        uses: actions/checkout@v2
```
rst0git, line 16: @tohojo do we need to set `submodules`?

```yaml
    - uses: actions/checkout@v2
      with:
        submodules: recursive
```

tohojo: Ah, neat, yeah we should add this. Didn't know there was an option to do this, and the configure script has been doing the submodule checkout, hiding the issue for the selftest. Will add, thanks for the pointer!

PR #158, opened 2022-01-26
Import xdp-samples from kernel sources

kkdwivedi:
TODO:
- [x] Convert to use libxdp
- [x] Remove -F/--force
- [x] Tidy commit logs, describe changes
- [x] Add man pages


kkdwivedi:
All items done, so ready for final review.

lib/util/xdp_sample.c
```diff
@@ -0,0 +1,1671 @@
...
+	}
+
+	memcpy(mac_addr, ifr.ifr_hwaddr.sa_data, 6 * sizeof(char));
+
+end:
+	close(fd);
+	return r;
+}
+
+__attribute__((constructor)) static void sample_ctor(void)
```
tohojo, line 1664: Still not a fan of these constructors; let's move this to the main functions instead...

lib/util/xdp_sample.c
```diff
@@ -1203,13 +1208,17 @@ static int __sample_remove_xdp(int ifindex, __u32 prog_id, int xdp_flags)
...
 
-	return bpf_set_link_xdp_fd(ifindex, -1, xdp_flags);
+	return xdp_program__detach(xdp_prog, ifindex, xdp_flags, 0);
 }
 
-int sample_install_xdp(struct bpf_program *xdp_prog, int ifindex, bool generic,
-		       bool force)
+int sample_load_and_install_xdp(struct bpf_object *obj,
+				struct bpf_program *prog, int ifindex,
+				bool generic)
```
tohojo, line 1219: This function is way too complicated now; and mostly redundant as well: the caching of prog_id is not longer needed (and won't work the way it's being done now), because libxdp essentially does the same thing inside the xdp_prog objects. So I'd just get rid of this function entirely and let the callers call xdp_program__attach/detach().

xdp-redirect/xdp_redirect_dummy.bpf.c
```diff
@@ -0,0 +1,10 @@
+#include <bpf/vmlinux.h>
+#include <bpf/bpf_helpers.h>
+
+SEC("xdp")
+int xdp_redirect_dummy_prog(struct xdp_md *ctx)
+{
+	return XDP_PASS;
+}
```
tohojo, line 8: Instead of creating a whole new object + skeleton, let's just use the dummy program from the XDP dispatcher object, like:

```C
DECLARE_LIBXDP_OPTS(xdp_program_opts, opts, .prog_name = "xdp_pass", .find_filename = "xdp-dispatcher.o");
struct xdp_program *dummy_prog = xdp_program__create(&opts);
```

lib/util/Makefile
```diff
@@ -5,13 +5,17 @@ LIB_DIR ?= ..
...
-all: $(UTIL_OBJS)
+UTIL_OBJS_ALL := $(UTIL_OBJS) xdp_sample.o
+
+all: $(UTIL_OBJS_ALL)
 
 # Create expansions for dependencies
 UTIL_H := ${UTIL_OBJS:.o=.h}
 
-$(UTIL_OBJS): %.o: %.c $(UTIL_H) $(LIBMK)
+xdp_sample.o: ../../headers/xdp_sample.h ../../headers/xdp_sample_shared.h
```
tohojo, line 15: `headers/` is for header files we install; if they're internal, just put them in lib/util along with the .c and don't make this adjustment...

lib/common.mk
```diff
@@ -99,10 +99,13 @@ LIB_H := ${LIB_OBJS:.o=.h}
 $(LIB_OBJS): %.o: %.c %.h $(LIB_H)
 	$(Q)$(MAKE) -C $(dir $@) $(notdir $@)
 
+# Objects that just need to be appended
+LIB_OBJS += $(LIB_OBJS_NODEP)
```
tohojo, line 103: Erm, what? What are you trying to do here?

xdp-redirect/Makefile
```diff
@@ -0,0 +1,37 @@
...
+xdp_redirect_devmap.skel.h-deps := xdp_redirect_devmap.bpf.o
+xdp_redirect_devmap_multi.skel.h-deps := xdp_redirect_devmap_multi.bpf.o
+
+BPF_SRCS_LINKED := $(notdir $(wildcard *.bpf.c))
+BPF_OBJS_LINKED := $(patsubst %.bpf.c,%.bpf.o, $(BPF_SRCS_LINKED))
+
+BPFTOOL := /usr/sbin/bpftool
+$(LINKED_SKELS): $(BPF_OBJS_LINKED) $(BPFTOOL)
+	@echo "  GEN-SKEL" $(@:.skel.h=)
+	$(Q)$(BPFTOOL) gen skeleton $(@:.skel.h=.bpf.o) name $(notdir $(@:.skel.h=)) > $@
```
tohojo, line 42: This should go into common.mk

xdp-redirect/Makefile
```diff
@@ -0,0 +1,37 @@
...
+xdp_redirect_basic.skel.h-deps := xdp_redirect_basic.bpf.o
+xdp_redirect_cpumap.skel.h-deps := xdp_redirect_cpumap.bpf.o
+xdp_redirect_devmap.skel.h-deps := xdp_redirect_devmap.bpf.o
+xdp_redirect_devmap_multi.skel.h-deps := xdp_redirect_devmap_multi.bpf.o
+
+BPF_SRCS_LINKED := $(notdir $(wildcard *.bpf.c))
+BPF_OBJS_LINKED := $(patsubst %.bpf.c,%.bpf.o, $(BPF_SRCS_LINKED))
+
+BPFTOOL := /usr/sbin/bpftool
+$(LINKED_SKELS): $(BPF_OBJS_LINKED) $(BPFTOOL)
```
tohojo: Huh? A dependency on bpftool itself?

xdp-redirect/Makefile
```diff
@@ -0,0 +1,37 @@
...
+USER_EXTRA_C := xdp_redirect_basic.c xdp_redirect_cpumap.c xdp_redirect_devmap.c \
+		xdp_redirect_devmap_multi.c
+
+LIB_OBJS_NODEP := $(LIB_DIR)/util/xdp_sample.o
+
+LINKED_SKELS := xdp_redirect_basic.skel.h xdp_redirect_cpumap.skel.h xdp_redirect_devmap.skel.h \
+		xdp_redirect_devmap_multi.skel.h
+
+USER_GEN 	:= $(LINKED_SKELS)
+EXTRA_USER_DEPS := $(USER_GEN) $(USER_EXTRA_C)
```
tohojo: If you're teaching common.mk about `USER_EXTRA_C` anyway, this really shouldn't be needed...

tohojo (changes requested):
Some detailed comments below, and a few high-level questions/comments:

- After looking at the diffstat, I'm on the fence as to whether we should really be checking in the skeletons, or just accept bpftool as a hard build dependency. 
- The other utilities use the `command subcommand` structure, but the samples do not. Since this import time is most likely the only chance we'll have to make major breaking changes, I'm wondering if we should port over the samples to use the same structure (and the command-line parser from lib/util). Thoughts?

tohojo (approved):
Alright, this is basically ready for merging, assuming the tests pass! :)

PR #159, opened 2022-01-26
testing: add test-tool to testing framework with basic load

### LINKED ISSUE #150: Find better solution for xdp-dump test relying on bpftool
A bpftool update broke the xdpdump selftests, and while https://github.com/xdp-project/xdp-tools/commit/6bb0d7103e9674c8bc116e0574385f8381586312 works around the issue, we really should find a better solution since the `--legacy` flag to bpftool may be going away eventually...

chaudron:
This commit adds a test-tool compiled binary to the testing
framework. The tool allows specific "C" tool needs to be
bundled in a single binary.

The first tool needed was a simple loader to get rid of the
bpftool used by the xdp-dump self-test.

Fixes #150

### DIFF
--- a/lib/common.mk
+++ b/lib/common.mk
@@ -14,6 +14,8 @@ XDP_C = ${XDP_TARGETS:=.c}
 XDP_OBJ = ${XDP_C:.c=.o}
 USER_C := ${USER_TARGETS:=.c}
 USER_OBJ := ${USER_C:.c=.o}
+TEST_C := ${TEST_TARGETS:=.c}
+TEST_OBJ := ${TEST_C:.c=.o}
 XDP_OBJ_INSTALL ?= $(XDP_OBJ)
 MAN_FILES := $(MAN_PAGE)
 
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -42,6 +42,9 @@
 *.idb
 *.pdb
 
+# Test artifacts
+/lib/testing/test-tool
```
tohojo, line 46: nit: we have separate per-directory .gitignore files for the tools directories instead of putting entries into the top-level file; so maybe do the same here for consistency?

lib/testing/Makefile
```diff
@@ -1,6 +1,6 @@
 # SPDX-License-Identifier: (GPL-2.0 OR BSD-2-Clause)
 
-USER_TARGETS :=
+USER_TARGETS := test-tool
```
tohojo, line 3: With this, 'make install' will install the binary into `/usr/sbin`. I think common.mk will need a bit of adjustment to have it go into `/usr/share/xdp-tools` along with the tests instead...

tohojo (changes requested):
A few nits, but the general approach LGTM :)

tohojo:
@chaudron ping? upstream bpftool changes have started breaking CI, so it would be good to get this merged so we can just get rid of the dependency...

.gitignore
chaudron: Ack, will move it to a local .gitignore

chaudron:
@tohojo, this should address your concerns. I created a new tests_target that will build the binary and install it to the script directory. This way it will allow running all the test cases without any problems. Also removed the bpftool manual build.

lib/common.mk
```diff
@@ -113,6 +116,10 @@ $(XDP_OBJ): %.o: %.c $(KERN_USER_H) $(EXTRA_DEPS) $(BPF_HEADERS) $(LIBMK)
 	    -O2 -emit-llvm -c -g -o ${@:.o=.ll} $<
 	$(QUIET_LLC)$(LLC) -march=$(BPF_TARGET) -filetype=obj -o $@ ${@:.o=.ll}
 
+$(TEST_TARGETS): %: %.c  $(OBJECT_LIBBPF) $(LIBMK) $(KERN_USER_H)
+	$(QUIET_CC)$(CC) -Wall $(CFLAGS) $(LDFLAGS) -o $@ $(LIB_OBJS) \
+	 $< $(LDLIBS)
+
```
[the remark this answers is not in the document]
chaudron: And maybe call ALL_USER_TARGETS something like ALL_EXEC_TARGETS/ALL_BINARY_TARGETS?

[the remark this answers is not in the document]
tohojo: Alright, let's go with that; no strong feelings about the name, pick whatever you feel like :)

tohojo (approved):

Issue #153, opened 2022-01-12
Provide API to load programs by function name


tohojo:
Loading programs by section name is being deprecated in upstream libbpf in favour of function name-based loading. We should provide an API in libxdp for loading by function name as well to match this.

Opening this issue so we don't forget.

PR #161, opened 2022-02-07
libxdp: Introduce OPTS based xdp_program__create

kkdwivedi:
Since libbpf has deprecated custom section names and recommends usage of
bpf_object__find_program_by_name instead of bpf_program_by_section_name,
provide a new unified OPTS style xdp_program__create that can perform
XDP program creation using various options.

See commits for details.

Closes: https://github.com/xdp-project/xdp-tools/issues/153

### DIFF
--- a/.github/workflows/selftests.yml
+++ b/.github/workflows/selftests.yml
@@ -37,6 +37,11 @@ jobs:
           echo "deb http://apt.llvm.org/focal/ llvm-toolchain-focal-10 main" | sudo tee -a /etc/apt/sources.list
           sudo apt-get -qq update
           sudo apt-get -qq -y install clang-10 lld-10 llvm-10
+      - name: Install latest bpftool
+        run: |
+          git clone --depth=1 --recurse-submodules https://github.com/libbpf/bpftool bpftool
+          make LLVM_STRIP=llvm-strip-10 -C bpftool/src
+          sudo make install -C bpftool/src prefix=/usr
       - name: Compile
         run: make
       - name: Prepare test tools
--- a/headers/xdp/libxdp.h
+++ b/headers/xdp/libxdp.h
@@ -103,4 +103,54 @@ struct xdp_program *xdp_multiprog__hw_prog(const struct xdp_multiprog *mp);
 bool xdp_multiprog__is_legacy(const struct xdp_multiprog *mp);
 int xdp_multiprog__program_count(const struct xdp_multiprog *mp);
 
+/* Only following members can be set at once:
+ *
+ * @obj, @prog_name
+ *	Create using BPF program with name @prog_name in BPF object @obj
+ *
+ *	@prog_name is optional. In absence of @prog_name, first program of BPF
+ *	object is picked.
+ *
+ * @find_filename, @prog_name, @opts
+ *	Create using BPF program with name @prog_name in BPF object located in
+ *	LIBXDP_OBJECT_PATH with filename @find_filename, using
+ *	bpf_object_open_opts @opts.
+ *
+ *	@prog_name and @opts is optional. In absence of @prog_name, first
+ *	program of BPF object is picked.
+ *
+ * @open_filename, @prog_name, @opts
+ *	Create using BPF program with name @prog_name in BPF object located at
+ *	path @open_filename, using bpf_object_open_opts @opts.
+ *
+ *	@prog_name and @opts is optional. In absence of @prog_name, first
+ *	program of BPF object is picked.
+ *
+ * @id
+ *	Load from BPF program with ID @id
+ *
+ * @fd
+ *	Load from BPF program with fd @fd
+ *
+ * When one of these combinations is set, all other members of the opts struct
+ * must be zeroed out.
+ */
+struct xdp_program_opts {
+	size_t sz;
+	struct bpf_object *obj;
+	struct bpf_object_open_opts *opts;
+	const char *prog_name;
+	const char *find_filename;
+	const char *open_filename;
+	const char *pin_path;
+	__u32 id;
+	int fd;
+	size_t :0;
+};
+#define xdp_program_opts__last_field fd
+
+#define DECLARE_LIBXDP_OPTS DECLARE_LIBBPF_OPTS
+
+struct xdp_program *xdp_program__create(struct xdp_program_opts *opts);
+
 #endif
--- a/lib/libxdp/libxdp.c
+++ b/lib/libxdp/libxdp.c
@@ -95,6 +95,7 @@ static const char *xdp_action_names[] = {
 
 static struct xdp_program *xdp_program__create_from_obj(struct bpf_object *obj,
 							const char *section_name,
+							const char *prog_name,
 							bool external);
 
 #ifdef LIBXDP_STATIC
@@ -149,7 +150,7 @@ static struct xdp_program *xdp_program__find_embedded(const char *filename,
 		err = libbpf_get_error(obj);
 		if (err)
 			return ERR_PTR(err);
-		return xdp_program__create_from_obj(obj, section_name, false);
+		return xdp_program__create_from_obj(obj, section_name, NULL, false);
 	}
 
 	return NULL;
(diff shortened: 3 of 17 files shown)


headers/xdp/libxdp.h
```diff
@@ -103,4 +103,17 @@ struct xdp_program *xdp_multiprog__hw_prog(const struct xdp_multiprog *mp);
...
+struct xdp_program_opts {
+	size_t sz;
+	struct bpf_object *obj;
+	const char *prog_name;
+	const char *pin_path;
+	__u32 id;
+	int fd;
+	size_t :0;
+};
+#define xdp_program_opts__last_field fd
```
tohojo, line 115: Please add a comment describing which combinations of options are valid...

headers/xdp/libxdp.h
```diff
@@ -103,4 +103,17 @@ struct xdp_program *xdp_multiprog__hw_prog(const struct xdp_multiprog *mp);
 bool xdp_multiprog__is_legacy(const struct xdp_multiprog *mp);
 int xdp_multiprog__program_count(const struct xdp_multiprog *mp);
 
+struct xdp_program_opts {
+	size_t sz;
+	struct bpf_object *obj;
```
tohojo, line 108: What about loading from file? It should also be possible to replace xdp_program__{find,open}_file() with this...

lib/libxdp/libxdp_internal.h
```diff
@@ -62,4 +66,54 @@ LIBXDP_HIDE_SYMBOL struct xdp_program *xdp_program__clone(struct xdp_program *pr
 
 #define ARRAY_SIZE(x) (sizeof(x) / sizeof(*(x)))
 
+/* OPTS macros, from libbpf_internal.h */
+
+static inline bool libbpf_is_mem_zeroed(const char *p, ssize_t len)
```
tohojo, line 71: We generally rename stuff we copy from libbpf_ to have a libxdp_ prefix.

lib/libxdp/libxdp_internal.h
```diff
@@ -62,4 +66,54 @@ LIBXDP_HIDE_SYMBOL struct xdp_program *xdp_program__clone(struct xdp_program *pr
...
+		return false;
+	}
+	if (!libbpf_is_mem_zeroed(opts + opts_sz, (ssize_t)user_sz - opts_sz)) {
+		pr_warn("%s has non-zero extra bytes\n", type_name);
+		return false;
+	}
+	return true;
+}
+
+#define OPTS_VALID(opts, type)						      \
```
tohojo, line 97: What about `DECLARE_LIBBPF_OPTS()`? We should probably define a `DECLARE_LIBXDP_OPTS()` macro just for consistency...

tohojo (changes requested):
Looks mostly good with just a few minor nits. Could you please convert the internal users while you're at it? In particular we should convert the dispatcher to load by function name, and exercising the new function from at least one of the command line tools would be good as well so it gets test coverage...

Hmm, I guess we may need a separate test program that uses the old API as well, so we don't accidentally break that after we've converted all the tools...

headers/xdp/libxdp.h
kkdwivedi: For replacing those, the parameters look the same, should we have different triplets of 'filename, prog_name, object_open_opts' for find_file and open_file? I think that wastes too much space.

or maybe, just two more const char * find_filename and open_filename, and a object_open_opts, then reusing prog_name?

tohojo: Yeah, or just a single filename and a 'bool find'?

kkdwivedi:
Hm, any clue about the CI? I thought setting LLVM_STRIP env var would help, but apparently that's not enough. ISTM llvm-10 carries llvm-strip-10.

lib/libxdp/libxdp.c
```diff
@@ -1202,6 +1223,54 @@ struct xdp_program *xdp_program__from_pin(const char *pin_path)
...
+	} else if (id) {
+		if (obj || obj_opts || prog_name || find_filename || open_filename || pin_path || fd)
+			return ERR_PTR(-EINVAL);
+		return xdp_program__from_id(id);
+	} else if (fd) {
+		if (obj || obj_opts || prog_name || find_filename || open_filename || pin_path || id)
+			return ERR_PTR(-EINVAL);
+		return xdp_program__from_fd(fd);
+	}
+	return ERR_PTR(-EINVAL);
```
tohojo, line 1271: Maybe we should use this opportunity to get rid of the user-facing ERR_PTR() returns? I.e., return NULL and set errno here instead?

We'll have to keep the old behaviour for the existing helpers, but we could modify them to *also* set errno if there's an error before returning. And make sure libxdp_get_error/libxdp_strerror work in all cases...

.github/workflows/selftests.yml
```diff
@@ -37,6 +38,10 @@ jobs:
           echo "deb http://apt.llvm.org/focal/ llvm-toolchain-focal-10 main" | sudo tee -a /etc/apt/sources.list
           sudo apt-get -qq update
           sudo apt-get -qq -y install clang-10 lld-10 llvm-10
+      - name: Install latest bpftool
+        run: |
+          git clone --depth=1 --recurse-submodules https://github.com/libbpf/bpftool bpftool
+          sudo make LLVM_STRIP=llvm-strip-10 install -C bpftool/src prefix=/usr
```
tohojo, line 44: Running 'make' as root is a bad habit, even if this is done in a container. So maybe split this into `make` followed by `sudo make install`?

xdp-loader/README.org
```diff
@@ -69,6 +69,12 @@ Specify which ELF section to load the XDP program(s) from in each file. The
 default is to use the first program in each file. If this option is set, it
 applies to all programs being loaded.
 
+** -n, --prog-name <prog_name>
+Specify which BPF program with the name to load the XDP program(s) from in
+each file. The default is to use the first program in each file. Only one
+of --section and --prog-name may be specified, as both options are exclusive.
```
tohojo, line 75: "as both options are exclusive" is just restating "only one of... may be specified" - pick one :)

xdp-loader/xdp-loader.c
```diff
@@ -101,12 +111,22 @@ int do_load(const void *cfg, __unused const char *pin_root_path)
...
+		if (opt->prog_name) {
+			xdp_opts.open_filename = opt->filenames.strings[i];
+			xdp_opts.prog_name = opt->prog_name;
+			xdp_opts.opts = &opts;
+
+			p = xdp_program__create(&xdp_opts);
+		} else {
+			p = xdp_program__open_file(opt->filenames.strings[i],
+						   opt->section_name, &opts);
+		}
```
tohojo, line 129: Hmm, seems a bit weird to have to do this kind of if statement; maybe we should add 'section_name' as an opts member so we can use `xdp_program__create()` for everything?

tohojo (changes requested):
Only a few comments, some of which are more like questions that you should feel free to disagree with :)

lib/libxdp/libxdp.c
kkdwivedi: Should this change be made for direct error (integer) returning helpers as well? Like use introduce and use a libxdp_err, libxdp_ptr_err wrappers like libbpf_err, libbpf_err_ptr?

I think then we also need to wrap libbpf calls with it when returning, so that we set errno regardless of strict mode setting, agreed?

kkdwivedi: One thing that is not clear to me is, suppose that I change libxdp_get_error to be like libbpf_get_error:
```
   209 long libxdp_get_error(const void *ptr)                                                                            
   210 {                                                                                                   
   211         if (!IS_ERR_OR_NULL(ptr))                                                                   
   212                 return 0;                                                                           
   213                                                                                                     
   214         if (IS_ERR(ptr))                                                                            
   215                 errno = -PTR_ERR(ptr);                                                              
   216                                                                                                     
   217         /* If ptr == NULL, then errno should be already set by the failing                          
   218          * API, because libxdp never returns NULL on success and it now always                      
   219          * sets errno on error. So no extra errno handling for ptr == NULL                          
   220          * case.                                                                                    
   221          */                                                                                         
   222         return -errno;                                                                              
   223 }                 
```

Then for APIs that already return NULL (e.g. xdp_program__btf), it would be a problem if user does libxdp_get_error on it. It will end up using a stale errno value.

I think then we need to do an all or nothing switch, perhaps based on libbpf strict mode? i.e. add a libxdp_strict_mode setting?

tohojo (changes requested):
Alright, only a few nits with the errno conversion; thanks for the cleanup! :)

tohojo (approved):

PR #165, opened 2022-02-11
Make xdp-tools compliant to libbpf-0.7

meox:


tohojo:
Thank you for the patch!

Please add a commit message and a more appropriate subject line; and also we'll need version detection in `configure` and an appropriate wrapper function for old versions of libbpf. See, e.g., https://github.com/xdp-project/xdp-tools/commit/44490299cf5de96afe3a68189b894c89ebd8408a for an example of how to do that.

Also, while you're rebasing, please drop the single-character update to the xdp-loader man page from the commit :)

meox:
> Thank you for the patch!
> 
> Please add a commit message and a more appropriate subject line; and also we'll need version detection in `configure` and an appropriate wrapper function for old versions of libbpf. See, e.g., [4449029](https://github.com/xdp-project/xdp-tools/commit/44490299cf5de96afe3a68189b894c89ebd8408a) for an example of how to do that.
> 
> Also, while you're rebasing, please drop the single-character update to the xdp-loader man page from the commit :)

bpf_prog_test_run_opts is present since LIBPF 0.2 as far as see from https://github.com/libbpf/libbpf/blob/master/src/libbpf.map

tohojo:
Ah, right, wasn't aware of that. Okay, cool, no need for the detection, then...

tohojo:
Right, so this started breaking some of my other tests, so I took the liberty of fixing up a few things in your patch and just merging it...

@magnus-karlsson or @cloftus it would be great if one of you could double-check that it doesn't break anything on the XSK side (and follow-up with a fix if it does) :)

cloftus:
Tested with DPDK + libxdp and no issues detected.

tohojo:
Ciara Loftus ***@***.***> writes:

> Tested with DPDK + libxdp and no issues detected.

Great, thanks! :)

-Toke

meox:
@tohojo can you tag it? thanks!

tohojo:
Gian Lorenzo Meocci ***@***.***> writes:

> @tohojo can you tag it? thanks!

Sure, cherry-picked it into the v1.2 branch and tagged v1.2.3 :)

PR #172, opened 2022-02-23
xdp-dump: skip tests that need ftrace attach when not supported.

### LINKED ISSUE #162: Skipping xdpdump tests on unsupported archs
The xdpdump tests check for the `bpf_xdp_output_proto` kernel symbol and abort if it is not found. However, on non-x86 architectures, the symbol is there, but xdpdump still doesn't work; so we need a better test.

One option could be to just rely on the skip_if_legacy_fallback test, since multiprog support should mostly coincide with the required xdpdump features (I think?).

See https://bugzilla.redhat.com/show_bug.cgi?id=2032912#c15 and the comments following this for details of the failure.

chaudron:
Fixes #162

### DIFF
--- a/lib/testing/test_runner.sh
+++ b/lib/testing/test_runner.sh
@@ -69,6 +69,33 @@ OUTSIDE_MAC=
 ENABLE_IPV4=0
 ENABLE_VLAN=0
 
+is_trace_attach_supported()
+{
+    if [[ -z "${TRACE_ATTACH_SUPPORT:-}" ]]; then
+        [ -f "$STATEDIR/trace_attach_support" ] && \
+            TRACE_ATTACH_SUPPORT=$(< "$STATEDIR/trace_attach_support")
+
+        if [[ -z "${TRACE_ATTACH_SUPPORT:-}" ]]; then
+            RESULT=$($XDP_LOADER load -v "$NS" "$TEST_PROG_DIR/xdp_pass.o" 2>&1)
+            PID=$(start_background "$XDPDUMP -i $NS")
+            RESULT=$(stop_background "$PID")
+            if [[ "$RESULT" == *"The kernel does not support fentry function attach"* ]]; then
+                TRACE_ATTACH_SUPPORT="false"
+            else
+                TRACE_ATTACH_SUPPORT="true"
+            fi
+            echo "$TRACE_ATTACH_SUPPORT" > "$STATEDIR/trace_attach_support"
+            $XDP_LOADER unload "$NS" --all
+        fi
+    fi
+
+    if [[ "$TRACE_ATTACH_SUPPORT" == "true" ]]; then
+        return 0
+    else
+        return 1
+    fi
+}
+
 is_multiprog_supported()
 {
     if [[ -z "${MULTIPROG_SUPPORT:-}" ]]; then
@@ -98,6 +125,13 @@ skip_if_missing_kernel_symbol()
 skip_if_legacy_fallback()
 {
     if ! is_multiprog_supported; then
+        exit "$SKIPPED_TEST"
+    fi
+}
+
+skip_if_missing_trace_attach()
+{
+    if ! is_trace_attach_supported; then
         exit "$SKIPPED_TEST"
     fi
 }
@@ -250,7 +284,8 @@ cleanup()
     local statefiles=("$STATEDIR"/*.state)
 
     if [ "${#statefiles[*]}" -eq 1 ] && [ ! -e "${statefiles[0]}" ]; then
-        rm -f "${STATEDIR}/highest_num" "${STATEDIR}/current"
+        rm -f "${STATEDIR}/highest_num" "${STATEDIR}/current" \
+           "${STATEDIR}/trace_attach_support"
         rmdir "$STATEDIR"
     fi
 }
--- a/xdp-dump/xdpdump.c
+++ b/xdp-dump/xdpdump.c
@@ -50,6 +50,10 @@
 #define PROG_NAME "xdpdump"
 #define DEFAULT_SNAP_LEN 262144
 
+#ifndef ENOTSUPP
+#define ENOTSUPP         524 /* Operation is not supported */
+#endif
+
 #define RX_FLAG_FENTRY (1<<0)
 #define RX_FLAG_FEXIT  (1<<1)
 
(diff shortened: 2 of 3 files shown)


lib/testing/test_runner.sh
```diff
@@ -102,6 +121,13 @@ skip_if_legacy_fallback()
     fi
 }
 
+skip_if_missing_trace_attach()
+{
+    if ! is_trace_attach_supported; then
```
tohojo, line 126: This is called multiple times; should we cache the result?

xdp-dump/xdpdump.c
```diff
@@ -1415,8 +1419,14 @@ static bool load_and_attach_trace(struct dumpopt *cfg,
 		trace_link_fentry = bpf_program__attach_trace(trace_prog_fentry);
 		err = libbpf_get_error(trace_link_fentry);
 		if (err) {
-			pr_warn("ERROR: Can't attach XDP trace fentry function: %s\n",
-				strerror(-err));
+			if (err == -ENOTSUPP)
+				pr_warn("ERROR: Can't attach XDP trace fentry "
+					"function due to missing architectural "
+					"support!");
```
tohojo, line 1425: Could we make the error message a bit more concrete so users understand what's going on? 

We could detect the arch and do something like: `pr_warn("The kernel does not support fentry function attach %s", arch==x86 ? "because it is too old" : "on the current CPU architecture");` ?

tohojo (changes requested):
A few nits :)

lib/testing/test_runner.sh
chaudron: Yes, not sure why I removed that part, it was originally there as IO coped the below.

xdp-dump/xdpdump.c
chaudron: Will do

lib/testing/test_runner.sh
```diff
@@ -69,6 +69,27 @@ OUTSIDE_MAC=
 ENABLE_IPV4=0
 ENABLE_VLAN=0
 
+is_trace_attach_supported()
+{
+    if [[ -z "${TRACE_ATTACH_SUPPORT:-}" ]]; then
```
tohojo, line 74: This doesn't actually work since each test is executed in a subshell, so the variable assignment is lost for subsequent runs. Maybe write a file in `$STATEDIR` instead?

chaudron: Done, see the latest change...

tohojo (approved):
Right, so the changes LGTM; however, seems the CI is stuck. Could you try doing another force-push to the branch and see if that gets it unstuck?

tohojo:
Also, I'm getting another test failure on my laptop:

```
 make test
    Running tests from tests/test-xdpdump.sh
     [test_help]                   PASS
     [test_interfaces]             PASS
     [test_capt_pcap]              PASS
     [test_capt_pcapng]            FAIL
	PING fc42:dead:cafe:1::2(fc42:dead:cafe:1::2) 56 data bytes
	64 bytes from fc42:dead:cafe:1::2: icmp_seq=1 ttl=64 time=0.061 ms
	
	--- fc42:dead:cafe:1::2 ping statistics ---
	1 packets transmitted, 1 received, 0% packet loss, time 0ms
	rtt min/avg/max/mdev = 0.061/0.061/0.061/0.000 ms
	PING fc42:dead:cafe:1::2(fc42:dead:cafe:1::2) 56 data bytes
	64 bytes from fc42:dead:cafe:1::2: icmp_seq=1 ttl=64 time=0.088 ms
	
	--- fc42:dead:cafe:1::2 ping statistics ---
	1 packets transmitted, 1 received, 0% packet loss, time 0ms
	rtt min/avg/max/mdev = 0.088/0.088/0.088/0.000 ms
	../lib/testing/test_runner.sh: line 176: kill: (-26544) - No such process
	PING fc42:dead:cafe:1::2(fc42:dead:cafe:1::2) 56 data bytes
	64 bytes from fc42:dead:cafe:1::2: icmp_seq=1 ttl=64 time=0.063 ms
	
	--- fc42:dead:cafe:1::2 ping statistics ---
	1 packets transmitted, 1 received, 0% packet loss, time 0ms
	rtt min/avg/max/mdev = 0.063/0.063/0.063/0.000 ms
	../lib/testing/test_runner.sh: line 176: kill: (-26570) - No such process
	Running as user "root" and group "root". This could be dangerous.
	
	ERROR: Failed attributes content
     [test_capt_term]              PASS
     [test_exitentry]              PASS
     [test_snap]                   PASS
     [test_multi_pkt]              PASS
     [test_perf_wakeup]            PASS
     [test_promiscuous_selfload]   PASS
     [test_promiscuous_preload]    PASS
     [test_none_xdp]               PASS
     [test_pname_parse]            PASS
     [test_multi_prog]             PASS
     [test_xdp_load]               PASS
make: *** [../lib/common.mk:136: test] Error 1
```

Seems `tshark` is crashing?
```
tshark -r /tmp/xdptest_PID_23177_15008.pcap -Tfields -e frame.interface_name -e frame.interface_queue -e frame.packet_id -e frame.verdict.ebpf_xdp
xdptest:xdp_test_prog_with_a_long_name()@fentry	0	1	
[1]    27055 segmentation fault (core dumped)  tshark -r /tmp/xdptest_PID_23177_15008.pcap -Tfields -e frame.interface_name 

$ tshark --version
TShark (Wireshark) 3.6.2 (Git commit 626020d9b3c3)
```

Oh, and the pcap files are not deleted if the check fails:
```
ls /tmp/*.pcap
/tmp/xdptest_PID_20065_9570.pcap   /tmp/xdptest_PID_22183_20213.pcap  /tmp/xdptest_PID_24133_13759.pcap
/tmp/xdptest_PID_20534_31086.pcap  /tmp/xdptest_PID_22511_12072.pcap  /tmp/xdptest_PID_24540_4326.pcap
/tmp/xdptest_PID_21315_18472.pcap  /tmp/xdptest_PID_22865_15948.pcap  /tmp/xdptest_PID_25122_6958.pcap
/tmp/xdptest_PID_21656_28430.pcap  /tmp/xdptest_PID_23177_15008.pcap  /tmp/xdptest_PID_26337_19030.pcap
```

chaudron:
Pushed it again, and it looks like tshark is also crashing on my Fedora setup. Will dig into this later, and follow up. Guess this needs a wireshark fix.

Issue #173, opened 2022-03-10
Add `static` or `extern` to header-declared functions in `xsk.h`


Emanuesson:
I was switching recently from the libbpf-headers to the libxdp-headers from this repository. I recognized that all the `static`-keywords have been removed from the inline-functions. This makes it impossible to include the header `xdp/xsk.h` mulitple times in any binary without getting linker-warnings about "multiple definitions" of these functions.

The commit-message of the mentioned change (https://github.com/xdp-project/xdp-tools/commit/8ed7e26968445a676fac7b3ced97699390cd6245) states that
> extern inline is used as it's slightly better

However, neither `extern` nor `static` inline was used. Instead the specifier was removed which leads to the situation that the linker complains, if the header is included in several different C-files.

As the header defines structs and functions, it was necessary for me to include it more then once.

I personally propose to simply add `extern` to the function declarations, as this seems to fit the performance-goal and the exported symbols in libxdp best for now.

tohojo:
Ping @kkdwivedi

kkdwivedi:
@Emanuesson Can you draft a small test case showing the problem?

kkdwivedi:
Please give #174 a try, I think the problem is GNU inline semantics.

Emanuesson:
I'm still on it to craft a minimal example. However, I can verify that https://github.com/xdp-project/xdp-tools/pull/174 works to resolve the mentioned problems.

kkdwivedi:
No problem, I was able to reproduce it, it needs to be compiled using -std=gnu89 (or pass -fgnu89-inline for other standards) to hit this.

PR #174, opened 2022-03-10
xdp/xsk.h: Fix compilation with -std=gnu89

kkdwivedi:
Commit 8ed7e2696844 ("libxdp: export inline helpers as symbols in
libxdp") used C99 inline semantics to rely on inline without extern not
emitting a symbol, and extern inline declaration doing that instead. On
C89, these C99 inline semantics are not followed.

Instead of requiring users to migrate away from C89, we can still retain
compatibility by marking XDP_ALWAYS_INLINE as static inline, while
ensuring libxdp continues to expose these symbols.

Closes: #173
Fixes: 8ed7e2696844 ("libxdp: export inline helpers as symbols in libxdp")

### DIFF
--- a/headers/xdp/xsk.h
+++ b/headers/xdp/xsk.h
@@ -21,7 +21,11 @@
 extern "C" {
 #endif
 
+#ifdef __GNUC_STDC_INLINE__
 #define XDP_ALWAYS_INLINE inline __attribute__((__always_inline__))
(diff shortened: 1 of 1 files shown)


kkdwivedi:
Note that this does require libxdp to be compiled with C99 and above (which I figured was ok, since it is already using -std=gnu11).

PR #186, opened 2022-04-05
xsk: Don't leak xdp_multiprog reference

tohojo:
The xsk_lookup_program() function uses xdp_multiprog__get_from_ifindex()
without closing the resulting multiprog. Fix this by moving the call to
xdp_program__clone() inside xsk_lookup_program() so we can close the
multiprog instance before returning.


magnus-karlsson (approved):
Thanks for spotting this Toke. The change looks good to me.

PR #196, opened 2022-06-20
- Initialising variable to allow build with -O3

cowleyrTHG:
- This initialises a variable to 0, to allow libxdp to be built with -O3 optimisation.
- Without this initialisation, it fails due to 'version' potentially being used unitialised in the function.


tohojo:
Hmm, seems a bit odd that `-O3` would trigger this, but I don't see anything wrong with merging this anyway...

PR #200, opened 2022-06-28
configure: Don't check for bpf_set_link_xdp_fd_opts()

tohojo:
The configure script checks for the presence of the old_fd argument in
bpf_set_link_xdp_fd_opts() to determine if it can use the system libbpf.
However, this function has been deprecated in newer versions of libbpf,
which means we refuse to use system libbpf versions 0.8 and above.
Fortunately, we can just remove this check, as we also check for
bpf_map__set_initial_value() which was added after the old_fd support.

### DIFF
--- a/configure
+++ b/configure
@@ -260,12 +260,10 @@ check_libbpf()
 int main(int argc, char **argv) {
     void *ptr;
     DECLARE_LIBBPF_OPTS(bpf_object_open_opts, opts, .pin_root_path = "/path");
-    DECLARE_LIBBPF_OPTS(bpf_xdp_set_link_opts, xlopts, .old_fd = -1);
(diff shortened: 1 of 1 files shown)


dangowrt:
Confirming that xdp-tools now builds fine with this patch applied: https://github.com/openwrt/openwrt/pull/10139/commits/cd53c4aef40cfd3586f14dfbe42f21cb5ed1079d

dangowrt:
[Updated](https://github.com/openwrt/openwrt/pull/10139/commits/001b9cbf34650677fa21546f4b7f649da6da52ce) the [OpenWrt package build](https://github.com/openwrt/openwrt/pull/10139) which now works without any additional patches on top.

Now all we need is a tagged release :wink:

PR #205, opened 2022-07-08
libxdp: make shared xsk creation network namespace aware

Dhiver:
Creating AF_XDP sockets with the XDP_SHARED_UMEM flag set fails when two devices
ifindex (and the queue_id) are the same. This can happen with devices in
different network namespaces.

In the shared UMEM mode, each Xsk is expected to have a dedicated fill and
completion queue. Context data about those queues is stored in a linked-list
within the UMEM object.
The comparison performed to pick the right context for an Xsk prior to calling
bind() only takes into account the ifindex and queue_id of the network device.
However, the ifindex is not a "system-wide" unique identifier, it is unique only
within a network namespace. This results in attributing the same xsk_ctx to
different Xsks.

To fix this, we retrieve the netns_cookie associated with the socket at its
creation and add it in the comparison operation. This value "provides a global
identifier that can be assumed unique" ([0]).
We use the netns cookie instead of the netns inode number because inode numbers
can be reused.

  [0]: https://lore.kernel.org/bpf/c47d2346982693a9cf9da0e12690453aded4c788.1585323121.git.daniel@iogearbox.net/

### DIFF
--- a/lib/libxdp/xsk.c
+++ b/lib/libxdp/xsk.c
@@ -48,6 +48,12 @@
  #define PF_XDP AF_XDP
 #endif
 
+#ifndef SO_NETNS_COOKIE
+ #define SO_NETNS_COOKIE 71
+#endif
+
+#define INIT_NS 1
+
 struct xsk_umem {
 	struct xsk_ring_prod *fill_save;
 	struct xsk_ring_cons *comp_save;
@@ -67,6 +73,7 @@ struct xsk_ctx {
 	__u32 queue_id;
 	int refcount;
 	int ifindex;
+	__u64 netns_cookie;
 	int xsks_map_fd;
 	struct list_head list;
 	struct xdp_program *xdp_prog;
@@ -733,7 +740,7 @@ err_prog_load:
 	return err;
 }
 
-static struct xsk_ctx *xsk_get_ctx(struct xsk_umem *umem, int ifindex, __u32 queue_id)
+static struct xsk_ctx *xsk_get_ctx(struct xsk_umem *umem, __u64 netns_cookie, int ifindex, __u32 queue_id)
 {
 	struct xsk_ctx *ctx;
 
@@ -741,7 +748,7 @@ static struct xsk_ctx *xsk_get_ctx(struct xsk_umem *umem, int ifindex, __u32 que
 		return NULL;
 
 	list_for_each_entry(ctx, &umem->ctx_list, list) {
-		if (ctx->ifindex == ifindex && ctx->queue_id == queue_id) {
+		if (ctx->netns_cookie == netns_cookie && ctx->ifindex == ifindex && ctx->queue_id == queue_id) {
 			ctx->refcount++;
 			return ctx;
 		}
@@ -777,7 +784,7 @@ out_free:
 }
 
 static struct xsk_ctx *xsk_create_ctx(struct xsk_socket *xsk,
-				      struct xsk_umem *umem, int ifindex,
+				      struct xsk_umem *umem, __u64 netns_cookie, int ifindex,
 				      const char *ifname, __u32 queue_id,
 				      struct xsk_ring_prod *fill,
 				      struct xsk_ring_cons *comp)
@@ -801,6 +808,7 @@ static struct xsk_ctx *xsk_create_ctx(struct xsk_socket *xsk,
 		memcpy(comp, umem->comp_save, sizeof(*comp));
 	}
 
+	ctx->netns_cookie = netns_cookie;
 	ctx->ifindex = ifindex;
 	ctx->refcount = 1;
 	ctx->umem = umem;
@@ -866,6 +874,8 @@ int xsk_socket__create_shared(struct xsk_socket **xsk_ptr,
 	struct xsk_socket *xsk;
 	struct xsk_ctx *ctx;
 	int err, ifindex;
+	__u64 netns_cookie;
+	socklen_t optlen;
 	bool unmap;
 
 	if (!umem || !xsk_ptr || !(rx || tx))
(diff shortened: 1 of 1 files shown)


tohojo:
Note that the socket option to retrieve the netns cookie was only added by commit `e8b9eab99232 ("net: retrieve netns cookie via getsocketopt")` in kernel 5.14, so failing when that is not available breaks things on older kernels. Probably not a good idea to do that unconditionally...

tohojo:
How do you end up creating multiple sockets in different namespaces using the same umem object anyway?

lib/libxdp/xsk.c
```diff
@@ -898,14 +902,20 @@ int xsk_socket__create_shared(struct xsk_socket **xsk_ptr,
 		tx_setup_done = umem->tx_ring_setup_done;
 	}
 
-	ctx = xsk_get_ctx(umem, ifindex, queue_id);
+	optlen = sizeof(netns_cookie);
+	if (getsockopt(xsk->fd, SOL_SOCKET, SO_NETNS_COOKIE, &netns_cookie, &optlen) != 0) {
+		err = -errno;
+		goto out_socket;
+	}
```
alexforster, line 909: Given [this comment](https://github.com/xdp-project/xdp-tools/pull/205#issuecomment-1179375437), maybe we could conditionally compile this `getsockopt` call and instead set `netns_cookie` to 0 on kernels older than 5.14.

Question for maintainers: is use of the `KERNEL_VERSION` macro acceptable here?

alexforster:
> How do you end up creating multiple sockets in different namespaces using the same umem object anyway?

We're attaching AF_XDP to both sides of a veth pair and sharing a Umem between them. We use this setup to inspect/manipulate traffic going into and out of containers.

lib/libxdp/xsk.c
tohojo: Nope, can't use `KERNEL_VERSION`, features may be backported. You can just detect the return value at runtime, though, and fall back to setting a fixed value. Something like this:

https://github.com/xdp-project/bpf-examples/commit/e92fd14fafb97603546e82814450d92396cd3636

tohojo:
> > How do you end up creating multiple sockets in different namespaces using the same umem object anyway?
> 
> We're attaching AF_XDP to both sides of a veth pair and sharing a Umem between them. We use this setup to inspect/manipulate traffic going into and out of containers.

Hmm, interesting. What do you gain by sharing the umem instead of having completely separate attachments on each side?

lib/libxdp/xsk.c
Dhiver: Thank you for the example. I've updated the PR based on this feedback (see commit 2c9573f88deb1587630285b3567f82894aa2a955).

Dhiver:
> > > How do you end up creating multiple sockets in different namespaces using the same umem object anyway?
> > 
> > 
> > We're attaching AF_XDP to both sides of a veth pair and sharing a Umem between them. We use this setup to inspect/manipulate traffic going into and out of containers.
> 
> Hmm, interesting. What do you gain by sharing the umem instead of having completely separate attachments on each side?

This allows for zero-copy between the two veth. UMEM frames don't need to be copied from one UMEM to the other.

lib/libxdp/xsk.c
tohojo: Alright, cool! Please squash the fixup into the original commit, though... :)

tohojo:
> This allows for zero-copy between the two veth. UMEM frames don't need to be copied from one UMEM to the other.

Right, okay, but, erm, I must be missing something because by that description it seems like you have the same process open AF_XDP sockets inside and outside the container and zero-copy frames to itself through the veth pair? Seems a bit of a roundabout way of doing shared memory? :)

lib/libxdp/xsk.c
Dhiver: squashed

Dhiver:
> > This allows for zero-copy between the two veth. UMEM frames don't need to be copied from one UMEM to the other.
> 
> Right, okay, but, erm, I must be missing something because by that description it seems like you have the same process open AF_XDP sockets inside and outside the container and zero-copy frames to itself through the veth pair? Seems a bit of a roundabout way of doing shared memory? :)

We are moving frame **descriptors** between the TX and RX queues of the AF_XDP sockets that we created on both interfaces of the veth pair (which are in different netns). That way we can forward, send back or drop packets depending on which Xsk RX/TX queue we produce the descriptor. We can also edit the frame content from our userspace application. I hope that this makes more sense.

tohojo:
> > > This allows for zero-copy between the two veth. UMEM frames don't need to be copied from one UMEM to the other.
> > 
> > 
> > Right, okay, but, erm, I must be missing something because by that description it seems like you have the same process open AF_XDP sockets inside and outside the container and zero-copy frames to itself through the veth pair? Seems a bit of a roundabout way of doing shared memory? :)
> 
> We are moving frame **descriptors** between the TX and RX queues of the AF_XDP sockets that we created on both interfaces of the veth pair (which are in different netns). That way we can forward, send back or drop packets depending on which Xsk RX/TX queue we produce the descriptor. We can also edit the frame content from our userspace application. I hope that this makes more sense.

Ah, and the frames themselves are coming in on a physical NIC in the host namespace? That makes sense, then; interesting!

Anyway, I don't have any objections to merging this, but let's see what @magnus-karlsson says :)

magnus-karlsson:
I am all for this fix. Thanks!

tohojo (approved):

Dhiver:
For more context and visual representation @tohojo : https://blog.cloudflare.com/a-story-about-af-xdp-network-namespaces-and-a-cookie/ ;)

PR #206, opened 2022-07-09
openwrt: fix cross platform compiling.

PolynomialDivision:
Fix cross platform compiling.

### DIFF
--- a/lib/util/stats.c
+++ b/lib/util/stats.c
@@ -107,8 +107,8 @@ int stats_print(struct stats_record *stats_rec, struct stats_record *stats_prev)
 			return 0;
 
 		if (first) {
-			printf("Period of %fs ending at %lu.%06lu\n", period,
-			       t.tv_sec, t.tv_nsec / 1000);
+			printf("Period of %fs ending at %ld.%06ld\n", period,
+			       (long) t.tv_sec, (long) t.tv_nsec / 1000);
 			first = false;
 		}
 
--- a/lib/util/util.c
+++ b/lib/util/util.c
@@ -795,7 +795,7 @@ int prog_lock_get(const char *progname)
 	if (prog_lock_fd < 0) {
 		err = -errno;
 		if (err == -EEXIST) {
-			unsigned long pid = 0;
+			pid_t pid = 0;
 			char buf[100];
 			ssize_t len;
 			int fd;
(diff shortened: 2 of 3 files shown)


lib/util/stats.c
```diff
@@ -107,7 +108,7 @@ int stats_print(struct stats_record *stats_rec, struct stats_record *stats_prev)
 			return 0;
 
 		if (first) {
-			printf("Period of %fs ending at %lu.%06lu\n", period,
+			printf("Period of %fs ending at %" PRId64 ".%06lu\n", period,
```
tohojo, line 111: Let's be consistent and replace the second `lu` here as well...

tohojo (changes requested):
Thank you for the fixes! Only one nit...

PolynomialDivision:
@tohojo Done. Also switched to PRIu64 for unsigned.

lib/util/stats.c
PolynomialDivision: Does not work:
```
stats.c: In function 'stats_print':
stats.c:111:32: error: format '%llu' expects argument of type 'long long unsigned int', but argument 4 has type 'long int' [-Werror=format=]
  111 |                         printf("Period of %fs ending at %" PRIu64 ".%06" PRIu64 "\n", period,
      |                                ^~~~~~~~~~~~~~~~~~~~~~~~~~~
  112 |                                t.tv_sec, t.tv_nsec / 1000);
      |                                          ~~~~~~~~~~~~~~~~
      |                                                    |
      |                                                    long int
In file included from stats.c:9:
```

PolynomialDivision: I restructured now the PR:
https://github.com/xdp-project/xdp-tools/pull/206/commits/6ebbbb221b8edf19a4b6262a51fa83710a4e47c8

PolynomialDivision:
It was said that using PRIu64 is wrong:
https://github.com/openwrt/openwrt/pull/10229#issuecomment-1179573186

tohojo:
Right, so how about we just add casts to long and use `%ld` everywhere?

PolynomialDivision:
> Right, so how about we just add casts to long and use `%ld` everywhere?

Updated the PR

tohojo (approved):

tohojo:
Thank you for the fixes!

PR #209, opened 2022-07-24
Allow the build to be reproducible and to make use of hardening flags

bluca:


tohojo:
Thank you for the fixes!

The man pages are exported from the README.org files in each directory, so just editing them like you're doing in your first commit won't work, unfortunately. Some hacking of the emacs org-mode export is probably needed for fixing this.

bluca:
Ok, I've dropped that commit

tohojo:
Could you please confirm that #212 fixes the man page issue?

bluca:
> Could you please confirm that #212 fixes the man page issue?

Yes, thank you!

PR #221, opened 2022-08-23
headers: Update kernel btf.h header file

### LINKED ISSUE #220: libxdp doesn't build with libbpf 1.0
I don't think it's an issue on my side.  It worked with libbpf 0.7.0.

```
    CC       staticobjs/libxdp.o
In file included from libxdp.c:29:
include/bpf/btf.h:496:18: error: incomplete definition of type 'struct btf_enum64'
        return ((__u64)e->val_hi32 << 32) | e->val_lo32;
                       ~^
include/bpf/btf.h:489:22: note: forward declaration of 'struct btf_enum64'
static inline struct btf_enum64 *btf_enum64(const struct btf_type *t)
                     ^
```

To me that's a blocker for libbpf 1.0, as libxdp from inside of it can't be used anymore.

tohojo:
Libbpf 1.0.0 fails to build with older versions of the btf.h header file
from the kernel because of a missing struct btf_enum64 definition. This was
not added in the upstream kernel until the 6.0 cycle, so this header update
is not yet in any released version of the kernel. Technically this makes it
a bit premature to import this, but seeing as libbpf already introduced a
dependency on the header file, let's do so anyway to work around the buggy
upstream.

Fixes #220

### DIFF
--- a/headers/linux/btf.h
+++ b/headers/linux/btf.h
@@ -33,13 +33,13 @@ struct btf_type {
 	/* "info" bits arrangement
 	 * bits  0-15: vlen (e.g. # of struct's members)
 	 * bits 16-23: unused
-	 * bits 24-27: kind (e.g. int, ptr, array...etc)
-	 * bits 28-30: unused
+	 * bits 24-28: kind (e.g. int, ptr, array...etc)
+	 * bits 29-30: unused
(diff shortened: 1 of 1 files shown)


tohojo:
Great, thanks for testing!

PR #224, opened 2022-09-05
xdp-filter: Update examples in documentation

aibor:
The examples use the outdated flag `-w`. This commit replaces them with the
policy option to achieve the same funcitonality.

Signed-off-by: Tobias Böhm <code@aibor.de>


tohojo (approved):
Thanks for the fix!

PR #225, opened 2022-09-09
WIP: Adding golang interface for libxdp using swig

msherif1234:
Signed-off-by: Mohamed Mahmoud <mmahmoud@redhat.com>


msherif1234:
after internal chat we think using swig at this repo might be too much and instead build local pkg will do so will close this for now

PR #239, opened 2022-11-23
xdp-filter: update comment in xdpfilt_prog.h

dmitris:
PR updates the comment on top of xdp-filter/xdpfilt_prog.h as there is no longer a `xdp_filt_kern.c` file.


tohojo (approved):
Thank you for the fix!

PR #243, opened 2022-11-29
xdp-loader/README: Mention lack of support for HW mode in most cards

### LINKED ISSUE #170: xdp-loader will not perform hardware offload mode
When trying to perform `xdp-loader load -m hw ens224 ./xdp_prog_kern.o` I keep getting the following error:

```
Couldn't attach XDP program on iface 'ens224': Invalid argument(-22)
```
My xdp program works fine with generic and native modes. 
How can I troubleshoot this? Its not clear to me what `Invalid argument(-22)` means.

tohojo:
We don't document that the 'hw' mode is unsupported on most hardware. Add a
note to the man page explaining this.

Fixes #170

### DIFF
--- a/xdp-loader/README.org
+++ b/xdp-loader/README.org
@@ -54,7 +54,9 @@ the program to be offloaded to the hardware, or 'unspecified' which leaves it up
 to the kernel to pick a mode (which it will do by picking native mode if the
 driver supports it, or generic mode otherwise). Note that using 'unspecified'
 can make it difficult to predict what mode a program will end up being loaded
-in. For this reason, the default is 'native'.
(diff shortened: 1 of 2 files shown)


baloo:
A bit late to the party, but I'm confused about the claim that solarflare cards can run in HW mode (fair warning: I've never used solarflare cards).

My understanding is that it is limited to drivers which declare `NETDEV_XDP_ACT_HW_OFFLOAD` and the only driver that do so in linux at the moment appears to be nfp (for netronome cards).
This observation lines up with cilium's https://docs.cilium.io/en/latest/bpf/architecture/#offloads

Am I missing something?

PR #269, opened 2022-12-28
xdp-loader: Support setting pri/act at runtime.

Dongdong-Wang:
Support passing run priority and chain call actions at runtime, instead of using the static prio/actions in the ELF file. With these options, we can change the prio/actions as we want and no need to change the code and recompile them to ELF file.

Signed-off-by: Dongdong Wang <wangdongdong.6@bytedance.com>

### DIFF
--- a/xdp-loader/README.org
+++ b/xdp-loader/README.org
@@ -77,6 +77,19 @@ file. The default is to use the first program in each file. Only one of
 --section and --prog-name may be specified. If this option is set, it applies to
 all programs being loaded.
 
+** -P, --prio <priority>
+Specify the priority to load the XDP program(s) with (this affects the order of
+programs running on the interface). The default is to use the value from the metadata
+in the program ELF file, or a value of 50 if the program has no such metadata.
+If this option is set, it applies to all programs being loaded.
+
+** -A, --actions <actions>
+Specify the "chain call actions" of the loaded XDP program(s). These are the XDP
+actions that will cause the next program loaded on the interface to be called,
+instead of returning immediately. The default is to use the value set in the metadata
+in the program ELF file, or XDP_PASS if no such metadata is set. If this option is set,
+it applies to all programs being loaded.
+
 ** -v, --verbose
 Enable debug logging. Specify twice for even more verbosity.
 
--- a/xdp-loader/xdp-loader.c
+++ b/xdp-loader/xdp-loader.c
@@ -27,6 +27,8 @@ static const struct loadopt {
 	char *section_name;
 	char *prog_name;
 	enum xdp_attach_mode mode;
+	__u32 prio;
+	__u32 actions;
 } defaults_load = {
 	.mode = XDP_MODE_NATIVE
 };
@@ -39,6 +41,14 @@ struct enum_val xdp_modes[] = {
        {NULL, 0}
 };
 
+struct flag_val load_actions[] = {
+	{"XDP_ABORTED", 1U << XDP_ABORTED},
+	{"XDP_DROP", 1U << XDP_DROP},
+	{"XDP_PASS", 1U << XDP_PASS},
+	{"XDP_TX", 1U << XDP_TX},
+	{"XDP_REDIRECT", 1U << XDP_REDIRECT},
+	{}
+};
 
 static struct prog_option load_options[] = {
 	DEFINE_OPTION("mode", OPT_ENUM, struct loadopt, mode,
@@ -67,6 +77,14 @@ static struct prog_option load_options[] = {
 		      .metavar = "<filenames>",
 		      .required = true,
 		      .help = "Load programs from <filenames>"),
+	DEFINE_OPTION("prio", OPT_U32, struct loadopt, prio,
+		      .short_opt = 'P',
+		      .help = "Set run priority of program"),
+	DEFINE_OPTION("actions", OPT_FLAGS, struct loadopt, actions,
+		      .short_opt = 'A',
+		      .typearg = load_actions,
+		      .metavar = "<actions>",
+		      .help = "Chain call actions (default: XDP_PASS). e.g. XDP_PASS,XDP_DROP"),
 	END_OPTIONS
 };
 
@@ -137,6 +155,26 @@ retry:
 			pr_warn("Couldn't open file '%s': %s\n",
 				opt->filenames.strings[i], errmsg);
 			goto out;
+		}
+
+		if (opt->prio) {
+			err = xdp_program__set_run_prio(p, opt->prio);
+			if (err) {
+				pr_warn("Error setting run priority: %u\n", opt->prio);
+				goto out;
+			}
+		}
+
+		if (opt->actions) {
+			__u32 a;
+
+			for (a = XDP_ABORTED; a <= XDP_REDIRECT; a++) {
+				err = xdp_program__set_chain_call_enabled(p, a, opt->actions & (1U << a));
+				if (err) {
+					pr_warn("Error setting chain call action: %u\n", a);
+					goto out;
+				}
+			}
 		}
 
 		xdp_program__print_chain_call_actions(p, errmsg, sizeof(errmsg));


headers/xdp/libxdp.h
```diff
@@ -75,13 +75,21 @@ const struct btf *xdp_program__btf(struct xdp_program *xdp_prog);
 uint32_t xdp_program__id(const struct xdp_program *xdp_prog);
 int xdp_program__fd(const struct xdp_program *xdp_prog);
 unsigned int xdp_program__run_prio(const struct xdp_program *xdp_prog);
+void __xdp_program__set_run_prio(struct xdp_program *xdp_prog,
+			      unsigned int run_prio);
```
tohojo, line 79: Why are you adding these new functions? Just use the ones that are already exported?

headers/xdp/libxdp.h
```diff
@@ -75,13 +75,21 @@ const struct btf *xdp_program__btf(struct xdp_program *xdp_prog);
...
 				     enum xdp_action action);
+void __xdp_program__set_chain_call_enabled(struct xdp_program *prog,
+					unsigned int action,
+					bool enabled);
 int xdp_program__set_chain_call_enabled(struct xdp_program *prog,
 					unsigned int action,
 					bool enabled);
+void xdp_program__parse_chain_call_actions(const char *actions,
+					struct xdp_program *prog, bool val,
+					void (*cb)(struct xdp_program *, unsigned int, bool));
```
tohojo, line 92: Don't really see why this is needed either, see below...

xdp-loader/xdp-loader.c
```diff
@@ -67,6 +69,13 @@ static struct prog_option load_options[] = {
 		      .metavar = "<filenames>",
 		      .required = true,
 		      .help = "Load programs from <filenames>"),
+	DEFINE_OPTION("prio", OPT_U32, struct loadopt, prio,
+		      .short_opt = 'P',
+		      .help = "Set run priority of program"),
+	DEFINE_OPTION("actions", OPT_STRING, struct loadopt, actions,
+		      .metavar = "<actions>",
+		      .short_opt = 'A',
+		      .help = "Chain call actions (default: XDP_PASS). e.g. XDP_PASS,XDP_DROP"),
```
tohojo, line 78: Make this an OPT_FLAGS instead and you don't need the `parse_chain_call_actions()` function (the option parsing will handle that).

xdp-loader/xdp-loader.c
```diff
@@ -139,6 +148,20 @@ int do_load(const void *cfg, __unused const char *pin_root_path)
...
 
+		if (opt->prio)
+			__xdp_program__set_run_prio(p, opt->prio);
+
+		if (opt->actions) {
+			int a;
+
+			// clear old actions
+			for (a = 0; a <= XDP_REDIRECT; a++)
+				__xdp_program__set_chain_call_enabled(p, a, 0);
```
tohojo, line 159: If you parse the option as flags, this just becomes a setter for each one instead of  this clear-then-parse-and-set dance...

tohojo (changes requested):
Thanks for the patch! You're right we support this in xdp-loader as well. However, I think your patch is going about this in a bit of a roundabout way, see comments below...

headers/xdp/libxdp.h
Dongdong-Wang: I tried `xdp_program__set_run_prio`, but it failed. Because the `prog->prog_fd` >= 0. So I remove the check routine 
and adding the `__xdp_program__set_run_prio` function.

xdp-loader/xdp-loader.c
Dongdong-Wang: Okay, but when it becomes a settter, the obsolete chain call actions parsed from the ELF file will not be cleared. If you think this is the acceptable operation, I will remove the clear routine.

xdp-loader/xdp-loader.c
Dongdong-Wang: Yes! The OPT_FLAGS is the right choice, I will use that.

Dongdong-Wang:
I've already fixed the issues we talked about, please review them again if you have time.

headers/xdp/libxdp.h
tohojo: The checks are there for a reason: setting the values needs to be done before the program is loaded into the kernel, or they won't take effect. So exporting these functions is absolutely not the right way to fix this. At the point you're calling the code `prog->prog_fd` shouldn't be set, so I don't think that's why you're getting that error... My guess would be it's because you're passing in the bit-shifted value instead of just the action value itself?

xdp-loader/xdp-loader.c
```diff
@@ -139,6 +157,12 @@ int do_load(const void *cfg, __unused const char *pin_root_path)
 			goto out;
 		}
 
+		if (opt->prio)
+			__xdp_program__set_run_prio(p, opt->prio);
+
+		if (opt->actions)
+			__xdp_program__init_chain_call(p, opt->actions);
```
tohojo, line 164: This should be a loop, like:

```C
for (i = XDP_ABORTED; i <= XDP_REDIRECT; i++) {
    err = xdp_program__set_chain_call_enabled(p, i, opt->actions & (1U << i));
    if (err)
        goto out;
}
```

This will also correctly override the values loaded from the ELF file...

xdp-loader/xdp-loader.c
```diff
@@ -139,6 +157,12 @@ int do_load(const void *cfg, __unused const char *pin_root_path)
 			goto out;
 		}
 
+		if (opt->prio)
+			__xdp_program__set_run_prio(p, opt->prio);
```
tohojo, line 161: This should check the return value and error out if it fails (and just use the `xdp_program__set_run_prio()` variant).

tohojo (changes requested):
This definitely needs to be done without exposing any new functions from libxdp; see the comments below on how that should work. If this doesn't work please let me know exactly what's failing.

tohojo:
Oh, and also, please update the README file to document the new options...

headers/xdp/libxdp.h
Dongdong-Wang: No, run priority is just int value, not bits. I know the reason that `xdp_program__set_run_prio` returns error. I developed these features at commit a7f32c110cd2e2e98a26c64080573e5ade353a06. At that commit, `xdp_program__set_run_prio` used `prog->prog_fd` instead of `prog->prog_fd`. Actually, `prog->prog_fd` == -1, so that check fails.

Dongdong-Wang:
I've already fixed the issues we talked about, please review them again if you have time. Thanks!

xdp-loader/xdp-loader.c
```diff
@@ -139,6 +157,21 @@ int do_load(const void *cfg, __unused const char *pin_root_path)
 			goto out;
 		}
 
+		if (opt->prio)
+			xdp_program__set_run_prio(p, opt->prio);
```
tohojo, line 161: We should check the return code here as well...

xdp-loader/README.org
```diff
@@ -77,6 +77,16 @@ file. The default is to use the first program in each file. Only one of
 --section and --prog-name may be specified. If this option is set, it applies to
 all programs being loaded.
 
+** -P, --prio <priority>
+Specify with BPF program with the run priority to load the XDP program(s) from
+in each file. The default is to use the program metadata of ELF section. If this
```
tohojo, line 82: The first sentence reads a bit strangely. Maybe: "Specify the priority to load the XDP program(s) with (this affects the order of programs running on the interface)." 

We should also specify the default if no metadata is set for the program, so the second sentence should be "The default is to use the value from the metadata in the program ELF file, or a value of 50 if the program has no such metadata."

xdp-loader/README.org
```diff
@@ -77,6 +77,16 @@ file. The default is to use the first program in each file. Only one of
...
+option is set, it applies to all programs being loaded.
+
+** -A, --actions <actions>
+Specify with BPF program with the "chain actions" to load the XDP program(s) from
+in each file. The default is to use the program metadata of ELF section. If this
+option is set, it applies to all programs being loaded.
```
tohojo, line 88: As above; here we could replace the first two sentences with: "Specify the "chain call actions" of the loaded XDP program(s). These are the XDP actions that will cause the next program loaded on the interface to be called, instead of returning immediately. The default is to use the value set in the metadata in the program elf file, or XDP_PASS if no such metadata is set."

tohojo (changes requested):
Better! Just a few more nits, then we're good to go with this :)

Dongdong-Wang:
I've already fixed the nits we talked about, please review them again if you have time. Thanks!

PR #276, opened 2023-01-10
libxdp: Supress unnecessary warning when converting xdp flags.

Martzki:
The marco `XDP_FLAGS_SUPPORTED` is only used in `xsk_convert_xdp_flags` and only contains flags about XDP mode.  
This would lead to warning output when invoker passes other flag like `XDP_FLAGS_UPDATE_IF_NOEXIST`.  
Let's replace it with `XDP_FLAGS_MASK` to supress such unnecessary warning. :)

### DIFF
--- a/lib/libxdp/xsk.c
+++ b/lib/libxdp/xsk.c
@@ -388,12 +388,10 @@ static int xsk_init_xsk_struct(struct xsk_socket *xsk, int ifindex)
 	return 0;
 }
 
-#define XDP_FLAGS_SUPPORTED (XDP_FLAGS_SKB_MODE | XDP_FLAGS_DRV_MODE | XDP_FLAGS_HW_MODE)
-
 static enum xdp_attach_mode xsk_convert_xdp_flags(__u32 xdp_flags)
 {
-	if (xdp_flags & ~XDP_FLAGS_SUPPORTED)
-		pr_warn("XDP flag not supported by libxdp.\n");
+	if (xdp_flags & ~XDP_FLAGS_MASK)
+		pr_warn("XDP flag: 0x%x contains flags not supported by libxdp.\n", xdp_flags);
 
 	if (xdp_flags & XDP_FLAGS_SKB_MODE)
 		return XDP_MODE_SKB;


heshengyuan1311:
LGTM

tohojo:
Ping @magnus-karlsson  - what was the rationale for the stricter filtering?

magnus-karlsson:
Nope, that is a bug. Replacing it with XDP_FLAGS_MASK is correct. This was the intended functionality. Thanks for spotting this!

lib/libxdp/xsk.c
```diff
@@ -388,12 +388,10 @@ static int xsk_init_xsk_struct(struct xsk_socket *xsk, int ifindex)
...
 }
 
-#define XDP_FLAGS_SUPPORTED (XDP_FLAGS_SKB_MODE | XDP_FLAGS_DRV_MODE | XDP_FLAGS_HW_MODE)
-
 static enum xdp_attach_mode xsk_convert_xdp_flags(__u32 xdp_flags)
 {
-	if (xdp_flags & ~XDP_FLAGS_SUPPORTED)
-		pr_warn("XDP flag not supported by libxdp.\n");
+	if (xdp_flags & ~XDP_FLAGS_MASK)
+		pr_warn("XDP flag: 0x%x not supported by libxdp.\n", xdp_flags);
```
tohojo, line 394: If we're printing the unsupported flags, shouldn't that be the masked value?

Martzki: The `xdp_flags` is usually from invoker, the printing here means to notify invoker the total flags are not fully supported.  
Whether it should be masked value depends on the perspective of the message we're printing.  
For me I think the whole value is better. :grin:

tohojo: Right, but then we should change the message so it's clearer. Something like: "XDP flags 0x%x contains flag values not supported by libxdp" or something like that...

lib/libxdp/xsk.c
```diff
@@ -388,12 +388,10 @@ static int xsk_init_xsk_struct(struct xsk_socket *xsk, int ifindex)
...
 }
 
-#define XDP_FLAGS_SUPPORTED (XDP_FLAGS_SKB_MODE | XDP_FLAGS_DRV_MODE | XDP_FLAGS_HW_MODE)
-
 static enum xdp_attach_mode xsk_convert_xdp_flags(__u32 xdp_flags)
 {
-	if (xdp_flags & ~XDP_FLAGS_SUPPORTED)
-		pr_warn("XDP flag not supported by libxdp.\n");
+	if (xdp_flags & ~XDP_FLAGS_MASK)
+		pr_warn("XDP flag: 0x%x contains unsupported flag by libxdp.\n", xdp_flags);
```
tohojo, line 394: Better, but nit: "contains flags unsupported by libxdp"

Martzki: Adjusted again.

tohojo (approved):
Great, thanks!

PR #289, opened 2023-02-07
Fix compile error due to variable declaration inside a 'switch' case

salcock:
A recently added 'switch' statement for handling different dispatcher types includes a line of code where a variable is declared immediately after a case label without being enclosed in braces.

This causes compilation failures with certain compilers, which appear as:

    libxdp.c: In function 'check_dispatcher_version':
    libxdp.c:2189:3: error: a label can only be part of a statement and a declaration is not a statement
     2189 |   struct xdp_dispatcher_config_v1 *config = (void *)buf;
          |   ^~~~~~

Adding braces around the code for the case will avoid this problem.

Observed when following the standard build instructions on Ubuntu 20.04 (gcc 9.4.0, clang 15.0.7).

### DIFF
--- a/lib/libxdp/libxdp.c
+++ b/lib/libxdp/libxdp.c
@@ -2186,6 +2186,7 @@ static int check_dispatcher_version(struct xdp_multiprog *mp,
 
 	switch (version) {
 	case XDP_DISPATCHER_VERSION_V1:
+	{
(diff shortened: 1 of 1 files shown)


tohojo:
Huh, interesting. Thanks for the fix!

tohojo (approved):

PR #301, opened 2023-03-04
make: disable stack protector for BPF bits built by clang

### LINKED ISSUE #299: Build error in bpf bits with clang & default-enabled stack protector
When the clang compiler has stack protection default-enabed (e.g. via the new config file support in clang-15), the build fails for obvious reasons:
```
holger>make

lib

  libxdp
    CC       staticobjs/libxdp.o
    CC       staticobjs/xsk.o
    M4       xdp-dispatcher.c
    CLANG    xdp-dispatcher.o
    LLC      xdp-dispatcher.o
    GEN      xdp-dispatcher.embed.o
    CLANG    xsk_def_xdp_prog.o
    LLC      xsk_def_xdp_prog.o
    GEN      xsk_def_xdp_prog.embed.o
    CLANG    xsk_def_xdp_prog_5.3.o
    LLC      xsk_def_xdp_prog_5.3.o
error: <unknown>:0:0: in function xsk_def_prog i32 (ptr): A call to built-in function '__stack_chk_fail' is not supported.
```
I recently fixed the same problem in mainline bpftool; the solution is simple: add `-fno-stack-protector` to the CFLAGS used with clang.

hhoffstaette:
The clang toolchain might have stack-protection enabled by default (e.g. via platform configuration) and that won't work for BPF, so unconditionally disable it via -fno-stack-protector.
Fixes #299

### DIFF
--- a/lib/defines.mk
+++ b/lib/defines.mk
@@ -1,5 +1,5 @@
 CFLAGS ?= -O2 -g
-BPF_CFLAGS ?= -Wno-visibility
+BPF_CFLAGS ?= -Wno-visibility -fno-stack-protector
 BPF_TARGET ?= bpf
 
 HAVE_FEATURES :=


tohojo (approved):
Thank you for the fix! :)

PR #332, opened 2023-06-08
Minor packaging improvements

shunghsiyu:
Some minor packaging-related changes while trying to package v1.3.1:
- Install BPF/XDP objects as non-executable to keep their symbol table from being stripped (only tested on openSUSE's build system though)
- Support GCC 7 and earlier
- Add a minimum version requirement for libbpf

### DIFF
--- a/lib/common.mk
+++ b/lib/common.mk
@@ -73,7 +73,7 @@ install: all install_local
 	install -m 0755 -d $(DESTDIR)$(SBINDIR)
 	install -m 0755 -d $(DESTDIR)$(BPF_OBJECT_DIR)
 	$(if $(USER_TARGETS),install -m 0755 $(USER_TARGETS) $(DESTDIR)$(SBINDIR))
-	$(if $(XDP_OBJ_INSTALL),install -m 0755 $(XDP_OBJ_INSTALL) $(DESTDIR)$(BPF_OBJECT_DIR))
+	$(if $(XDP_OBJ_INSTALL),install -m 0644 $(XDP_OBJ_INSTALL) $(DESTDIR)$(BPF_OBJECT_DIR))
 	$(if $(MAN_FILES),install -m 0755 -d $(DESTDIR)$(MANDIR)/man8)
 	$(if $(MAN_FILES),install -m 0644 $(MAN_FILES) $(DESTDIR)$(MANDIR)/man8)
 	$(if $(SCRIPTS_FILES),install -m 0755 -d $(DESTDIR)$(SCRIPTSDIR))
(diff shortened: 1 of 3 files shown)


tohojo:
Thank you for the patches!

Are there any side effects to installing the BPF objects as non-executable? Logically they kinda-sorta are executable (i.e., they contain code that will be executed...). I *think* that is why we went with that originally, although TBH I don't quite recall the details.

Also, please drop the last commit, we added detection for that in the configure script, so the current git version should build correctly for older versions of libbpf again, now...

shunghsiyu:
> Thank you for the patches!
> 
> Are there any side effects to installing the BPF objects as non-executable? Logically they kinda-sorta are executable (i.e., they contain code that will be executed...). I _think_ that is why we went with that originally, although TBH I don't quite recall the details.

While they contain executable code, I think the generally expectation of users is that a file with executable bit set can be directly executed just by typing it's path in shell; but these BPF objects can't be use that way (without additional work such as setting up `binfmt_misc` or introducing some more code). So I'd say installing them as non-executable is less confusing for the user.

To the best of my knowledge, libbpf seems to not care about the whether a BPF object is an executable or not, so where side effect would pop up is probably in the packaging system. For me the side effect of symbol table being preserved is a welcomed one, but I'm not sure if the same can be said for other.

> Also, please drop the last commit, we added detection for that in the configure script, so the current git version should build correctly for older versions of libbpf again, now...

Sorry that I miss those changes, dropped the last patch.

shunghsiyu:
(commenting here since it's not entirely a issue with xdp-tools itself, but I can open a new one if needed)

Hi @tohojo,

I ran into issue with `strip` recently, exactly the same issue as the one mentioned in [Bug 2245296 - 'strip' barfing on BPF files again - "unsupported relocation type 0x4"](https://bugzilla.redhat.com/show_bug.cgi?id=2245296).

IIRC d013587b52317e3c6267e7fef0a5e7b53c2afe60 was added to keep BPF object files from being stripped, which is important for preserving their symbol table; and should have avoided such issue (for good or bad).

Given `strip` is barfing on BPF object files, it looks like d013587b52317e3c6267e7fef0a5e7b53c2afe60 may not be doing anything and BPF object files get stripped anyway? (haven't look too deep into this yet, just wondering if you have already worked this out)

shunghsiyu:
> Given `strip` is barfing on BPF object files, it looks like [d013587](https://github.com/xdp-project/xdp-tools/commit/d013587b52317e3c6267e7fef0a5e7b53c2afe60) may not be doing anything and BPF object files get stripped anyway? (haven't look too deep into this yet, just wondering if you have already worked this out)

Okay, they do prevent symbol table from being stripped, at least on the openSUSE build system. While BPF object files still [get passed as argument to `strip`, debuginfo is preserved](https://github.com/openSUSE/brp-check-suse/blob/e36182d/brp-15-strip-debug#L34-L38) as long as they are not an executable.

tohojo:
> Not sure if similar situation happens on the Fedora build system.

Looks like the Fedora scripts just strip everything that looks like an ELF binary: https://github.com/rpm-software-management/rpm/blob/master/scripts/brp-strip

shunghsiyu:
Hmm... so `strip -g` yet the symbol table on the BPF object file stays. I'll do a bit more digging. Thanks for the pointer to brp-strip.

PR #338, opened 2023-06-28
libxdp: Allow creation of xsk with exclusive umem without CAP_NET_RAW

igsilya:
Adding a new API xsk_umem__create_with_fd() that accepts an extra file descriptor of already open AF_XDP socket.  This function can be used as a substitute of a regular xsk_umem__create() in case where process doesn't have CAP_NET_RAW privileges.  Privileged process may open a socket beforehand and send the file descriptor via UNIX domain socket or pass it during exec().  Subsequent xsk_socket__create() will re-use this socket.  No other operations require privileges in case the program load is inhibited.

This change allows use of AF_XDP with libxdp by processes with no extra privileges.

xsk_socket__create_shared() requires opening a new socket for each queue after the first one, so it will require more work in order to create a pretty API.  Not doing that right now.  The target use-case is AF_XDP network backend in QEMU that doesn't need shared umem, but requires running without any privileges.

### DIFF
--- a/lib/libxdp/xsk.c
+++ b/lib/libxdp/xsk.c
@@ -306,10 +306,11 @@ out_mmap:
 	return err;
 }
 
-int xsk_umem__create(struct xsk_umem **umem_ptr, void *umem_area,
-		     __u64 size, struct xsk_ring_prod *fill,
-		     struct xsk_ring_cons *comp,
-		     const struct xsk_umem_config *usr_config)
+int xsk_umem__create_with_fd(struct xsk_umem **umem_ptr, int fd,
+			     void *umem_area, __u64 size,
+			     struct xsk_ring_prod *fill,
+			     struct xsk_ring_cons *comp,
+			     const struct xsk_umem_config *usr_config)
 {
 	struct xdp_umem_reg mr;
 	struct xsk_umem *umem;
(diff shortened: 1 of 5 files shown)


igsilya:
The Ci failure seems unrelated.

lib/libxdp/xsk.c
```diff
@@ -363,6 +364,25 @@ int xsk_umem__create(struct xsk_umem **umem_ptr, void *umem_area,
...
+
+int xsk_umem__create_with_fd(struct xsk_umem **umem_ptr, int fd,
+			     void *umem_area, __u64 size,
+			     struct xsk_ring_prod *fill,
+			     struct xsk_ring_cons *comp,
+			     const struct xsk_umem_config *usr_config)
+{
+	return __xsk_umem_create(umem_ptr, fd, umem_area, size,
+				 fill, comp, usr_config);
+}
```
tohojo, line 384: This wrapper seems a bit pointless? Why not just have `xsk_umem__create()` call the `_with_fd()` variant directly?

igsilya: It seemed easier to read this way, but I can change, no problem.

igsilya: done.

tohojo (approved):

tohojo:
Alright, LGTM. I'll merge this now, but also take the opportunity to point out that we really don't have any selftest coverage for any of the AF_XDP code; care to look into that?

igsilya:
> Alright, LGTM. I'll merge this now, but also take the opportunity to point out that we really don't have any selftest coverage for any of the AF_XDP code; care to look into that?

Looks like test_xsk_refcnt.c is covering some of the basic functionality.

I wrote a mall C program to test ability to create sockets from a non-privileged process.  I guess, I can try to turn it into a test.  It has a dependency on a libcap-ng though, but I guess we could try to detect the library during configuration and skip the test if not present?  What do you think?

tohojo:
That would be awesome! We already do plenty of feature checking and skipping of tests in various places, so don't think that's a problem :)

PR #340, opened 2023-07-07
libxdp: Add selftest for non-privileged xsk creation

igsilya:
Test loads a program, creates bare AF_XDP socket, then drops all the capabilities and tries to create umem and xsk with previously created resources.

Uses libcap-ng as a dependency for capability management.  Skipped if the library is not available at the configuration time.

### DIFF
--- a/.github/workflows/selftests.yml
+++ b/.github/workflows/selftests.yml
@@ -41,7 +41,7 @@ jobs:
       - name: Prepare packages
         run: |
           sudo apt-get update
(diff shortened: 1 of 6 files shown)


igsilya:
Existing `test_xsk_prog_refcnt` tests seem flaky.

tohojo:
> Existing `test_xsk_prog_refcnt` tests seem flaky.

Yeah. If you want to take a stab at improving that, I'd me much obliged :)

tohojo (approved):
LGTM; thanks for doing this :)

Issue #302, opened 2023-03-06
Odd warning from check_abi: Warning: Num of global symbols in libxdp.o xsk.o (68) does NOT match with num of versioned symbols in libxdp.so (71)


hhoffstaette:
After getting 1.3.1 to build properly (thanks!) the Gentoo CI promptly found .. "something", but we're not really sure what causes it.
The report is here (with log attached): https://bugs.gentoo.org/899742

As stated in the bug I cannot recreate this at all when building myself - I get no warnings from ar creating the libxdp archive, using the same toolchain options, binutils, elfutils etc.
Any idea what might be responsible for the ar warnings from the check_abi target?

tohojo:
Holger Hoffstätte ***@***.***> writes:

> After getting 1.3.1 building properly (thanks!) our CI promptly found .. "something", but we're not really sure what it is.
> The report is here (with log attached): https://bugs.gentoo.org/899742
>
> As stated in the bug I cannot recreate this at all when building myself - I get no warnings from ar creating the libxdp archive, using the same toolchain options, binutils, elfutils etc.
> Any idea what might be responsible for the ar warnings from the check_abi target?

The libxdp Makefile contains this check. My guess would be that maybe
the CI system is missing one of the utilities used to compute those
numbers, most likely 'readelf'? They're defined as:

GLOBAL_SYM_COUNT = $(shell readelf -s --wide $(XDP_IN_SHARED) | \
			   cut -d "@" -f1 | sed 's/_v[0-9]_[0-9]_[0-9].*//' | \
			   sed 's/\[.*\]//' | \
			   awk '/GLOBAL/ && /DEFAULT/ && !/UND/ {print $$NF}' | \
			   sort -u | wc -l)
VERSIONED_SYM_COUNT = $(shell readelf --dyn-syms --wide $(OBJDIR)/libxdp.so | \
			      grep -Eo '[^ ***@***.***_' | cut -d@ -f1 | sort -u | wc -l)

hhoffstaette:
Thanks for the idea, but a missing readelf seems unlikely since it's part of binutils and the CI has the same version as I do (and I can build & use everything just fine). So it seems there's definitely something off in the CI - we'll keep looking.

hhoffstaette:
[...]

PR #341, opened 2023-07-08
libxdp: don't use direct call to readelf for symbol table comparison

hhoffstaette:
Gentoo CI previously reported symbol mismatches in libxdp (see #302 and https://bugs.gentoo.org/899742) when multiple toolchains are installed. This was caused by a direct call to 'readelf' instead of using the command injected by the environment/toolchain.
The fix is simple: use $(READELF).


tohojo:
Thank you for the fix! :)

Issue #343, opened 2023-07-14
Potential fd leak in xdp_program__is_attached


AdrianJab:
Hello.
We have a watchdog class (using libxdp), which periodically calls *xdp_program__is_attached* in order to check if our XDP programs are still attached on sockets. The watchdog fires every second, and every second it creates two new files, which we can track with `lsof` command.

Example:
```
data_engi 336030 root    8u  a_inode      0,14        0     12569 [eventpoll]
data_engi 336030 root    9u     sock       0,8      0t0 166896037 protocol: XDP
data_engi 336030 root   10u     sock       0,8      0t0 166896052 protocol: XDP
data_engi 336030 root   11u  a_inode      0,14        0     12569 [eventfd]
data_engi 336030 root   12r      CHR     246,3      0t0       115 /dev/ptp3
data_engi 336030 root   13u  a_inode      0,14        0     12569 [eventfd]
data_engi 336030 root   14u  a_inode      0,14        0     12569 [eventfd]
data_engi 336030 root   15r  a_inode      0,14        0     12569 btf
data_engi 336030 root   16u  a_inode      0,14        0     12569 bpf-map
data_engi 336030 root   17u  a_inode      0,14        0     12569 bpf-prog
data_engi 336030 root   18u  a_inode      0,14        0     12569 bpf-prog
data_engi 336030 root   19u  a_inode      0,14        0     12569 [eventfd]
data_engi 336030 root   20r  a_inode      0,14        0     12569 btf
data_engi 336030 root   21u  a_inode      0,14        0     12569 bpf-map
data_engi 336030 root   22u  a_inode      0,14        0     12569 bpf-prog
data_engi 336030 root   23u  a_inode      0,14        0     12569 bpf-prog
data_engi 336030 root   24u  a_inode      0,14        0     12569 [eventfd]
data_engi 336030 root   25u  a_inode      0,14        0     12569 [eventfd]
data_engi 336030 root   26u  a_inode      0,14        0     12569 bpf-prog
data_engi 336030 root   27u  a_inode      0,14        0     12569 bpf-prog
data_engi 336030 root   28u  a_inode      0,14        0     12569 bpf-prog
data_engi 336030 root   29u  a_inode      0,14        0     12569 bpf-prog
data_engi 336030 root   30u  a_inode      0,14        0     12569 bpf-prog
data_engi 336030 root   31u  a_inode      0,14        0     12569 bpf-prog
[...]

PR #345, opened 2023-07-18
Fix bpf program original file  descriptor management

marcos44pl:
All fd that are taken from bpf shall be closed right after they were cloned:
`fd = bpf_prog_get_fd_by_id(id);`

### DIFF
--- a/lib/libxdp/libxdp.c
+++ b/lib/libxdp/libxdp.c
@@ -1433,9 +1433,10 @@ struct xdp_program *xdp_program__from_id(__u32 id)
 	}
 
 	prog = xdp_program__from_fd(fd);
+	// duplicated fd already in prog, close original
+	close(fd);
 	if (IS_ERR(prog)) {
 		err = errno;
-		close(fd);
 		errno = err;
 	}
 	return prog;
@@ -1455,9 +1456,10 @@ struct xdp_program *xdp_program__from_pin(const char *pin_path)
 	}
 
 	prog = xdp_program__from_fd(fd);
+	// duplicated fd already in prog, close original
+	close(fd);
 	if (IS_ERR(prog)) {
 		err = errno;
-		close(fd);
 		errno = err;
 	}
 	return prog;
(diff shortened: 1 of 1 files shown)


marcos44pl:
PR is related to the issue: #343

marcos44pl:
I tested it with valgrind and seems to be working and fixing the problem mentioned in the issue

lib/libxdp/libxdp.c
```diff
@@ -1433,9 +1433,11 @@ struct xdp_program *xdp_program__from_id(__u32 id)
 	}
 
 	prog = xdp_program__from_fd(fd);
+	// duplicated fd already in prog, close original
+	if(fd > 0)
+		close(fd);
```
tohojo, line 1438: There's a check right above this for the validity of fd, so the if statement here is redundant. The same goes for all the other places you're doing changing below :)

tohojo (changes requested):
Thank you for the fix! The general change LGTM, but please get rid of all the redundant checks of the fds :)

lib/libxdp/libxdp.c
marcos44pl: Yeah, you're right, I went too far, but thanks to that I realized one problem with fd/hw_fd handling

lib/libxdp/libxdp.c
```diff
@@ -2491,6 +2493,9 @@ static struct xdp_multiprog *xdp_multiprog__from_id(__u32 id, __u32 hw_id,
 		err = PTR_ERR(mp);
 		goto err;
 	}
+	// duplicated fd/hw_fd already in prog, close originals
+	close(fd);
+	close(hw_fd);
```
tohojo, line 2498: Hmm, actually, looking at the code again here, this function is not guaranteed to be called with both id and hw_id set, so here we *do* need a check for `fd > 0` before closing. Sorry for not realising that the first time around :(

tohojo (approved):
Alright, LGTM; thanks for the fix! :)

Issue #344, opened 2023-07-17
"Unable" to read output through pipes - large buffer size


enhaut:
Default large pipe buffer (64KiB on linux) causes issues with reading `xdp-bench` output using pipes.

I am actually trying to read `xdp-bench` output for further analysis, i am simply piping `stdout` to `stdin` of my application, however due to large pipe buffer `stdout` seems to be empty.

There are multiple ways to resolve this issue:

* set size of `stdout` (and eventually `stderr`) buffer to `0` so prints will be immediately flushed/printed to stdout
* flushing `stdout` buffer using `fflush(stdout)` once a while e.g. right after summary/sample line is printed: https://github.com/xdp-project/xdp-tools/blob/00ff5bf76f28f518acbd40a7aed7606ff21dc364/lib/util/xdp_sample.c#L1119
* setting system-wide pipe buffer size. However, I think that's pretty clumsy 

I've got a first option implemented and tested in my fork, see commit https://github.com/xdp-project/xdp-tools/commit/c608351b30544abf08c183a2ea85a6c135833692

tohojo:
Hmm, fair point. I think I would prefer to add an explicit flush after newline; care to open a pull request to do that? :)

PR #351, opened 2023-08-08
xdp-bench: flushing outputs after writing stats

enhaut:
Having buffering enabled (by default in linux) for `stdout` causes issues for pipes. Since pipe buffer is pretty large (64KiB) it would take ages to fill that up. Therefore, applications in pipe are "unable" to reliably read `xdp-bench`'s `stdout`.

This approach may affect performance, but since `xdp-bench` tool does not rely on `stdout`/`stderr` operations, it's safe to flush both `stdout` and `stderr` buffers immediately.

#### Tests

before:
```python
>>> import time
>>> import subprocess as s
>>> p = s.Popen("xdp-bench drop eno12399", stdout=s.PIPE, stderr=s.PIPE, shell=True)  # piped back to python process
>>> time.sleep(15)
>>> p.kill()
>>> p.communicate()
(b'', b'Dropping packets on eno12399 (ifindex 4; driver ice)\n')  # (stdout, stderr)
>>>
```

after:  
```python
>>> import subprocess as s
>>> import time
>>> p = s.Popen("xdp-bench drop eno12399", stdout=s.PIPE, stderr=s.PIPE, shell=True)  # piped back to python process
>>> time.sleep(15)
>>> p.kill()
>>> p.communicate()
(b'Summary                         0 rx/s                  0 err/s        \nSummary                         0 rx/s                  0 err/s        \nSummary                         0 rx/s                  0 err/s        \nSummary                         0 rx/s                  0 err/s        \nSummary                         0 rx/s                  0 err/s        \nSummary                         0 rx/s                  0 err/s        \nSummary                         0 rx/s                  0 err/s        \nSummary                         0 rx/s                  0 err/s        \nSummary                         0 rx/s                  0 err/s        \nSummary                         0 rx/s                  0 err/s        \nSummary                         0 rx/s                  0 err/s        \nSummary                         0 rx/s                  0 err/s        \n', b'Dropping packets on eno12399 (ifindex 4; driver ice)\n')  # (stdout, stderr)
>>>
```

Closes https://github.com/xdp-project/xdp-tools/issues/344

### DIFF
--- a/lib/util/xdp_sample.c
+++ b/lib/util/xdp_sample.c
@@ -1192,6 +1192,10 @@ static void stats_print(const char *prefix, int mask, struct stats_record *r,
 		sample_err_exp = false;
 		printf("\n");
 	}
+
+	fflush(stdout);
+	fflush(stderr);
+	// Flushing both outputs to "bypass" buffering
 }
 
 static int get_num_rxqs(const char *ifname)


tohojo:
So I don't mind merging this, but I'll also point out that your example application is buggy: calling `p.kill()` sends a SIGKILL to the process, which means it won't have a chance to do any cleanup. Such cleanup includes flushing output buffers, but for xdp-bench it also includes detaching the XDP program loaded onto the interface. So definitely don't do that in a production setup, use `p.terminate()` instead (and only resort to `kill()` if the application doesn't exit within a second or two).

Substituting `p.terminate()` in your first example above means the output is displayed just fine, BTW. However, flushing for every line can still be useful for cases where you want to read the lines immediately as they appear, so doing that anyway is fine by me :)

enhaut:
Yep, agree with terminating apps gracefully :) .  
Attached tests are just quick demonstration of the problem and fix, `.terminate()` would mask out the fact that buffers are not flushed in the mean time. Which is crucial for many apps :)

PR #358, opened 2023-09-13
Fix compilation error for arm64

sachintiptur:
This PR adds ARCH specific include files to CFLAGS and BPF_FLAGS.

### DIFF
--- a/configure
+++ b/configure
@@ -67,6 +67,7 @@ check_toolchain()
     : ${EMACS=emacs}
     : ${BPFTOOL=bpftool}
     : ${ARCH_INCLUDES=}
+    : ${ARCH_NAME=}
 
     CLANG=$(find_tool clang "$CLANG")
     LLC=$(find_tool llc "$LLC")
@@ -77,6 +78,8 @@ check_toolchain()
             exit 1;
         fi;
     done
+
+    ARCH_NAME=$($CC -print-multiarch 2>/dev/null)
 
     clang_version=$($CLANG --version | grep -Po '(?<=clang version )[[:digit:]]+')
     if [ "$?" -ne "0" ]; then
@@ -116,10 +119,10 @@ check_toolchain()
         exit 1
     fi
 
-    if [ -z "$ARCH_INCLUDES" ]; then
+    if [ -z "$ARCH_INCLUDES" ] && [ -n "$ARCH_NAME" ]; then
         for dir in $(echo | $CC -Wp,-v -E - 2>&1 | grep '^ '); do
             local idir
-            idir="${dir}/$(uname -m)-linux-gnu/"
+            idir="${dir}/${ARCH_NAME}/"
             [ -d "$idir" ] && ARCH_INCLUDES="-I${idir} $ARCH_INCLUDES"
         done
     fi
--- a/lib/defines.mk
+++ b/lib/defines.mk
@@ -40,8 +40,8 @@ endif
 
 DEFINES += -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64
 
-CFLAGS += -std=gnu11 -Wextra -Werror $(DEFINES)
-BPF_CFLAGS += $(DEFINES) $(filter -ffile-prefix-map=%,$(CFLAGS))
+CFLAGS += -std=gnu11 -Wextra -Werror $(DEFINES) $(ARCH_INCLUDES)
+BPF_CFLAGS += $(DEFINES) $(filter -ffile-prefix-map=%,$(CFLAGS)) $(ARCH_INCLUDES)
 
 CONFIGMK := $(LIB_DIR)/../config.mk
 LIBMK := Makefile $(CONFIGMK) $(LIB_DIR)/defines.mk $(LIB_DIR)/common.mk $(LIB_DIR)/../version.mk


configure
```diff
@@ -119,7 +119,7 @@ check_toolchain()
     if [ -z "$ARCH_INCLUDES" ]; then
         for dir in $(echo | $CC -Wp,-v -E - 2>&1 | grep '^ '); do
             local idir
-            idir="${dir}/$(uname -m)-linux-gnu/"
+            idir="${dir}/$($CC -print-multiarch)/"
```
tohojo, line 122: There should probably be a separate test for whether `$CC -print-multiarch` works at all. If `$CC` is set to clang, this command will fail, but then the dir check will succeed because the path expands to the parent dir and we'll end up duplicating all include dirs into `$ARCH_INCLUDES`...

sachintiptur: In my system(ARM based), it works for both gcc and clang.
```
linux:~$ gcc -print-multiarch
aarch64-linux-gnu
linux:~$ clang -print-multiarch
aarch64-linux-gnu
```

tohojo:
Sachin Tiptur ***@***.***> writes:

> @sachintiptur commented on this pull request.
>
>
>
>> @@ -119,7 +119,7 @@ check_toolchain()
>      if [ -z "$ARCH_INCLUDES" ]; then
>          for dir in $(echo | $CC -Wp,-v -E - 2>&1 | grep '^ '); do
>              local idir
> -            idir="${dir}/$(uname -m)-linux-gnu/"
> +            idir="${dir}/$($CC -print-multiarch)/"
>
> In my system(ARM based), it works for both gcc and clang.
> ```
> linux:~$ gcc -print-multiarch
> aarch64-linux-gnu
> linux:~$ clang -print-multiarch
> aarch64-linux-gnu
> ```

Hmm, with which Clang version? It doesn't on mine (x86_64):

% clang -print-multiarch
clang-16: error: unsupported option '-print-multiarch'
clang-16: error: no input files
% gcc -print-multiarch


(gcc just prints an empty line - which would also lead to the
duplication I mentioned, BTW, so we'll definitely need to check if the
output is non-empty)

sachintiptur:
> Sachin Tiptur ***@***.***> writes:
> @sachintiptur commented on this pull request. > @@ -119,7 +119,7 @@ check_toolchain() if [ -z "$ARCH_INCLUDES" ]; then for dir in $(echo | $CC -Wp,-v -E - 2>&1 | grep '^ '); do local idir - idir="${dir}/$(uname -m)-linux-gnu/" + idir="${dir}/$($CC -print-multiarch)/" In my system(ARM based), it works for both gcc and clang. ``` linux:~$ gcc -print-multiarch aarch64-linux-gnu linux:~$ clang -print-multiarch aarch64-linux-gnu ```
> Hmm, with which Clang version? It doesn't on mine (x86_64): % clang -print-multiarch clang-16: error: unsupported option '-print-multiarch' clang-16: error: no input files % gcc -print-multiarch (gcc just prints an empty line - which would also lead to the duplication I mentioned, BTW, so we'll definitely need to check if the output is non-empty)

yeah sure, I will add a check.
```
linux:~$ clang --version
Ubuntu clang version 15.0.7
Target: aarch64-unknown-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```

sachintiptur:
> Sachin Tiptur ***@***.***> writes:
> @sachintiptur commented on this pull request. > @@ -119,7 +119,7 @@ check_toolchain() if [ -z "$ARCH_INCLUDES" ]; then for dir in $(echo | $CC -Wp,-v -E - 2>&1 | grep '^ '); do local idir - idir="${dir}/$(uname -m)-linux-gnu/" + idir="${dir}/$($CC -print-multiarch)/" In my system(ARM based), it works for both gcc and clang. ``` linux:~$ gcc -print-multiarch aarch64-linux-gnu linux:~$ clang -print-multiarch aarch64-linux-gnu ```
> Hmm, with which Clang version? It doesn't on mine (x86_64): % clang -print-multiarch clang-16: error: unsupported option '-print-multiarch' clang-16: error: no input files % gcc -print-multiarch (gcc just prints an empty line - which would also lead to the duplication I mentioned, BTW, so we'll definitely need to check if the output is non-empty)

clang-15.0.7
yeah sure, will add a check

configure
```diff
@@ -67,6 +67,7 @@ check_toolchain()
     : ${EMACS=emacs}
     : ${BPFTOOL=bpftool}
     : ${ARCH_INCLUDES=}
+    : ${ARCH_NAME=$($CC -print-multiarch)}
```
tohojo, line 70: This needs to silence stderr, otherwise we get output like this:

```
$ CC=clang ./configure
clang-16: error: unsupported option '-print-multiarch'
clang-16: error: no input files
Found clang binary 'clang' with version 16 (from 'clang version 16.0.6')
using emacs: GNU Emacs 29.1
using bpftool v7.2.0
libbpf support: system v1.2.2
  perf_buffer__consume support: yes
  btf__load_from_kernel_by_id support: yes
  btf__type_cnt support: yes
  bpf_object__next_map support: yes
  bpf_object__next_program support: yes
  bpf_program__insn_cnt support: yes
  bpf_program__type support: yes
  bpf_program__flags support: yes
  bpf_program__expected_attach_type support: yes
  bpf_map_create support: yes
  perf_buffer__new_raw support: yes
  bpf_xdp_attach support: yes
  bpf_map__set_autocreate support: yes
  bpf_prog_test_run_opts support: yes
  bpf_xdp_query support: yes
secure_getenv support: yes
cap-ng support: yes
```

Also, probably better to put this below the check for whether `$CC` actually exists, below...

tohojo (approved):
Alright, LGTM. Could you please squash the commits into a single one and write a few lines of commit message explaining that using `uname -m` is the wrong way to get the architecture, but that `gcc -print-multiarch` does the right thing?

tohojo:
Merged as 6ce3063 (but Github didn't seem to notice)

PR #362, opened 2023-09-20
Support detaching xdp_link-based attachments fix #361

### LINKED ISSUE #361: Support force-detaching xdp_link-based attachments
We currently don't handle bpf_link-based XDP attachments at all; we should
support force-detaching such attachments, not just the netlink-based ones.

msherif1234:
supporting force detach for link based attachments using `xdp-loader unload -a <intf>` fixes #361

### DIFF
--- a/lib/libxdp/libxdp.c
+++ b/lib/libxdp/libxdp.c
@@ -3223,6 +3223,51 @@ out:
 	return err;
 }
 
+static int xdp_detach_link(__u32 ifindex, __u32 prog_id) {
+	struct bpf_link_info link_info;
+	__u32 link_info_len, id = 0;
+	int err, fd;
+
+	while (true) {
+		err = bpf_link_get_next_id(id, &id);
+		if (err) {
+			err = -errno;
+			pr_debug("Can't get next link for id %u: %s", id, strerror(errno));
+			return err;
+		}
+
+		fd = bpf_link_get_fd_by_id(id);
+		if (fd < 0) {
+			err = -errno;
+			pr_debug("Can't get link by id %u: %s", id, strerror(errno));
+			return err;
+		}
+
+		memset(&link_info, 0, sizeof(link_info));
+		link_info_len = sizeof(link_info);
+
+		err = bpf_obj_get_info_by_fd(fd, &link_info, &link_info_len);
+		if (err) {
+			err = -errno;
+			pr_debug("Can't get link info for %u: %s", id, strerror(errno));
+			break;
+		}
+
+		if (link_info.type == BPF_LINK_TYPE_XDP && link_info.xdp.ifindex == ifindex && link_info.prog_id == prog_id) {
+			pr_debug("Detach link for id %u for prog %u on interface %u", id, prog_id, ifindex);
+			err = bpf_link_detach(fd);
+			if (err) {
+				err = -errno;
+				pr_warn("Can't detach link %u: %s", id, strerror(errno));
+			}
+			break;
+		}
+		close(fd);
+	}
+	close(fd);
+	return err;
+}
+
 static int xdp_multiprog__attach(struct xdp_multiprog *old_mp,
 				 struct xdp_multiprog *mp,
 				 enum xdp_attach_mode mode)
@@ -3253,8 +3298,13 @@ static int xdp_multiprog__attach(struct xdp_multiprog *old_mp,
 
 
 	err = xdp_attach_fd(prog_fd, old_fd, ifindex, mode);
-	if (err < 0)
+	if (err < 0) {
+		if (errno == EBUSY && !mp && mode == XDP_MODE_NATIVE) {
+			pr_debug("Detaching link on ifindex %d\n", ifindex);
+			return xdp_detach_link(ifindex, xdp_multiprog__main_id(old_mp));
+		}
 		goto err;
+	}
 
 	if (mp)
 		pr_debug("Loaded %zu programs on ifindex %d%s\n",
--- a/lib/libxdp/tests/test-libxdp.sh
+++ b/lib/libxdp/tests/test-libxdp.sh
@@ -1,6 +1,6 @@
 # SPDX-License-Identifier: (GPL-2.0 OR BSD-2-Clause)
 
-ALL_TESTS="test_link_so test_link_a test_old_dispatcher test_xdp_frags test_xsk_prog_refcnt_bpffs test_xsk_prog_refcnt_legacy test_xsk_non_privileged"
+ALL_TESTS="test_link_so test_link_a test_old_dispatcher test_xdp_frags test_xsk_prog_refcnt_bpffs test_xsk_prog_refcnt_legacy test_xsk_non_privileged test_link_detach"
 
 TESTS_DIR=$(dirname "${BASH_SOURCE[0]}")
 
@@ -102,9 +102,20 @@ test_xsk_non_privileged()
 	ip link delete xdp_veth0
 }
 
+test_link_detach()
+{
+        if test ! -f $TEST_PROG_DIR/test_link_detach; then
+		exit "$SKIPPED_TEST"
+	fi
+	ip link add xdp_veth0 type veth peer name xdp_veth1
+	check_run $TESTS_DIR/test_link_detach xdp_veth0
+	ip link delete xdp_veth0
+}
+
 cleanup_tests()
 {
     ip link del dev xdp_veth_big0 >/dev/null 2>&1
     ip link del dev xdp_veth_small0 >/dev/null 2>&1
     ip link del dev xsk_veth0 >/dev/null 2>&1
+    ip link del dev xdp_veth0 >/dev/null 2>&1
 }
--- a/lib/libxdp/tests/.gitignore
+++ b/lib/libxdp/tests/.gitignore
@@ -3,3 +3,4 @@ check_kern_compat
 test_xdp_frags
 test_dispatcher_versions
 test_xsk_non_privileged
+test_link_detach
--- a/lib/libxdp/tests/Makefile
+++ b/lib/libxdp/tests/Makefile
@@ -1,6 +1,6 @@
 # SPDX-License-Identifier: (GPL-2.0 OR BSD-2-Clause)
 
-USER_TARGETS := test_xsk_refcnt check_kern_compat test_xdp_frags test_dispatcher_versions
+USER_TARGETS := test_xsk_refcnt check_kern_compat test_xdp_frags test_dispatcher_versions test_link_detach
 BPF_TARGETS := xdp_dispatcher_v1 xdp_pass
 USER_LIBS := -lpthread
 
--- a/lib/libxdp/tests/test_link_detach.c
+++ b/lib/libxdp/tests/test_link_detach.c
@@ -1,0 +1,112 @@
+ /* SPDX-License-Identifier: GPL-2.0 */
+
+#define _GNU_SOURCE
+
+#include <errno.h>
+#include <linux/err.h>
+#include <net/if.h>
+#include <pthread.h>
+#include <stdbool.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/resource.h>
+#include <unistd.h>
+
+#include "test_utils.h"
+
+#include <xdp/libxdp.h>
+#include <bpf/libbpf.h>
+
+#define SKIPPED_TEST 249 // needs to match SKIPPED_TEST value in test_runner.sh
+
+static void usage(char *progname)
+{
+	fprintf(stderr, "Usage: %s <ifname>\n", progname);
+	exit(EXIT_FAILURE);
+}
+
+static int check_link_detach(int ifindex) {
+	struct bpf_object *obj_prog = NULL;
+	struct bpf_program *prog;
+	struct xdp_multiprog *mp = NULL;
+	struct bpf_link *link = NULL;
+	int ret;
+
+	if (!ifindex)
+		return -EINVAL;
+
+	obj_prog = bpf_object__open("xdp_pass.o");
+	if (!obj_prog) {
+		ret = -errno;
+		goto out;
+	}
+
+	prog = bpf_object__find_program_by_name(obj_prog, "xdp_pass");
+	if (!prog) {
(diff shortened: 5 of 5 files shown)


lib/libxdp/libxdp.c
```diff
@@ -3223,6 +3223,43 @@ static int xdp_multiprog__unpin(struct xdp_multiprog *mp)
 	return err;
 }
 
+static void xdp_detach_links(uint32_t ifindex) {
+	struct bpf_link_info link_info;
+	u_int32_t link_info_len = sizeof(link_info), id;
+	int err, fd;
+
+	while (true) {
+		err = bpf_link_get_next_id(id, &id);
```
tohojo, line 3232: The first time through the loop, `id` will be passed in uninitialised here.

lib/libxdp/libxdp.c
```diff
@@ -3223,6 +3223,43 @@ static int xdp_multiprog__unpin(struct xdp_multiprog *mp)
...
+			continue;
+		}
+		if (link_info.type == BPF_LINK_TYPE_XDP && link_info.xdp.ifindex == ifindex) {
+			err = bpf_link_detach(fd);
+			if (err) {
+				pr_debug("can't detach link %u: %s", id, strerror(errno));
+			}
+			close(fd);
+			break;
+		}
```
tohojo, line 3259: If we get to the end of the loop without detaching, we end up not closing `fd`.

lib/libxdp/libxdp.c
```diff
@@ -3223,6 +3223,43 @@ static int xdp_multiprog__unpin(struct xdp_multiprog *mp)
...
+			continue;
+		}
+		memset(&link_info, 0, sizeof(link_info));
+		err = bpf_obj_get_info_by_fd(fd, &link_info, &link_info_len);
+		if (err) {
+			pr_debug("can't get link info for %u: %s", id, strerror(errno));
+			close(fd);
+			continue;
+		}
+		if (link_info.type == BPF_LINK_TYPE_XDP && link_info.xdp.ifindex == ifindex) {
```
tohojo, line 3252: Check against prog ID here.

lib/libxdp/libxdp.c
```diff
@@ -3223,6 +3223,43 @@ static int xdp_multiprog__unpin(struct xdp_multiprog *mp)
...
+			err = bpf_link_detach(fd);
+			if (err) {
+				pr_debug("can't detach link %u: %s", id, strerror(errno));
```
tohojo, line 3255: At this point we actually found a link to detach, so if it fails we should be printing at a higher severity than debug (`pr_warn()` would be suitable I think). A `pr_debug()` would be useful either *before* attempting the detach, or on success, though...

lib/libxdp/libxdp.c
```diff
@@ -3223,6 +3223,43 @@ static int xdp_multiprog__unpin(struct xdp_multiprog *mp)
...
+			if (errno == ENOENT)
+				break;
+			pr_debug("can't get next link: %s (id %d)", strerror(errno), id);
+			break;
+		}
+
+		fd = bpf_link_get_fd_by_id(id);
+		if (fd < 0) {
+			pr_debug("can't get link by id (%u): %s", id, strerror(errno));
+			continue;
```
tohojo, line 3243: Not sure we should just be continuing here; if we can't get the link fd for one link, we probably can't get it for the next one either, so better to just bail out (with the same `err = -errno` assignment as above).

lib/libxdp/libxdp.c
```diff
@@ -3223,6 +3223,43 @@ static int xdp_multiprog__unpin(struct xdp_multiprog *mp)
...
+		}
+		memset(&link_info, 0, sizeof(link_info));
+		err = bpf_obj_get_info_by_fd(fd, &link_info, &link_info_len);
+		if (err) {
+			pr_debug("can't get link info for %u: %s", id, strerror(errno));
+			close(fd);
+			continue;
```
tohojo, line 3250: Same here - break instead of continuing.

lib/libxdp/libxdp.c
```diff
@@ -3253,8 +3290,15 @@ static int xdp_multiprog__attach(struct xdp_multiprog *old_mp,
 
 
 	err = xdp_attach_fd(prog_fd, old_fd, ifindex, mode);
-	if (err < 0)
+	if (err < 0) {
+		if (!mp) {
```
tohojo, line 3294: We should only attempt the link detachment if the kernel returns EBUSY, otherwise we're just doing useless work.

tohojo (changes requested):
The basic functionality is basically there, nice job! Just a whole bunch of nits :)

Also, it would be good to add a selftest for this; could you please write add a small test utility in libxdp/tests that does something like:

```c
link = bpf_link_attach(ifindex);
mp = xdp_multiprog__get_from_ifindex(ifindex);
xdp_multiprog__detach(mp);
bpf_link_close(link);
```

You can look at some of the other examples in the tests/ directory for how to execute the tests from the test script :)

lib/libxdp/libxdp.c
msherif1234: done

lib/libxdp/libxdp.c
msherif1234: done

lib/libxdp/libxdp.c
msherif1234: done

lib/libxdp/libxdp.c
msherif1234: changed to warn and added debug before attempting to detach

lib/libxdp/libxdp.c
msherif1234: done

lib/libxdp/libxdp.c
msherif1234: fixed

lib/libxdp/libxdp.c
msherif1234: fixed it

lib/libxdp/libxdp.c
```diff
@@ -3223,6 +3223,43 @@ static int xdp_multiprog__unpin(struct xdp_multiprog *mp)
...
 }
 
+static void xdp_detach_links(uint32_t ifindex) {
+	struct bpf_link_info link_info;
+	u_int32_t link_info_len = sizeof(link_info), id;
+	int err, fd;
+
+	while (true) {
+		err = bpf_link_get_next_id(id, &id);
+		if (err) {
```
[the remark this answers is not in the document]
msherif1234: done

lib/libxdp/libxdp.c
```diff
@@ -3223,6 +3223,43 @@ static int xdp_multiprog__unpin(struct xdp_multiprog *mp)
 	return err;
 }
 
+static void xdp_detach_links(uint32_t ifindex) {
+	struct bpf_link_info link_info;
+	u_int32_t link_info_len = sizeof(link_info), id;
```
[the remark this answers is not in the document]
msherif1234: fixed

lib/libxdp/libxdp.c
```diff
@@ -3223,6 +3223,43 @@ static int xdp_multiprog__unpin(struct xdp_multiprog *mp)
 	return err;
 }
 
+static void xdp_detach_links(uint32_t ifindex) {
```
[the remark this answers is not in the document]
msherif1234: added prog id

lib/libxdp/libxdp.c
```diff
@@ -3253,8 +3290,15 @@ static int xdp_multiprog__attach(struct xdp_multiprog *old_mp,
 
 
 	err = xdp_attach_fd(prog_fd, old_fd, ifindex, mode);
-	if (err < 0)
+	if (err < 0) {
+		if (!mp) {
+			pr_debug("Detaching links on ifindex %d%s\n", ifindex,
+			mode == XDP_MODE_SKB? " in skb mode":"");
+			xdp_detach_links(ifindex);
+			return 0;
```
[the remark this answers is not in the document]
msherif1234: done

lib/libxdp/libxdp.c
```diff
@@ -3223,6 +3223,50 @@ static int xdp_multiprog__unpin(struct xdp_multiprog *mp)
...
+			pr_debug("detach link for %u for prog %u on interface %u", id, prog_id, ifindex);
+			err = bpf_link_detach(fd);
+			if (err) {
+				err = -errno;
+				pr_warn("can't detach link %u: %s", id, strerror(errno));
+			}
+			break;
+		}
+	}
+	close(fd);
```
tohojo, line 3266: Still need a `close(fd)` at the end of the loop :)

lib/libxdp/libxdp.c
```diff
@@ -3253,8 +3297,14 @@ static int xdp_multiprog__attach(struct xdp_multiprog *old_mp,
 
 
 	err = xdp_attach_fd(prog_fd, old_fd, ifindex, mode);
-	if (err < 0)
+	if (err < 0) {
+		if (errno == EBUSY && !mp) {
+			pr_debug("Detaching link on ifindex %d%s\n", ifindex,
+			mode == XDP_MODE_SKB? " in skb mode":"");
```
tohojo, line 3303: `bpf_link` is native-mode only, so instead of this we shouldn't attempt this at all if `mode == XDP_MODE_SKB`

lib/libxdp/libxdp.c
```diff
@@ -3223,6 +3223,50 @@ static int xdp_multiprog__unpin(struct xdp_multiprog *mp)
...
+static int xdp_detach_link(__u32 ifindex, __u32 prog_id) {
+	struct bpf_link_info link_info;
+	__u32 link_info_len, id = 0;
+	int err, fd;
+
+	while (true) {
+		err = bpf_link_get_next_id(id, &id);
+		if (err) {
+			err = -errno;
+			pr_debug("can't get next link: %s (id %d)", strerror(errno), id);
```
tohojo, line 3235: Lets make the order of id and error message consistent across the different debug prints. We also generally capitalise sentences in debug output throughout libxdp...

tohojo (changes requested):
A few more nits, but we're getting there :)

lib/libxdp/tests/test_link_dispatch.c
```diff
@@ -0,0 +1,111 @@
...
+
+
+    ret = bpf_object__load(obj_prog);
+    if (ret) {
+        ret = -errno;
+        fprintf(stderr, "Couldn't load object: %s\n", strerror(-ret));
+        goto out;
+    }
+
+    link = bpf_program__attach_xdp(prog, ifindex);
```
tohojo, line 58: This needs a `bpf_link__destroy()` on exit.

lib/libxdp/tests/test_link_dispatch.c
```diff
@@ -0,0 +1,111 @@
...
+    if (!link) {
+        ret = -errno;
+        fprintf(stderr, "Couldn't attach XDP prog to ifindex %d: %s\n", ifindex, strerror(-ret));
+        goto out;
+    }
+
+
+    mp = xdp_multiprog__get_from_ifindex(ifindex);
```
tohojo, line 66: And this needs an `xdp_multiprog__close()`

lib/libxdp/tests/test-libxdp.sh
```diff
@@ -102,6 +102,16 @@ test_xsk_non_privileged()
...
 }
 
+test_link_dispatch()
+{
+        if test ! -f $TEST_PROG_DIR/test_link_dispatch; then
+		exit "$SKIPPED_TEST"
+	fi
+	ip link add xdp_veth0 type veth peer name xdp_veth1
+	check_run $TESTS_DIR/test_link_dispatch xdp_veth0
+	ip link delete xdp_veth0
```
tohojo, line 112: This should be added to `cleanup_tests()` below as well

lib/libxdp/tests/test_link_dispatch.c
```diff
@@ -0,0 +1,111 @@
...
+#include <bpf/libbpf.h>
+
+
+static void usage(char *progname)
+{
+	fprintf(stderr, "Usage: %s <ifname>\n", progname);
+	exit(EXIT_FAILURE);
+}
+
+static int check_link_detach(int ifindex) {
```
tohojo, line 27: The indentation is off in this and the `main()` function below. We use a single tab character per indentation level everywhere.

lib/libxdp/tests/test_link_dispatch.c
```diff
@@ -0,0 +1,111 @@
...
+    struct bpf_object *obj_prog = NULL;
+    struct bpf_program *prog;
+    struct xdp_multiprog *mp;
+    struct bpf_link *link;
+    int ret;
+
+    if (!ifindex) {
+        return -EINVAL;
+    }
```
tohojo, line 36: Nit: Don't need the braces here.

lib/libxdp/tests/test_link_dispatch.c
```diff
@@ -0,0 +1,111 @@
...
+    ret = bpf_object__load(obj_prog);
+    if (ret) {
+        ret = -errno;
+        fprintf(stderr, "Couldn't load object: %s\n", strerror(-ret));
+        goto out;
+    }
+
+    link = bpf_program__attach_xdp(prog, ifindex);
+    if (!link) {
+        ret = -errno;
```
tohojo, line 60: Instead of returning the errno here, I think we should exit with the special return code that makes the test framework mark the test as skipped (249 - add a define for it, though). That way the CI won't fail on this test on older kernels (which I suspect it does now).

We could have a separate detection step first, but that would basically end up doing the same that this function does up to this point, so I think it's easier to just have a direct return of SKIPPED_TEST here :)

tohojo (changes requested):
Only a few style nits and some missing cleanup, otherwise the test LGTM.

lib/libxdp/tests/test-libxdp.sh
msherif1234: test cases b4 this one deleting at the end as well ?

tohojo (approved):

PR #374, opened 2023-10-26
xdp-bench: Add l4-sport and l4-dport cpumap modes

danobi:
These modes assigns CPUs based on L4 source and dest port. This is useful for load testing scenarios where you can assign sequential source ports to flows (like with neper).

For router/gateway use cases you use sport mapping on the forward direction and dport mapping on the reverse direction.

### DIFF
--- a/xdp-bench/xdp_redirect_cpumap.bpf.c
+++ b/xdp-bench/xdp_redirect_cpumap.bpf.c
@@ -96,7 +96,7 @@ bool parse_eth(struct ethhdr *eth, void *data_end,
 }
 
 static __always_inline
-__u16 get_dest_port_ipv4_udp(struct xdp_md *ctx, __u64 nh_off)
+__u16 get_port_ipv4_udp(struct xdp_md *ctx, __u64 nh_off, bool src)
 {
 	void *data_end = (void *)(long)ctx->data_end;
 	void *data     = (void *)(long)ctx->data;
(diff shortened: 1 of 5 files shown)


danobi:
FWIW, this was used as part of the "Multi-core IPsec tunnels" work that I will be presenting at netdev next week.

xdp-bench/xdp_redirect_cpumap.bpf.c
```diff
@@ -491,6 +564,94 @@ int  cpumap_l4_hash(struct xdp_md *ctx)
...
+	__u32 key = bpf_get_smp_processor_id();
+	struct ethhdr *eth = data;
+	__u8 ip_proto = IPPROTO_UDP;
+	struct datarec *rec;
+	__u16 eth_proto = 0;
+	__u64 l3_offset = 0;
+	__u32 cpu_dest = 0;
+	__u32 *cpu_lookup;
+	__u32 cpu_idx = 0;
+	__u16 src_port;
```
tohojo, line 581: Could we name this differently? Since it's either a source or destination port, depending on the 'src' parameter? I suppose just 'port' would be fine :)

tohojo (changes requested):
Nice, thank you for the patch! Just a small nit, otherwise LGTM :)

xdp-bench/xdp_redirect_cpumap.bpf.c
danobi: Whoops! Good catch

tohojo (approved):

PR #379, opened 2023-12-13
Added support for clang-17 in configure script

pavel-odintsov:
Hello!

I hope you're doing well.

Thank you very much for working on this code.

I recently did build with clang-17 from https://apt.llvm.org/ and it uses name of binary with suffix "-17": clang-17 which is not supported by configure script.

I added 17 as possible suffix for clang compiler name.

### DIFF
--- a/configure
+++ b/configure
@@ -39,7 +39,7 @@ find_tool()
 
     # we're looking for a binary with the same name as tool_name; try version
     # suffixes in order until we find one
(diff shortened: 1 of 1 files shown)


tohojo (approved):
Thank you for the patch! :)

RhennanCordeiro:
Thank you ❤️

PR #391, opened 2024-02-05
make: ensure every object/file that is built also gets removed on make clean

bluca:


tohojo (approved):
LGTM - thank you for the fixes!

PR #395, opened 2024-02-22
packaging/rpm: remove runtime dependency on kernel-header

shunghsiyu:
Seems like we don't really need `kernel-header` on the binary/non-devel packages as libxdp itself don't do any compilation at runtime (unlike bcc or bpftrace).

Having such dependency in `libxdp-devel` alone should be enough. WDYT?


tohojo:
Shung-Hsi Yu ***@***.***> writes:

> Seems like we don&#39;t really need `kernel-header` on the binary/non-devel packages as libxdp itself don&#39;t do any compilation at runtime (unlike bcc or bpftrace).
>
> Having such dependency in `libxdp-devel` alone should be enough. WDYT?

Hmm, yeah, I don't think there's any reason we need xdp-tools itself to
depend on kernel-headers. We probably need to add it as a BuildRequires,
then, though. There's a transient dependency through libbpf-devel, but
feels like it's better to add it explicitly?

shunghsiyu:
> Hmm, yeah, I don't think there's any reason we need xdp-tools itself to depend on kernel-headers. We probably need to add it as a BuildRequires, then, though. There's a transient dependency through libbpf-devel, but feels like it's better to add it explicitly?

Agree we should.

I've added it as a separate commit, but feel free to squash the two commits if you found them to be too verbose.

tohojo:
Squashed and merged as 3d7c83cc233a3343361e2c550b5ad1a4bdbc7c5d

shunghsiyu:
Thanks for the amendment, not sure what I was thinking when putting the `BuildRequire` down at libxdp-devel 😅

PR #408, opened 2024-03-23
Fix include path of libxdp for each arch

ami-GS:
Fix of #4 
Doesn't need to install gcc-multilib anymore nor set sympath as `ln -s /usr/include/x86_64-linux-gnu/asm /usr/include/asm`


tohojo:
Including arch-specific headers in a BPF program should not be needed; what error are you seeing that this fixes, exactly?

ami-GS:
This is the error. This doesn't happen with gcc-multilib or my change


root@861ec1d9b02c:~/msquic/submodules/xdp-tools# make

lib

  libbpf
    CC       libbpf/src/libbpf.a
    INSTALL  libbpf/src/libbpf.a

  libxdp
    CC       staticobjs/libxdp.o
    CC       staticobjs/xsk.o
    M4       xdp-dispatcher.c
    CLANG    xdp-dispatcher.o
In file included from xdp-dispatcher.c:3:
In file included from ../../headers/linux/bpf.h:11:
/usr/include/linux/types.h:5:10: fatal error: 'asm/types.h' file not found
#include <asm/types.h>
         ^~~~~~~~~~~~~
1 error generated.
make[2]: *** [Makefile:137: xdp-dispatcher.o] Error 1
make[1]: *** [Makefile:20: libxdp] Error 2
make: *** [Makefile:25: lib] Error 2

image

tohojo:
Hmm, okay, seems we do indeed use the arch includes for the uXX definitions. I wonder if that works correctly when compiling on a 32-bit arch? Anyway, I don't suppose there's anything wrong with fixing it this way, then :)

tohojo (approved):

PR #409, opened 2024-03-25
xdp-bench: Add support for xdp_buff with empty linear part

nimrod-oren:
In some network driver implementations, the linear part of the xdp_buff structure might be empty, with all data stored in the frags section. Accessing packet data directly is impossible in this case, and requires an API like the bpf_xdp_load/store_bytes helpers or dynptr.

This pull request ensures support for all xdp-bench packet operations (swap-macs, read-data, etc.) of basic actions (DROP, PASS, TX) when the linear part is empty.

The next step is to extend this functionality to the REDIRECT actions and explore integration with other tools beyond xdp-bench. After that, the plan is to introduce the option to use the superior dynptr API as another packet access alternative.

### DIFF
--- a/xdp-bench/tests/test-xdp-bench.sh
+++ b/xdp-bench/tests/test-xdp-bench.sh
@@ -36,6 +36,7 @@ test_xdp_load_bytes()
 
     for action in drop pass tx; do
         check_run $XDP_BENCH $action $NS -p parse-ip -l load-bytes -vv
+        check_run $XDP_BENCH $action $NS -p read-data -l load-bytes -vv
     done
 }
 
--- a/xdp-bench/tests/test-xdp-bench.sh
+++ b/xdp-bench/tests/test-xdp-bench.sh
@@ -37,6 +37,7 @@ test_xdp_load_bytes()
     for action in drop pass tx; do
         check_run $XDP_BENCH $action $NS -p parse-ip -l load-bytes -vv
         check_run $XDP_BENCH $action $NS -p read-data -l load-bytes -vv
+        check_run $XDP_BENCH $action $NS -p swap-macs -l load-bytes -vv
     done
 }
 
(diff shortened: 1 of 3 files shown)


tohojo:
nimrod-oren ***@***.***> writes:

> In some network driver implementations, the linear part of the
> xdp_buff structure might be empty, with all data stored in the frags
> section. Accessing packet data directly is impossible in this case,
> and requires an API like the bpf_xdp_load/store_bytes helpers or
> dynptr.

Which driver does that? None that's in-tree, AFAIK?

gal-pressman:
> nimrod-oren ***@***.***> writes:
> In some network driver implementations, the linear part of the xdp_buff structure might be empty, with all data stored in the frags section. Accessing packet data directly is impossible in this case, and requires an API like the bpf_xdp_load/store_bytes helpers or dynptr.
> Which driver does that? None that's in-tree, AFAIK?

mlx5 (in certain configurations) is an example.

nimrod-oren:
> > nimrod-oren _**@**_.***> writes:
> > In some network driver implementations, the linear part of the xdp_buff structure might be empty, with all data stored in the frags section. Accessing packet data directly is impossible in this case, and requires an API like the bpf_xdp_load/store_bytes helpers or dynptr.
> > Which driver does that? None that's in-tree, AFAIK?
> 
> mlx5 (in certain configurations) is an example.

Introduced in commit cd02a1a24897.

tohojo:
> Introduced in commit cd02a1a24897.

Huh, I never noticed that go in. I remember there was some discussion around it (in https://lore.kernel.org/r/20220621175402.35327-1-gospo@broadcom.com), but I also remember the outcome differently. Ah well, in that case I certainly have no objections to merging this :)

tohojo (approved):
LGTM, thanks for the patch :)

PR #410, opened 2024-03-25
xdp-bench: Fix packet-operation values in documentation

nimrod-oren:
Update the readme and manpage to match the values in xdp-bench.c.


tohojo (approved):
LGTM, thanks for the fix!

PR #411, opened 2024-03-25
xdp-bench: Add correct logging message for XDP_PASS

nimrod-oren:
Before this commit, the XDP_TX message was used for the XDP_PASS case.


tohojo (approved):
LGTM, thanks for the patch!

PR #418, opened 2024-04-19
Stop opening xdp-dispatcher.o when LIBXDP_SKIP_DISPATCHER=1 is set

ami-GS:
Even though LIBXDP_SKIP_DISPATCHER=1 is set, it still look for xdp-dispatcher.o and print stderror.
```
libxdp: Couldn't find a BPF file with name xdp-dispatcher.o
libxdp: Couldn't open BPF file xdp-dispatcher.o
```

Check the environment variable before opening file.

### DIFF
--- a/lib/libxdp/libxdp.c
+++ b/lib/libxdp/libxdp.c
@@ -1659,6 +1659,12 @@ static bool kernel_has_frags_support(void)
 	struct xdp_program *test_prog;
 	bool ret = false;
 	int err;
+	char *envval;
+
+	envval = secure_getenv(XDP_SKIP_ENVVAR);
+	if (envval && envval[0] == '1' && envval[1] == '\0') {
+		return false;
+	}
 
 	pr_debug("Checking for kernel frags support\n");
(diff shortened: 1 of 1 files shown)


tohojo:
This will break frags support when the environment variable is set, which is not a good idea. The `xdp-dispatcher.o` file is still needed even if the dispatcher is not used when loading things into the kernel.

ami-GS:
You mean loading xdp-dispatcher.o is required to check capabilities even if `LIBXDP_SKIP_DISPATCHER is set?
It means that `LIBXDP_OBJECT_PATH` is also still needed with `LIBXDP_SKIP_DISPATCHER`?

tohojo:
Daiki AMINAKA ***@***.***> writes:

> You mean loading xdp-dispatcher.o is required to check capabilities even if `LIBXDP_SKIP_DISPATCHER is set?
> It means that `LIBXDP_OBJECT_PATH` is also still needed with `LIBXDP_SKIP_DISPATCHER`?

Yes, unless you are statically linking the library, in which case the
dispatcher object will be embedded into the binary itself.

Issue #415, opened 2024-04-07
test_promiscuous_selfload and test_promiscuous_preload fail in Debian testing/unstable in QEMU


bluca:
Any idea what the issue could be?

```
     [test_promiscuous_selfload]   FAIL
          No XDP program loaded on xdptest
          PING fc42:dead:cafe:1::2 (fc42:dead:cafe:1::2) 56 data bytes
          64 bytes from fc42:dead:cafe:1::2: icmp_seq=1 ttl=64 time=0.160 ms
          64 bytes from fc42:dead:cafe:1::2: icmp_seq=2 ttl=64 time=0.114 ms
          64 bytes from fc42:dead:cafe:1::2: icmp_seq=3 ttl=64 time=0.122 ms
          64 bytes from fc42:dead:cafe:1::2: icmp_seq=4 ttl=64 time=0.128 ms
          
          --- fc42:dead:cafe:1::2 ping statistics ---
          4 packets transmitted, 4 received, 0% packet loss, time 3065ms
          rtt min/avg/max/mdev = 0.114/0.131/0.160/0.017 ms
          ERROR: Failed enabling promiscuous mode on legacy interface
          ==== RESULT: ====
          [  514.477824] xdptest: entered promiscuous mode
          [  519.565647] xdptest: left promiscuous mode
          ==== END ====
          Test test_promiscuous_selfload exited with return code: 1
     [test_promiscuous_preload]    FAIL
          PING fc42:dead:cafe:1::2 (fc42:dead:cafe:1::2) 56 data bytes
          64 bytes from fc42:dead:cafe:1::2: icmp_seq=1 ttl=64 time=0.223 ms
          
          --- fc42:dead:cafe:1::2 ping statistics ---
          1 packets transmitted, 1 received, 0% packet loss, time 0ms
          rtt min/avg/max/mdev = 0.223/0.223/0.223/0.000 ms
          ERROR: Failed enabling promiscuous mode on interface
          ==== RESULT: ====
          [  521.679030] xdptest: entered promiscuous mode
          [  523.692814] xdptest: left promiscuous mode
          ==== END ====
          Test test_promiscuous_preload exited with return code: 1
```

To reproduce, on a Debian/Ubuntu machine:

```
sudo apt install autopkgtest
git clone https://salsa.debian.org/debian/xdp-tools.git
cd xdp-tools
autopkgtest-build-qemu unstable ../autopkgtest-unstable-amd64.img
[...]

PR #428, opened 2024-06-19
xdp-dump: add check for new promiscuous mode message in tests

liangxiao1:
kernel updated the promiscuous mode message in below commit, so update the previous check condition.

net/core: refactor promiscuous mode message

old message:
[  406.034418] device eth0 entered promiscuous mode 
[  408.424703] device eth0 left promiscuous mode

new message:
[  406.034431] ice 0000:17:00.0 eth0: entered promiscuous mode 
[  408.424715] ice 0000:17:00.0 eth0: left promiscuous mode

### DIFF
--- a/xdp-dump/tests/test-xdpdump.sh
+++ b/xdp-dump/tests/test-xdpdump.sh
@@ -448,11 +448,11 @@ test_promiscuous_selfload()
     fi
 
     RESULT=$(dmesg)
-    if [[ "$RESULT" != *"device $NS entered promiscuous mode"* ]]; then
+    if [[ "$RESULT" != *"device $NS entered promiscuous mode"* ]] && [[ "$RESULT" != *"$NS: entered promiscuous mode"* ]]; then
         print_result "Failed enabling promiscuous mode on legacy interface"
         return 1
     fi
-    if [[ "$RESULT" != *"device $NS left promiscuous mode"* ]]; then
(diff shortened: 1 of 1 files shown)


liangxiao1:
I guess it fixes the failure mentioned in [415](https://github.com/xdp-project/xdp-tools/issues/415).

xdp-dump/tests/test-xdpdump.sh
```diff
@@ -448,11 +448,11 @@ test_promiscuous_selfload()
     fi
 
     RESULT=$(dmesg)
-    if [[ "$RESULT" != *"device $NS entered promiscuous mode"* ]]; then
+    if [[ "$RESULT" != *"device $NS entered promiscuous mode"* && "$RESULT" != *"$NS: entered promiscuous mode"* ]]; then
```
tohojo, line 451: Better to combine these as two distinct 'test' invocations, like:
`    if [[ "$RESULT" != *"device $NS entered promiscuous mode"* ]] && [[ "$RESULT" != *"$NS: entered promiscuous mode"* ]]; then`

same for all of the checks below, obviously :)

tohojo (changes requested):
Thank you for the fix! Just a small nit on the formatting :)

xdp-dump/tests/test-xdpdump.sh
liangxiao1: Thanks, @tohojo I updated and tested the patch as your suggestion.

tohojo (approved):

PR #429, opened 2024-07-30
xdp-forward: Introduce xdp-fwd-flowtable support

LorenzoBianconi:
Introduce xdp-fwd-flowtable sample in order to perform XDP_REDIRECT between net_devices inserted in a netfilter flowtable.

### DIFF
--- a/headers/linux/netfilter.h
+++ b/headers/linux/netfilter.h
@@ -1,0 +1,114 @@
+#ifndef _LINUX_NETFILTER_H
+#define _LINUX_NETFILTER_H
+
+#include <stdbool.h>
+#include <linux/types.h>
+#include <bpf/bpf_helpers.h>
+#include <xdp/parsing_helpers.h>
+
+#include "hlist.h"
+
+struct flow_ports {
+	__be16 source, dest;
+};
+
+enum ip_conntrack_dir {
+	IP_CT_DIR_ORIGINAL,
+	IP_CT_DIR_REPLY,
+	IP_CT_DIR_MAX
+};
+
+enum flow_offload_tuple_dir {
+	FLOW_OFFLOAD_DIR_ORIGINAL = IP_CT_DIR_ORIGINAL,
+	FLOW_OFFLOAD_DIR_REPLY = IP_CT_DIR_REPLY,
+};
+#define FLOW_OFFLOAD_DIR_MAX	IP_CT_DIR_MAX
+
+enum flow_offload_type {
+	NF_FLOW_OFFLOAD_UNSPEC,
+	NF_FLOW_OFFLOAD_ROUTE,
+};
+
+enum nf_flow_flags {
+	NF_FLOW_SNAT,
+	NF_FLOW_DNAT,
+	NF_FLOW_TEARDOWN,
+	NF_FLOW_HW,
+	NF_FLOW_HW_DYING,
+	NF_FLOW_HW_DEAD,
+	NF_FLOW_HW_PENDING,
+	NF_FLOW_HW_BIDIRECTIONAL,
+	NF_FLOW_HW_ESTABLISHED,
+};
+
+enum flow_offload_xmit_type {
+	FLOW_OFFLOAD_XMIT_UNSPEC,
+	FLOW_OFFLOAD_XMIT_NEIGH,
+	FLOW_OFFLOAD_XMIT_XFRM,
+	FLOW_OFFLOAD_XMIT_DIRECT,
+	FLOW_OFFLOAD_XMIT_TC,
+};
+
+#define NF_FLOW_TABLE_ENCAP_MAX		2
+struct flow_offload_tuple {
+	union {
+		struct in_addr		src_v4;
+		struct in6_addr		src_v6;
+	};
+	union {
+		struct in_addr		dst_v4;
+		struct in6_addr		dst_v6;
+	};
+	struct {
+		__be16			src_port;
+		__be16			dst_port;
+	};
+
+	int				iifidx;
+
+	__u8				l3proto;
+	__u8				l4proto;
+	struct {
+		__u16			id;
+		__be16			proto;
+	} encap[NF_FLOW_TABLE_ENCAP_MAX];
+
+	/* All members above are keys for lookups, see flow_offload_hash(). */
+	struct { }			__hash;
+
+	__u8				dir:2,
+					xmit_type:3,
+					encap_num:2,
+					in_vlan_ingress:2;
+	__u16				mtu;
+	union {
+		struct {
+			struct dst_entry *dst_cache;
+			__u32		dst_cookie;
+		};
+		struct {
+			__u32		ifidx;
+			__u32		hw_ifidx;
+			__u8		h_source[ETH_ALEN];
+			__u8		h_dest[ETH_ALEN];
+		} out;
+		struct {
+			__u32		iifidx;
+		} tc;
+	};
+};
+
+struct flow_offload_tuple_rhash {
+	struct rhash_head		node;
+	struct flow_offload_tuple	tuple;
+};
+
+struct flow_offload {
+	struct flow_offload_tuple_rhash		tuplehash[FLOW_OFFLOAD_DIR_MAX];
+	struct nf_conn				*ct;
+	unsigned long				flags;
+	__u16					type;
+	__u32					timeout;
+};
+
+#endif /* _LINUX_NETFILTER_H */
--- a/xdp-forward/xdp-forward.c
+++ b/xdp-forward/xdp-forward.c
@@ -1,0 +1,252 @@
+#define _GNU_SOURCE
+#include <stdio.h>
+#include <errno.h>
+#include <string.h>
+
+#include <xdp/libxdp.h>
+
+#include "params.h"
+#include "util.h"
+#include "logging.h"
+
+#include "xdp_forward.skel.h"
+
+#define MAX_IFACE_NUM 32
+#define PROG_NAME "xdp-forward"
+
+int do_help(__unused const void *cfg, __unused const char *pin_root_path)
+{
+	fprintf(stderr,
+		"Usage: xdp-forward COMMAND [options]\n"
+		"\n"
+		"COMMAND can be one of:\n"
+		"       load           - Load the XDP forwarding plane\n"
+		"       unload         - Unload the XDP forwarding plane\n"
+		"       help           - show this help message\n"
+		"\n"
+		"Use 'xdp-forward COMMAND --help' to see options for each command\n");
+	return -1;
+}
+
+
+struct enum_val xdp_modes[] = { { "native", XDP_MODE_NATIVE },
+				{ "skb", XDP_MODE_SKB },
+				{ NULL, 0 } };
+
+enum fwd_mode {
+	FWD_FIB_DIRECT,
+	FWD_FIB_FULL
+};
+
+struct enum_val fwd_modes[] = { { "fib-direct", FWD_FIB_DIRECT },
+				{ "fib-full", FWD_FIB_FULL },
+				{ NULL, 0 } };
+
+static int find_prog(struct iface *iface, bool detach)
+{
+	struct xdp_program *prog = NULL;
+	enum xdp_attach_mode mode;
+	struct xdp_multiprog *mp;
+	int ret = -ENOENT;
+
+	mp = xdp_multiprog__get_from_ifindex(iface->ifindex);
+	if (!mp)
+		return ret;
+
+	if (xdp_multiprog__is_legacy(mp)) {
+		prog = xdp_multiprog__main_prog(mp);
+		goto check;
+	}
+
+	while ((prog = xdp_multiprog__next_prog(prog, mp))) {
+	check:
+		if (!strcmp(xdp_program__name(prog), "xdp_fwd_fib_full") ||
+		    !strcmp(xdp_program__name(prog), "xdp_fwd_fib_direct")) {
+			mode = xdp_multiprog__attach_mode(mp);
+			ret = 0;
+			if (detach) {
+				ret = xdp_program__detach(prog, iface->ifindex,
+							  mode, 0);
+				if (ret)
+					pr_warn("Couldn't detach XDP program from interface %s: %s\n",
+						iface->ifname,
+						strerror(errno));
+				break;
+			}
+		}
+	}
+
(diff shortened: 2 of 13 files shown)


xdp-forward/xdp_forward.bpf.c
```diff
@@ -144,4 +158,552 @@ int xdp_fwd_direct(struct xdp_md *ctx)
...
+		if (tcph->fin || tcph->rst)
+			return -1;
+	}
+
+	return 0;
+}
+
+static __always_inline void
+xdp_flowtable_offload_nat_port(struct flow_ports *ports, void *data_end,
+			       __u8 proto, __be16 port, __be16 nat_port)
```
tohojo, line 228: Shouldn't this function be named `xdp_flowtable_update_port_csum()` or something to that effect?

xdp-forward/xdp_forward.bpf.c
```diff
@@ -144,4 +158,552 @@ int xdp_fwd_direct(struct xdp_md *ctx)
...
+	__be16 port, nat_port;
+
+	if (ports + 1 > data_end)
+		return;
+
+	switch (dir) {
+	case FLOW_OFFLOAD_DIR_ORIGINAL:
+		port = ports->dest;
+		bpf_core_read(&nat_port, bpf_core_type_size(nat_port),
+			      &flow->tuplehash[FLOW_OFFLOAD_DIR_REPLY].tuple.src_port);
```
tohojo, line 305: This looks backwards to me?
In the original direction, for dnat, don't we need the `dst_port` of the tuple? And vice versa below, of course. Either way, a comment somewhere explaining how the tuple works would be good :)

xdp-forward/xdp_forward.bpf.c
```diff
@@ -144,4 +158,552 @@ int xdp_fwd_direct(struct xdp_md *ctx)
...
+	default:
+		return;
+	}
+
+	xdp_flowtable_offload_nat_port(ports, data_end, proto, port, nat_port);
+}
+
+static __always_inline void
+xdp_flowtable_offload_ip_l4(struct iphdr *iph, void *data_end,
+			    __be32 addr, __be32 nat_addr)
```
tohojo, line 323: As above, `xdp_flowtable_update_ipv4_csum()`?

xdp-forward/xdp_forward.bpf.c
```diff
@@ -144,4 +158,552 @@ int xdp_fwd_direct(struct xdp_md *ctx)
...
+	}
+	iph->check = csum_replace4((__u32)iph->check, addr, nat_addr);
+
+	xdp_flowtable_offload_ip_l4(iph, data_end, addr, nat_addr);
+}
+
+static __always_inline void
+xdp_flowtable_offload_get_dnat_ip(const struct flow_offload *flow,
+				  enum flow_offload_tuple_dir dir,
+				  __be32 *addr)
```
tohojo, line 387: Most 'getter' type functions put the destination variable first (like indeed `bpf_core_read()` does inside the function), so let's do the same here?

xdp-forward/xdp_forward.bpf.c
```diff
@@ -144,4 +158,552 @@ int xdp_fwd_direct(struct xdp_md *ctx)
...
+		return;
+	}
+	iph->check = csum_replace4((__u32)iph->check, addr, nat_addr);
+
+	xdp_flowtable_offload_ip_l4(iph, data_end, addr, nat_addr);
+}
+
+static __always_inline void
+xdp_flowtable_offload_ipv6_l4(struct ipv6hdr *ip6h, void *data_end,
+			      struct in6_addr *addr, struct in6_addr *nat_addr)
```
tohojo, line 431: `xdp_flowtable_update_ipv6_csum()`?

xdp-forward/xdp_forward.bpf.c
```diff
@@ -144,4 +158,552 @@ int xdp_fwd_direct(struct xdp_md *ctx)
...
+		return;
+	}
+
+	xdp_flowtable_offload_ipv6_l4(ip6h, data_end, &addr, &nat_addr);
+}
+
+static __always_inline void
+xdp_flowtable_offload_get_dnat_ipv6(const struct flow_offload *flow,
+				    enum flow_offload_tuple_dir dir,
+				    struct in6_addr *addr)
```
tohojo, line 496: As above wrt argument order

xdp-forward/xdp_forward.bpf.c
```diff
@@ -144,4 +158,552 @@ int xdp_fwd_direct(struct xdp_md *ctx)
...
+	switch (tuplehash->tuple.xmit_type) {
+	case FLOW_OFFLOAD_XMIT_NEIGH:
+		/* update the destination address in case of dnatting before
+		 * performing the route lookup
+		 */
+		if (tuple.family == AF_INET6)
+			xdp_flowtable_offload_get_dnat_ipv6(flow, dir,
+					(struct in6_addr *)&tuple.ipv6_dst);
+		else
+			xdp_flowtable_offload_get_dnat_ip(flow, dir, &tuple.ipv4_src);
```
tohojo, line 685: This uses `ipv4_src`, but above it's `ipv6_dst`?

xdp-forward/xdp_forward.bpf.c
```diff
@@ -144,4 +158,552 @@ int xdp_fwd_direct(struct xdp_md *ctx)
...
+
+		if (bpf_fib_lookup(ctx, &tuple, sizeof(tuple), 0))
```
tohojo, line 687: Since the API defines constants for the result of this, let's explicitly check for `== BPF_FIB_LKUP_RET_SUCCESS`

xdp-forward/xdp_forward.bpf.c
```diff
@@ -144,4 +158,552 @@ int xdp_fwd_direct(struct xdp_md *ctx)
...
+	dir = tuplehash->tuple.dir;
+	if (dir >= FLOW_OFFLOAD_DIR_MAX)
+		return XDP_PASS;
+
+	flow = container_of(tuplehash, struct flow_offload, tuplehash[dir]);
+	if (bpf_core_read(&flags, sizeof(flags), &flow->flags))
+		return XDP_PASS;
+
+	switch (tuplehash->tuple.xmit_type) {
+	case FLOW_OFFLOAD_XMIT_NEIGH:
```
tohojo, line 677: Instead of a switch here, how about just making it an early return?

```c
if (tuplehash->tuple.xmit_type != FLOW_OFFLOAD_XMIT_NEIGH)
  return XDP_PASS;
```

This could even be combined with the direction value check above...

headers/linux/netfilter.h
```diff
@@ -0,0 +1,114 @@
...
+#include <stdbool.h>
+#include <linux/types.h>
+#include <bpf/bpf_helpers.h>
+#include <xdp/parsing_helpers.h>
+
+#include "hlist.h"
+
+struct flow_ports {
+	__be16 source, dest;
+} __attribute__((preserve_access_index));
```
tohojo, line 13: Why the attribute?

xdp-forward/xdp_forward.bpf.c
```diff
@@ -144,4 +158,552 @@ int xdp_fwd_direct(struct xdp_md *ctx)
...
+	struct flow_offload_tuple_rhash *tuplehash;
+	struct bpf_fib_lookup tuple = {
+		.ifindex = ctx->ingress_ifindex,
+	};
+	void *data = (void *)(long)ctx->data;
+	struct bpf_flowtable_opts opts = {};
+	enum flow_offload_tuple_dir dir;
+	struct ethhdr *eth = data;
+	struct flow_offload *flow;
+	struct flow_ports *ports;
```
tohojo, line 602: Using this struct as the pointer type for the L4 header is odd. AFAICT `struct flow_ports` is never used as an actual struct, it's always just cast to the L4 header. So why not just make this a `void *l4_header` pointer (everywhere, not just here), to make it clear that it's only ever used as an offset?

xdp-forward/xdp_forward.bpf.c
```diff
@@ -144,4 +158,552 @@ int xdp_fwd_direct(struct xdp_md *ctx)
...
+		tuple.dport		= ports->dest;
+		break;
+	}
+	case bpf_htons(ETH_P_IPV6): {
+		struct in6_addr *src = (struct in6_addr *)tuple.ipv6_src;
+		struct in6_addr *dst = (struct in6_addr *)tuple.ipv6_dst;
+		struct ipv6hdr *ip6h = data + sizeof(*eth);
+
+		ports = (struct flow_ports *)(ip6h + 1);
+		if (ports + 1 > data_end)
```
tohojo, line 641: This check doesn't really make much sense either? All the functions actually reading the L4 header redo the check with the correct length anyway...

xdp-forward/xdp_forward.bpf.c
```diff
@@ -144,4 +158,552 @@ int xdp_fwd_direct(struct xdp_md *ctx)
...
+	case bpf_htons(ETH_P_IP): {
+		struct iphdr *iph = data + sizeof(*eth);
+
+		ports = (struct flow_ports *)(iph + 1);
+		if (ports + 1 > data_end)
+			return XDP_PASS;
+
+		/* sanity check on ip header */
+		if (xdp_flowtable_offload_check_iphdr(iph) < 0)
+			return XDP_PASS;
```
tohojo, line 619: Having this as a separate function for v4 but not for v6 is a bit odd. Let's just move the two checks from the function inline here, so it's easier to follow what's going on.

xdp-forward/xdp-forward.c
```diff
@@ -81,17 +81,18 @@ static int find_prog(struct iface *iface, bool detach)
...
 struct prog_option load_options[] = {
-	DEFINE_OPTION("fib-mode", OPT_ENUM, struct load_opts, fib_mode,
+	DEFINE_OPTION("fwd-mode", OPT_ENUM, struct load_opts, fwd_mode,
 		      .short_opt = 'f',
-		      .typearg = fib_modes,
+		      .typearg = fwd_modes,
 		      .metavar = "<direct|full>",
-		      .help = "Perform full or direct (skipping rules) FIB lookups; default full"),
+		      .help = "Exec xdp-redirect performing lookup on the FIB table (full or direct); "
+		      "fib-direct skips fib-rules. Default value is fib-full"),
```
tohojo: The metavar should be updated with the new names as well

tohojo (changes requested):
A bunch of nits here and there.

Also, we should document how to use this (and what it does). But I guess I need to start by adding a README file to the xdp-forward utility that you can add that to. I'll do so, and get it merged into master; then you can just rebase this afterwards :)

xdp-forward/xdp_forward.bpf.c
LorenzoBianconi: @tohojo ack, I will fix it

xdp-forward/xdp_forward.bpf.c
LorenzoBianconi: @tohojo netfilter stores the selected destination port as source port for traffic in the opposite direction. I will add a comment.

xdp-forward/xdp_forward.bpf.c
LorenzoBianconi: @tohojo ack, I will fix it.

xdp-forward/xdp_forward.bpf.c
LorenzoBianconi: @tohojo ack, I will fix it.

xdp-forward/xdp_forward.bpf.c
LorenzoBianconi: @tohojo ack, I will fix it

xdp-forward/xdp_forward.bpf.c
LorenzoBianconi: @tohojo ack, right. We should use ipv4_dst to perform lookup in the fib table for IPv4. I will fix it.

xdp-forward/xdp_forward.bpf.c
LorenzoBianconi: @tohojo ack, I will fix it

xdp-forward/xdp_forward.bpf.c
LorenzoBianconi: @tohojo ack, fine.

headers/linux/netfilter.h
LorenzoBianconi: @tohojo I just copied it from kernel source code. I guess we can remove it

xdp-forward/xdp_forward.bpf.c
LorenzoBianconi: @tohojo this is needed for the verifier, otherwise it will complain about accessing ip header and l4 headers.

xdp-forward/xdp_forward.bpf.c
LorenzoBianconi: @tohojo ack, I will remove it

xdp-forward/xdp_forward.bpf.c
LorenzoBianconi: @tohojo this is same approach used by netfilter. Moreover it will be used later for packet boundary check.

xdp-forward/xdp-forward.c
LorenzoBianconi: @tohojo this should be updated.

xdp-forward/xdp_forward.bpf.c
tohojo: Right, "stores the selected destination port as source port for traffic *in the opposite direction*" - but here we are in the `DIR_ORIGINAL` branch, so why are we flipping it?

LorenzoBianconi: @tohojo this is because, when we do dnat, the destination port after performing the dnat can be different with respect to the "original" destination port of the incoming traffic, something like:

`tcp      6 src=a0.b0.c0.d0 dst=a1.b1.c1.d1 sport=63858 dport=11223 src=a2.b2.c2.d2 dst=a0.b0.c0.d0 sport=11224 dport=63858 [OFFLOAD] mark=0 use=2`

here the server is listening on port 11224 and we need to use sport=11224 in the reply direction as dport for the original traffic when we offload the dnat rule.

PR #432, opened 2024-08-06
libxdp: Expose xdp_program__attach_single()

mozzieongit:
Hi,
I (and potentially other library users) would like to load a single XDP program while skipping the dispatcher without modifying the environment.

I successfully tested this patch in my environment, but if there's another place that needs changing for exposing this function, please let me know.

### DIFF
--- a/headers/xdp/libxdp.h
+++ b/headers/xdp/libxdp.h
@@ -98,6 +98,8 @@ int xdp_program__attach(struct xdp_program *xdp_prog,
 int xdp_program__attach_multi(struct xdp_program **progs, size_t num_progs,
 			      int ifindex, enum xdp_attach_mode mode,
 			      unsigned int flags);
+int xdp_program__attach_single(struct xdp_program *prog, int ifindex,
+                               enum xdp_attach_mode mode);
 int xdp_program__detach(struct xdp_program *xdp_prog,
 			int ifindex, enum xdp_attach_mode mode,
 			unsigned int flags);
--- a/lib/libxdp/libxdp.c
+++ b/lib/libxdp/libxdp.c
@@ -1681,7 +1681,7 @@ static bool kernel_has_frags_support(void)
 }
 #endif // HAVE_LIBBPF_BPF_PROGRAM__FLAGS
 
-static int xdp_program__attach_single(struct xdp_program *prog, int ifindex,
+int xdp_program__attach_single(struct xdp_program *prog, int ifindex,
 				      enum xdp_attach_mode mode)
 {
 	int err;
--- a/lib/libxdp/libxdp.map
+++ b/lib/libxdp/libxdp.map
@@ -78,5 +78,6 @@ LIBXDP_1.3.0 {
 } LIBXDP_1.2.0;
 
 LIBXDP_1.4.0 {
+		xdp_program__attach_single;
 		xsk_umem__create_with_fd;
 } LIBXDP_1.3.0;


tohojo:
That function is not exposed on purpose: the dispatcher is needed for interoperability between different applications, and making it possible to use the library in a way that does not load the dispatcher mean it is no longer true that any two programs that load XDP programs via libxdp can coexist on the same system (and interface).

Could you please elaborate on why you want to load programs without the dispatcher? Perhaps there is a way to solve the problem you are seeing without doing away with the dispatcher entirely.

mozzieongit:
Ah, I see. But isn't that also the case with using the environment variable LIBXDP_SKIP_DISPATCHER?

The user-space program loads the XDP program, and shortly after drops privileges (except for `CAP_BPF` and `CAP_NET_ADMIN`). On program exit, I want to detach the XDP program, but the function `xdp_program__detach` requires `CAP_SYS_ADMIN`, so I switched to using the libbpf function `bpf_xdp_detach` (which might be bad design, mixing the two libraries, now that I think of it more...). Now, for there to be only one XDP program on the interface, I would need to set the environment variable LIBXDP_SKIP_DISPATCHER=1, so that the file descriptor of the XDP program can be used to unload it later, but I don't really want to touch the environment for that if I can do it in "pure code"/dedicated functions.

In summary, I would like to avoid having `CAP_SYS_ADMIN` around after the privilege drop, but still be able to unload our/the XDP program.

ajmmm:
> That function is not exposed on purpose: the dispatcher is needed for interoperability between different applications, and making it possible to use the library in a way that does not load the dispatcher mean it is no longer true that any two programs that load XDP programs via libxdp can coexist on the same system (and interface).
> 
> Could you please elaborate on why you want to load programs without the dispatcher? Perhaps there is a way to solve the problem you are seeing without doing away with the dispatcher entirely.

Just throwing in my two pence because it's come up before - apologies for hijacking the thread. :-)

I feel the assumption/assertion that, if one is using libxdp, one **must** want to load multiple programs, isn't true in every use case.

I work as part of a team that heavily depend on AF_XDP. We have built a very powerful packet inspection engine for our organisation - primarily focused around DDOS Mitigation.

Our stack runs in a highly managed environment that is not shared with users outside the team. The servers are dedicated to running the software we run on them. Our software takes a lot of programmatic control over every interface it's bound to. (We don't even subscribe to things like k8s or containers -- primarily because we aim to be as close to hardware as possible.)

Our system is compiled across a number of targets - and because it's sensitive to upstream changes, we pull in libbpf/libxdp at a repo level, and we statically link to them at build time. If we decide to re-base to a new libbpf/libxdp, we issue a new build of the software. (We actually pull in libbpf as its own dependency, build that ourselves, then force the libxdp build to use our libbpf build, because the libbpf submodule of libxdp does not track upstream libbpf!)

Our need to pull in libxdp as a dependency came about because AF_XDP higher level APIs got moved out of libbpf and into libxdp. It didn't come about for a desire to load multiple programs.

Some time ago you kindly accepted a PR that allowed us to switch the dispatcher off at build time. Perhaps this is a "good enough" fix for anyone who have a preference to remove the dispatcher - the condition they must accept is to manage the build of libxdp and use a static link instead.

ajmmm:
> Perhaps this is a "good enough" fix for anyone who have a preference to remove the dispatcher - the condition they must accept is to manage the build of libxdp and use a static link instead.

(Note: the benefit of a static link is that you're not actually affecting the system-level libxdp for those who do expect to just load a program via Bash.)

tohojo:
Well, the goal of libxdp is to give you the guarantee that "as long as
you use this library, you will be interoperable with all other users of
the library". And from a system integrator (distro) PoV, it is then
possible to say "we will only support XDP applications that use the
system libxdp, to ensure all applications on the system are
interoperable".

Which is why a compile-time switch is OK (users who control their system
and statically compile can do what they want), but exposing the
non-dispatcher API is not - the latter breaks the promise that using the
(system) libxdp ensures interoperability.

And yes, the LIBXDP_SKIP_DISPATCHER environment variable is bending this
a little bit. However, the intention is that since it's an environment
variable, it is set from outside the application, so it's a system-level
policy that can be changed without recompiling the application.

Now, being able to drop privileges and still unload programs is
certainly a valid use case, and I think it may be possible to support
this. See #434. I'll be away for the rest of this month, but I'll circle
back to this after I get back; feel free to comment and/or take a stab
at it in the meantime, of course :)

mozzieongit:
Ok got it, thank you

PR #436, opened 2024-09-09
Set BPF Skeleton name to base name of ebpf file

jalalmostafa:
The current build system sets the name of the generated skeleton directly to the file's name minus the `.skel.h` suffix.
Passing names of eBPF targets that contain a directory will result in wrong names e.g. `bpf/xdp_program.bpf.o` will have a skeleton file of the name `bpf/xdp_program.skel.h` and consequently the program name will be `bpf/xdp_program` instead of `xdp_program`. The commit pops out the directory part from the name.

### DIFF
--- a/lib/common.mk
+++ b/lib/common.mk
@@ -121,7 +121,7 @@ $(XDP_OBJ): %.o: %.c $(KERN_USER_H) $(EXTRA_DEPS) $(BPF_HEADERS) $(LIBMK)
 	$(QUIET_LLC)$(LLC) -march=$(BPF_TARGET) -filetype=obj -o $@ ${@:.o=.ll}
 
 $(BPF_SKEL_H): %.skel.h: %.bpf.o
-	$(QUIET_GEN)$(BPFTOOL) gen skeleton $< name ${@:.skel.h=} > $@
+	$(QUIET_GEN)$(BPFTOOL) gen skeleton $< name $(notdir ${@:.skel.h=}) > $@
 
 .PHONY: man
 ifeq ($(EMACS),)


jalalmostafa:
Not a bug but more of a feature. I use the xdp-tools build system as a scaffold for my libxdp-based programs. If I or someone else wants to have the eBPF programs in a special directory then he needs this.

tohojo:
Hmm, okay, well I guess this doesn't hurt to include. Do note that using the repository like this is not something that's explicitly supported, so I won't promise not to make changes that you'll have to adapt to :)

tohojo (changes requested):
Could you please update the commit message to include the rationale (including the fact that this does not affect anything currently in the repo)?

jalalmostafa:
I just fixed the commit message. I added removing *.ll files in nested directories in the `clean` target too.

lib/common.mk
```diff
@@ -66,7 +66,7 @@ all: $(USER_TARGETS) $(XDP_OBJ) $(EXTRA_TARGETS) $(TEST_TARGETS) man
 
 .PHONY: clean
 clean::
-	$(Q)rm -f $(USER_TARGETS) $(XDP_OBJ) $(TEST_TARGETS) $(USER_OBJ) $(TEST_OBJ) $(USER_GEN) $(BPF_SKEL_H) *.ll
+	$(Q)rm -f $(USER_TARGETS) $(XDP_OBJ) $(TEST_TARGETS) $(USER_OBJ) $(TEST_OBJ) $(USER_GEN) $(BPF_SKEL_H) *.ll **/*.ll
```
tohojo, line 69: Please drop this change. It is way too broad to include in the repository, and relies on the shell expanding `**` correctly, which is not a given.

tohojo (changes requested):
The DCO bot is complaining that the commit author and S-o-b tag; please fix that (you can reset the author with `git commit --amend --author=`, or just change the S-o-b tag.

tohojo:
Better, but now the commit message disappeared again :(

jalalmostafa:
Oops. Sorry. 
I just added it.

tohojo (approved):
Alright, that's better, thanks! :)

PR #445, opened 2024-10-14
Use socat as default utility for tcp/udp client/server connections

LorenzoBianconi:
Use socat as default utility for tcp/udp client/server connections

### DIFF
--- a/lib/testing/test_runner.sh
+++ b/lib/testing/test_runner.sh
@@ -25,7 +25,7 @@ ALL_TESTS=""
 VERBOSE_TESTS=${V:-0}
 NUM_NS=2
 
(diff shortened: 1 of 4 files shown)


tohojo (changes requested):
Please add a bit of description to the commit message, maybe something like:

> Right now we are using a mix of socat and nc in the selftests, so we depend on both tools. Since socat is more capable, and there are multiple incompatible versions of nc, let's move everything over to socat and get rid of the nc dependency.

LorenzoBianconi:
> Please add a bit of description to the commit message, maybe something like:
> 
> > Right now we are using a mix of socat and nc in the selftests, so we depend on both tools. Since socat is more capable, and there are multiple incompatible versions of nc, let's move everything over to socat and get rid of the nc dependency.

@tohojo ack, done

tohojo (approved):

PR #451, opened 2024-11-05
xdp-bench: Enable the xdp_frags flag for dummy programs

nimrod-oren:
Before this commit, the redirect dummy programs would fail to load on interfaces with non-linear MTU.


tohojo:
Any particular reason why this is marked as draft? :)

nimrod-oren:
ready for review :)

tohojo (changes requested):
Ah, seems I don't get an email notification when the state changes.

Could you please add the description (from the first comment on the PR) into the commit message as well? PR comments are not saved as part of the git history...

nimrod-oren:
Done.

PR #452, opened 2024-11-05
xdp-bench: Add support for the redirect action with --load-mode load-bytes

nimrod-oren:
This pull request extends the --load-mode flag functionality to xdp_redirect_basic.

This continues the work done in pull request #409,
to ensure support for the xdp-bench basic redirect action when the linear part is empty.


nimrod-oren:
Ready for review :)

PR #457, opened 2024-11-29
improve portability

rbgarga:
- **configure: improve portability**
- **libxdp: improve portability**
- **configure: detect systems without gcc**
- **libxdp: improve portability**

No functional changes intended

### DIFF
--- a/lib/libxdp/Makefile
+++ b/lib/libxdp/Makefile
@@ -131,7 +131,7 @@ $(TEMPLATED_SOURCES): %.c: %.c.in Makefile
 	$(QUIET_M4)$(M4) $(DEFINES) $< > $@ || ( ret=$$?; rm -f $@; exit $$ret )
 
 $(EMBEDDED_XDP_OBJS): %.embed.o: %.o
-	$(QUIET_GEN)$(LD) -r -b binary -o $@ -z noexecstack --format=binary $<
+	$(QUIET_CC)$(CC) -r -nostdlib -Wl,-z,noexecstack,--format=binary -o $@ $<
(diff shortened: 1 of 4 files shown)


rbgarga:
I'm still unable to build it on arm64 system due to linking error

```
ld -r -b binary -o xdp-dispatcher.embed.o -z noexecstack --format=binary xdp-dispatcher.o
ld: error: target emulation unknown: -m or at least one .o file required```

I don't know expected approach for this one, if you prefer to keep calling `ld` and pass `-m` parameter for current architecture or instead, use gcc/clang for linking

**Comment by rbgarga** [comment]
@tohojo one test failed in one env, is it possible to try just that again? I wonder if it's related to the proposed change

File: lib/libxdp/Makefile [hunk 1]
```diff
@@ -131,7 +131,7 @@ $(TEMPLATED_SOURCES): %.c: %.c.in Makefile
 	$(QUIET_M4)$(M4) $(DEFINES) $< > $@ || ( ret=$$?; rm -f $@; exit $$ret )
 
 $(EMBEDDED_XDP_OBJS): %.embed.o: %.o
-	$(QUIET_GEN)$(LD) -r -b binary -o $@ -z noexecstack --format=binary $<
+	$(QUIET_CC)$(CC) -r -nostdlib -Wl,-z,noexecstack,--format=binary -o $@ $<
```
tohojo, line 134: With this change we're not using $LD at all anymore, so I think we should get rid of that from the configure script to avoid confusion...

rbgarga: done

PR #468, opened 2025-01-14
xdpdump: A variable-sized type which should be placed at the end of a…

nanfengnan1:
… structure or union

if use -Werror,-Wgnu-variable-sized-type-not-at-end gnu flags in externel, will prevent compile code


tohojo:
Hmm, no, that is deliberate, the 'modes' member of ereq is there exactly to
contain the data pointed to by the variable-type member.

PR #475, opened 2025-02-11
Fix sample PPS report with longer poll intervals

tohojo:
The PPS values in xdp_sample output were calculated incorrectly if the reporting
interval is longer than one second. Fix this by reworking the accounting to
calculate packet and PPS values separately and printing the right values as
appropriate.


marcomole00:
I just did a quick test and the numbers are correct. Thanks @tohojo

tohojo:
Great, thanks for testing! :)

PR #477, opened 2025-02-17
libxdp: Fix incorrect setsockopt xdp_umem_reg size.

fnordahl:
The commit in the fixes tag made an uneccesary and incorrect attempt at backwards compatibility with kernels that do not support tx_metadata_len.

The kernel will either expect a size matching its version of struct xdp_umem_reg, if it encounters a shorter size, clamp it to 24 byte struct xdp_umem_reg_v1 [0].

The code in question attempted to provide a 28 byte length when tx_metadata_len was not set, which would invoke the kernel v1 processing.

This causes issues downstream as revealed in the Open vSwitch system testsuite.

I have confirmed that by removing this conditional, the test passes both on 6.7 and newer kernels.

0: https://github.com/torvalds/linux/blob/0ad2507d5d93/net/xdp/xsk.c#L1369-L1372
Fixes: 44b5e00692db ("libxdp: Add tx_metadata_len/xsk_umem__create_opts() to support AF_XDP Tx metadata")
Reported-at: https://launchpad.net/bugs/2098005
Signed-off-by: Frode Nordahl <fnordahl@ubuntu.com>

### DIFF
--- a/lib/libxdp/xsk.c
+++ b/lib/libxdp/xsk.c
@@ -297,7 +297,6 @@ struct xsk_umem *xsk_umem__create_opts(void *umem_area,
 				       struct xsk_umem_opts *opts) {
 	struct xdp_umem_reg mr;
 	struct xsk_umem *umem;
-	size_t mr_size;
 	int err, fd;
 	__u64 size;
 	
@@ -342,11 +341,7 @@ struct xsk_umem *xsk_umem__create_opts(void *umem_area,
 	mr.flags = umem->config.flags;
 	mr.tx_metadata_len = OPTS_GET(opts, tx_metadata_len, XSK_UMEM__DEFAULT_TX_METADATA_LEN);
 	
-	mr_size = sizeof(mr);
-	/* Older kernels don't support tx_metadata_len, skip if we are not setting a value */
-	if (!mr.tx_metadata_len)
-		mr_size = offsetof(struct xdp_umem_reg, tx_metadata_len);
-	err = setsockopt(umem->fd, SOL_XDP, XDP_UMEM_REG, &mr, mr_size);
+	err = setsockopt(umem->fd, SOL_XDP, XDP_UMEM_REG, &mr, sizeof(mr));
 	if (err) {
 		err = -errno;
 		goto out_socket;
--- a/lib/libxdp/tests/.gitignore
+++ b/lib/libxdp/tests/.gitignore
@@ -4,3 +4,4 @@ test_xdp_frags
 test_dispatcher_versions
 test_xsk_non_privileged
 test_link_detach
+test_xsk_umem_flags
--- a/lib/libxdp/tests/Makefile
+++ b/lib/libxdp/tests/Makefile
@@ -1,6 +1,6 @@
 # SPDX-License-Identifier: (GPL-2.0 OR BSD-2-Clause)
 
-USER_TARGETS := test_xsk_refcnt check_kern_compat test_xdp_frags test_dispatcher_versions test_link_detach
+USER_TARGETS := test_xsk_refcnt check_kern_compat test_xdp_frags test_dispatcher_versions test_link_detach test_xsk_umem_flags
 BPF_TARGETS := xdp_dispatcher_v1 xdp_pass
 USER_LIBS := -lpthread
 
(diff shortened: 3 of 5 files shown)


fnordahl:
cc @chaudron @igsilya for awareness. We discovered this while preparing an update to Open vSwitch in the current Ubuntu development release where we have xdp-tools 1.5.1.

lib/libxdp/xsk.c
```diff
@@ -343,9 +343,6 @@ struct xsk_umem *xsk_umem__create_opts(void *umem_area,
 	mr.tx_metadata_len = OPTS_GET(opts, tx_metadata_len, XSK_UMEM__DEFAULT_TX_METADATA_LEN);
 	
 	mr_size = sizeof(mr);
-	/* Older kernels don't support tx_metadata_len, skip if we are not setting a value */
-	if (!mr.tx_metadata_len)
-		mr_size = offsetof(struct xdp_umem_reg, tx_metadata_len);
```
tohojo, line 348: Let's just get rid of the mr_size variable completely, then...

tohojo:
Thanks for the fix! Just a small nit on the code, but otherwise looks good!

tohojo:
And just to make sure I understand the reason for the OVS regression, I assume that OVS is setting XDP_UMEM_UNALIGNED_CHUNK_FLAG, right?

fnordahl:
> And just to make sure I understand the reason for the OVS regression, I assume that OVS is setting XDP_UMEM_UNALIGNED_CHUNK_FLAG, right?

While it is the DPDK code that sets it, yes:
https://github.com/DPDK/dpdk/blob/e5176f23ae8b31437c3e5eb875c81f95bf3a9942/drivers/net/af_xdp/rte_eth_af_xdp.c#L1139

Should I perhaps mention this in the commit message?

tohojo:
Also, could you please add a test for this (creating umem with flags) to libxdp/tests? I guess we could just extend test_xsk_refcnt.c (and maybe rename it).

fnordahl:
> Frode Nordahl ***@***.***> writes:
> fnordahl left a comment ([xdp-project/xdp-tools#477](https://github.com/xdp-project/xdp-tools/pull/477)) > And just to make sure I understand the reason for the OVS regression, I assume that OVS is setting XDP_UMEM_UNALIGNED_CHUNK_FLAG, right? While it is the DPDK code that sets it, yes: https://github.com/DPDK/dpdk/blob/e5176f23ae8b31437c3e5eb875c81f95bf3a9942/drivers/net/af_xdp/rte_eth_af_xdp.c#L1139 Should I perhaps mention this in the commit message?
> Yes, please. AFAICT, the regression only affects users that set the flag, documenting that properly would be good :)

Hold on, when the size of the option is clamped, the kernel will not consider the flags due to the v1 processing:
```c
struct xdp_umem_reg_v1 {
	__u64 addr; /* Start of packet data area */
	__u64 len; /* Length of packet data area */
	__u32 chunk_size;
	__u32 headroom;
};
```

So I think the reason for the regression is that the application uses a chunk_size that is not a power of 2 (3008 in this case) and that the kernel does not consider the flags provided:
```c
	if (!unaligned_chunks && !is_power_of_2(chunk_size))
		return -EINVAL;
```

fnordahl:
> Also, could you please add a test for this (creating umem with flags) to libxdp/tests? I guess we could just extend test_xsk_refcnt.c (and maybe rename it).

In light of the previous comment, I guess the test would be making sure flags are processed by the kernel. Hopefully any future change on the kernel side will come with a better backward compatibility and such a test would help us catch it if they don't.

tohojo (changes requested):
Please split the test into its own commit. Otherwise, LGTM :)

tohojo:
Frode Nordahl ***@***.***> writes:

> fnordahl left a comment (xdp-project/xdp-tools#477)
>
>> Alright, looks good, and tests pass. Thanks for the fix, and the quick turnaround!
>> 
>> I'll tag a new release with this fix so we can get it into distributions ASAP :)
>
> Thank you for quick turnaround on reviews and release @tohojo, much
> appreciated! (JIT for February 20th Ubuntu feature freeze!!! :))

You're welcome! :)

Issue #473, opened 2025-02-03
Return value sign mix up


mrx3000:
Looks like the `errno` is returned directly (i.e. without its sign being negated) in `lib/libxdp/xsk.c`.
This propagates to such api functions as `xsk_socket__create()`, for example, resulting in them returning either an actual or a negated errno, depending on the specific case.

```
int xsk_socket__create_shared(..)
{
...
	if (!xsk_ptr)
		return -EFAULT;
...
	xsk = xsk_socket__create_opts(ifname, queue_id, umem, &opts);
	if (!xsk)
		return errno;
...
}
```

mrx3000:
P.S. The entirety of the proposed patch is:
```
diff --git a/lib/libxdp/xsk.c b/lib/libxdp/xsk.c
index 2e593f5..8374fe9 100644
--- a/lib/libxdp/xsk.c
+++ b/lib/libxdp/xsk.c
@@ -1288,7 +1288,7 @@ int xsk_socket__create_shared(struct xsk_socket **xsk_ptr,
        }
        xsk = xsk_socket__create_opts(ifname, queue_id, umem, &opts);
        if (!xsk)
-               return errno;
+               return -errno;
 
        *xsk_ptr = xsk;
        return 0;
```

tohojo:
Sure, that sounds reasonable - thanks for pointing it out! Care to open a PR with the proposed patch? :)

mrx3000:
I am a github.com noob, so pardon my ignorance. How can I create a pull request without access to the repo (and, as such, without ability to push a new branch)?

tohojo:
mrx3000 ***@***.***> writes:

> I am a github.com noob, so pardon my ignorance. How can I create a
> pull request without access to the repo (and, as such, without ability
> to push a new branch)?

You fork the repo to your own github account (button at the top right of
the repo page), create the commit in a branch there, then create a pull
request from the branch in your own repo to this one :)

PR #479, opened 2025-02-20
Fix function return to be -errno not errno

brian90013:
I believe the expected API behavior is functions return `-errno` on error. There are two locations, `xsk_umem__create_with_fd()` and `xsk_socket__create_shared()` where `return errno` is used instead of `return -errno`. This led libxdp library functions to return positive values. This problem was reported in issue #473.


tohojo (approved):
Thank you for the fix!

PR #481, opened 2025-02-23
buildsystem: honour OBJDIR

marckleinebudde:
The lib/libxdp/Makefile supports setting of OBJDIR, the directory where compiled object are placed. However when set, linking fails.

Set OBJDIR and pass it down the tests, so that linking works if OBJDIR is set.

### DIFF
--- a/lib/common.mk
+++ b/lib/common.mk
@@ -38,12 +38,12 @@ LIB_OBJS += $(foreach obj,$(UTIL_OBJS),$(LIB_DIR)/util/$(obj))
 EXTRA_DEPS +=
 EXTRA_USER_DEPS +=
 
-LDFLAGS+=-L$(LIBXDP_DIR)
+LDFLAGS+=-L$(LIBXDP_DIR)/$(OBJDIR)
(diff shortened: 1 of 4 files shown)


lib/common.mk
```diff
@@ -38,12 +38,12 @@ LIB_OBJS += $(foreach obj,$(UTIL_OBJS),$(LIB_DIR)/util/$(obj))
 EXTRA_DEPS +=
 EXTRA_USER_DEPS +=
 
-LDFLAGS+=-L$(LIBXDP_DIR)
+LDFLAGS+=-L$(LIBXDP_DIR)/$(OBJDIR)
```
tohojo, line 41: This implies that OBJDIR is always relative to LIBXDP_DIR - I am not sure there is any such guarantee? I.e., OBJDIR could be an absolute path?

How were you setting OBJDIR, exactly, when you ran into this?

marckleinebudde: Indeed I'm setting `OBJDIR=output`.

Will take care both both cases OBJDIR is absolute and relative.

PR #484, opened 2025-03-03
Support building with Emacs 30

K900:


configure
```diff
@@ -97,7 +97,7 @@ check_toolchain()
         EMACS=""
     else
         emacs_version=$($EMACS --version 2>/dev/null | head -n 1)
-        if echo $emacs_version | grep -Eq 'GNU Emacs 2[6789]'; then
+        if echo $emacs_version | grep -Eq 'GNU Emacs (2[6789]|30)'; then
```
tohojo, line 100: Let's make this future-proof - something like `3[0-9]` instead of `30`

K900: Is the intent to just support any version newer than 26? Maybe in that case we should just parse the version properly?

tohojo: Yeah, basically. Parsing it properly is fine as well :)

K900: Tried my take at doing a proper(ish) version check.

configure
```diff
@@ -96,8 +96,10 @@ check_toolchain()
     if ! command -v $EMACS >/dev/null 2>&1; then
         EMACS=""
     else
-        emacs_version=$($EMACS --version 2>/dev/null | head -n 1)
-        if echo $emacs_version | grep -Eq 'GNU Emacs 2[6789]'; then
+        # extract major version from "GNU Emacs [major].[minor]"
+        emacs_version=$($EMACS --version 2>/dev/null | head -n 1 | cut -d " " -f 3)
+        emacs_major=$(echo "$emacs_version" | cut -d "." -f 1)
```
jian-lin, line 101: You can get Emacs major version by printing `emacs-major-version`.

```console
> emacs -Q --batch --eval='(message "%s" emacs-major-version)'
30
```

K900: Applied that, and also cleaned up the errors a bit.

tohojo (approved):
Looks good, thanks!

PR #486, opened 2025-03-03
Increase MAX_CPUS to 512

### LINKED ISSUE #485: Increase MAX_CPUS
On AMD EPYC systems with two processors of 96 cores  and hyperthreading enabled, one has 384 CPUs. 
The xdpdump tool fails on such systems. 
Please increase MAX_CPUS to 512
I will add a small pull request for that

lucashc:
Increase MAX_CPUS from 256 to 512 to support higher cpu count systems.
Fixes #485

### DIFF
--- a/xdp-dump/xdpdump.h
+++ b/xdp-dump/xdpdump.h
@@ -11,7 +11,7 @@
  ******************************************************************************/
 #define PERF_MAX_WAKEUP_EVENTS   64
 #define PERF_MMAP_PAGE_COUNT	256
-#define MAX_CPUS		256
+#define MAX_CPUS		512
 
 /******************************************************************************
  * General used macros


tohojo (changes requested):
Thanks for the fix! Please add the information from the issue about which systems this affects to the commit message, and add a signed-off-by line as well.

lucashc:
I added the requested changes and updated the commit message.
One more thought: The Linux kernel allows many more CPUs than just 512. For instance, the kernel of my Ubuntu 24.04 LTS has been configured to support 8192 CPUs. Should we increase it even further?
On the other hand, most common servers will not exceed 512 CPUs.

tohojo (approved):
Looks good, thanks!

I don't think we should increase it further as this adds overhead for all systems (being a compile-time value). If we need to scale to a higher number of CPUs we should start sizing things dynamically at runtime (or maybe find some way to avoid the per-CPU stuff if possible).

PR #497, opened 2025-03-28
xdp-forward-queueing: Update outstanding bytes

luisjira:
At the moment, xdp-forward-queueing includes logic to track the number of outstanding bytes in hardware, but it is unused.

This pull request attaches the frame return logic to the corresponding tracepoint to allow subtracting the completed data. It also adds the counter increment when data is sent.

### DIFF
--- a/xdp-forward/xdp-forward.c
+++ b/xdp-forward/xdp-forward.c
@@ -2,6 +2,8 @@
 #include <stdio.h>
 #include <errno.h>
 #include <string.h>
+#include <unistd.h>
+#include <errno.h>
 
 #include <bpf/libbpf.h>
 #include <xdp/libxdp.h>
@@ -140,7 +142,10 @@ static int do_load(const void *cfg, __unused const char *pin_root_path)
 {
 	struct xdp_program *xdp_prog = NULL, *init_prog = NULL;
 	DECLARE_LIBBPF_OPTS(xdp_program_opts, opts);
+	struct bpf_program *return_prog = NULL;
 	const struct load_opts *opt = cfg;
+	struct bpf_link *link = NULL;
+	char pin_link_path[PATH_MAX];
 	struct xdp_forward *skel;
 	int ret = EXIT_FAILURE;
 	struct iface *iface;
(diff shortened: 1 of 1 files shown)


xdp-forward/xdp-forward.c
```diff
@@ -217,6 +222,34 @@ static int do_load(const void *cfg, __unused const char *pin_root_path)
 		pr_info("Loaded on interface %s\n", iface->ifname);
 	}
 
+	/* Attach xdp_check_return to xdp_frame_return tracepoint and
+	 * pin the link to pin_root_path/xdp_check_return.
+	 */
+
+	return_prog = bpf_object__find_program_by_name(skel->obj,"xdp_check_return");
```
tohojo, line 229: The program is available in the skeleton as skel->progs.xdp_check_return - no need to use `bpf_object__find_program_by_name()`

xdp-forward/xdp-forward.c
```diff
@@ -217,6 +222,34 @@ static int do_load(const void *cfg, __unused const char *pin_root_path)
...
+	if (!return_prog) {
+		pr_warn("Failed to find xdp_check_return program\n");
+		goto end_detach;
+	}
+
+	link = bpf_program__attach_raw_tracepoint(return_prog, "xdp_frame_return");
```
tohojo, line 235: It should be possible to just call `bpf_program__attach()`, since the target is specified in the section definition in the BPF source file.

xdp-forward/xdp-forward.c
```diff
@@ -217,6 +222,34 @@ static int do_load(const void *cfg, __unused const char *pin_root_path)
...
+	if (libbpf_get_error((const void *) link)) {
```
tohojo, line 236: It should be possible to just pass the pointers to `libbpf_get_error()` without the cast...

xdp-forward/xdp-forward.c
```diff
@@ -217,6 +222,34 @@ static int do_load(const void *cfg, __unused const char *pin_root_path)
...
+		pr_warn("Couldn't attach to xdp_frame_return: %s\n",
+			strerror(libbpf_get_error((const void *) link)));
```
tohojo, line 238: ...same here

xdp-forward/xdp-forward.c
```diff
@@ -217,6 +222,34 @@ static int do_load(const void *cfg, __unused const char *pin_root_path)
...
+		goto end_detach;
+	}
+	skel->links.xdp_check_return = link;
+
+	snprintf(pin_link_path, sizeof(pin_link_path), "%s/%s", pin_root_path,
```
tohojo, line 243: This should use the `try_snprintf()` function (from lib/util), and check the return code. We should probably also disambiguate the pin path right now, with this path bad things will happen if multiple instances of xdp-forward are loaded at the same time.

xdp-forward/xdp-forward.c
```diff
@@ -225,6 +258,10 @@ static int do_load(const void *cfg, __unused const char *pin_root_path)
 end_detach:
 	for (iface = opt->ifaces; iface; iface = iface->next)
 		xdp_program__detach(xdp_prog, iface->ifindex, opt->xdp_mode, 0);
+	
+	bpf_link__unpin(link);
```
tohojo, line 262: `bpf_link__unpin()` will crash if `link` is NULL

xdp-forward/xdp-forward.c
```diff
@@ -258,6 +297,24 @@ static int do_unload(const void *cfg, __unused const char *pin_root_path)
 		pr_info("Unloaded from interface %s\n", iface->ifname);
 	}
 
+	snprintf(pin_link_path, sizeof(pin_link_path), "%s/%s", pin_root_path,
```
tohojo, line 300: See above wrt using `try_snprintf()`

xdp-forward/xdp-forward.c
```diff
@@ -258,6 +297,24 @@ static int do_unload(const void *cfg, __unused const char *pin_root_path)
...
+			 "xdp_check_return");
+
+	link = bpf_link__open(pin_link_path);
```
tohojo, line 303: Instead of going through this open/unpin/destroy dance, just unlink the pin file?

xdp-forward/xdp-forward.c
```diff
@@ -217,6 +222,34 @@ static int do_load(const void *cfg, __unused const char *pin_root_path)
...
+
+	snprintf(pin_link_path, sizeof(pin_link_path), "%s/%s", pin_root_path,
+			 "xdp_check_return");
+	ret = bpf_link__pin(link, pin_link_path);
+	if (ret) {
+		pr_warn("Failed to pin link: %s\n",
+			strerror(errno));
+		goto end_detach;
+	}
+	pr_info("Pinned xdp_check_return to %s\n", pin_link_path);
```
tohojo, line 251: This should probably be `pr_debug()`

tohojo (changes requested):
Thank you for the patch; a bunch of nits, see below.

PR #504, opened 2025-05-02
xdp-forward: added VLANS support

enhaut:
This patch adds support for VLAN processing. All the directions are supported:
* tagged network to tagged
* tagged to untagged
* untagged to tagged

However, this adds ~4% overhead, therefore it's disabled by default. To enabled it, `xdp-forward` needs to be recompiled with either `VLANS_USERSPACE` or `VLANS_PATCHED` based on desired mode. 2 modes are supported:

* userspace - before loading `xdp-forward`, userspace part uses `netlink` to get all the VLAN interfaces on top of `xdp-forward`-enabled devices. The map that maps vlan ifindex to underlying physical ifindex+VLAN ID is then passed to XDP program. (`bpf_fib_lookup` returns VLAN interface ifindex when packet is forwarded to VLANed network, but VLAN interfaces does not implement xmit function for xdp. Therefore, packet needs to be sent out of physical inf). Limitation of this version is, that mapping map is not automatically updated, so when VLAN interface is changed (removed, added, vlan id) `xdp-forward` requires manual reload.
* kernel patched - this is "native" mode which requires patched kernel with a patch, that extends `bpf_fib_lookup` for support of physical device lookup for vlan interfaces. (@tohojo is working on this patch)

Performance comparison:

vlans_compared

Before this patch, the performance of VLAN tagged traffic was even slower with `xdp-forward` enabled as it was just passing it to the stack but added some overhead.

### DIFF
--- a/xdp-forward/Makefile
+++ b/xdp-forward/Makefile
@@ -5,6 +5,11 @@ CFLAGS += -DVLANS_PATCHED
 BPF_CFLAGS += -DVLANS_PATCHED
 endif
 
+ifdef VLANS_USERSPACE
+CFLAGS += -DVLANS_USERSPACE
+BPF_CFLAGS += -DVLANS_USERSPACE
+endif
+
 XDP_TARGETS := xdp_forward.bpf xdp_flowtable.bpf xdp_flowtable_sample.bpf
 BPF_SKEL_TARGETS := $(XDP_TARGETS)
 
--- a/xdp-forward/Makefile
+++ b/xdp-forward/Makefile
@@ -1,5 +1,10 @@
 # SPDX-License-Identifier: GPL-2.0
 
+ifdef VLANS_PATCHED
+CFLAGS += -DVLANS_PATCHED
+BPF_CFLAGS += -DVLANS_PATCHED
+endif
+
 XDP_TARGETS := xdp_forward.bpf xdp_flowtable.bpf xdp_flowtable_sample.bpf
 BPF_SKEL_TARGETS := $(XDP_TARGETS)
 
--- a/xdp-forward/README.org
+++ b/xdp-forward/README.org
@@ -167,7 +167,19 @@ EOF
 
 * VLAN support
 Vlan support adds ~4% overhead for packet processing. Therefore, it's disabled
-by default. To enable it, compilation flag is required.
+by default. To enable it, compilation flags are required, based on desired mode.
+
+** Userspace mode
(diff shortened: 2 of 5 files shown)


csszep:
Hi!

Is there any news about the integration of the VLAN patch?

Thx

tohojo:
> Is there any news about the integration of the VLAN patch?

Sadly, no. I got sidetracked before I managed to submit the kernel patch
to add VLAN support to the fib lookup, and this stalled out as a result.
As I recall, that was quite a small patch, so should be possible to
resurrect it.

Alternatively, if someone wants to rebase this and drop the kernel patch
path, we can merge that first and add the other thing once the kernel
patch lands.

Strykar:
@enhaut Following up on @tohojo's "alternatively" suggestion above.
Happy to rebase this and split out the `VLANS_USERSPACE` half so it can land on its own, if you're okay with me picking it up from here.

In parallel I have the kernel-side `BPF_FIB_LOOKUP_VLAN` patch built and runtime-verified, shippable to bpf-next today, so `VLANS_PATCHED` can come back as a follow-up once that's in mainline.

tohojo:
"Avinash H. Duduskar" ***@***.***> writes:

> Strykar left a comment (xdp-project/xdp-tools#504)
>
> @enhaut Following up on @tohojo's "alternatively" suggestion above.
> Happy to rebase this and split out the `VLANS_USERSPACE` half so it can land on its own, if you're okay with me picking it up from here.
>
> In parallel I have the kernel-side `BPF_FIB_LOOKUP_VLAN` patch built and runtime-verified, shippable to bpf-next today, so `VLANS_PATCHED` can come back as a follow-up once that's in mainline.

Sure, please go ahead with both of those :)

enhaut:
> Alternatively, if someone wants to rebase this and drop the kernel patch
path, we can merge that first and add the other thing once the kernel
patch lands.


Sorry, I missed this somehow...


@Strykar looks like you already spent some time on this :) feel free to go ahead

Strykar:
I have picked this up as discussed above; the userspace-only version is at https://github.com/xdp-project/xdp-tools/pull/584. It is a fresh implementation on top of the parsing-helpers refactor rather than a rebase of these patches, with the design notes in the PR description. The kernel side (`BPF_FIB_LOOKUP_VLAN`) is in bpf-next, and the `VLANS_PATCHED` half follows once it reaches a released kernel.

enhaut:
Closing, @Strykar reimplemented this in #584

PR #506, opened 2025-05-09
install xdp_forward XDP bpf program

vincentmli:
`XDP_OBJ_INSTALL := `stops `make install` to install xdp_forward.bpf xdp_flowtable.bpf xdp_flowtable_sample.bpf program. for linux distribution, we need to install these bpf programs.


vincentmli:
@tohojo I build xdp-tools and run `make install`  to install xdp-tools user and bpf program in BPFire open source firewall OS, noticed xdp-forward bpf program not installed. please review and see if it makes sense :)

vincentmli:
oh, it looks xdp-forward user space program already included the bpf program, no need to install the bpf program, not installing the bpf program makes sense to me too

Issue #507, opened 2025-05-13
xdpdump does not support overriding BPF file location in environment


lschrecken:
In our yocto build it is not possible to use xdpdump. The problem is that xdpdump looks for the file xdpdump_bpf.o in a path of the host build system, even if LIBXDP_OBJECT_PATH is set to the build's local path where xdpdump_bpf.o is stored.

I looked at the code and found two possible causes:
1. It seems like the find_bpf_file function in the file util.c was forgotten to be changed when the feature was introduced in libxdp with commit https://github.com/xdp-project/xdp-tools/commit/b68065705c58ace3b6f20df1c329316cbd0c077b. Therefore the util.c variant does not use the environment variable and instead uses the BPF_OBJECT_PATH which is set at compile time. https://github.com/xdp-project/xdp-tools/blob/5950651ba76e43f55b1ae9e0e8f47173f4704ac4/lib/util/util.c#L151  https://github.com/xdp-project/xdp-tools/blob/5950651ba76e43f55b1ae9e0e8f47173f4704ac4/lib/libxdp/libxdp.c#L1284
2. in this case I'm not so sure, but there might be a logical error in the xdpdump function that leads to the call to find_bpf_file(). https://github.com/xdp-project/xdp-tools/blob/5950651ba76e43f55b1ae9e0e8f47173f4704ac4/xdp-dump/xdpdump.c#L1637 At the end of the if condition in line 1675, load_xdp is not set to true, even if find_target() was successful. As a result, the wrong branch is used in line 1698.

When I fix these locally, I can call xdpdump.

tohojo:
Lukas Schreckenberg ***@***.***> writes:

> lschrecken created an issue (xdp-project/xdp-tools#507)
>
> In our yocto build it is not possible to use xdpdump. The problem is
> that xdpdump looks for the file xdpdump_bpf.o in a path of the host
> build system, even if LIBXDP_OBJECT_PATH is set to the build's local
> path where xdpdump_bpf.o is stored.
>
> I looked at the code and found two possible causes:
> 1. It seems like the find_bpf_file function was forgotten to be
> changed when the feature was introduced in libxdp with commit
> https://github.com/xdp-project/xdp-tools/commit/b68065705c58ace3b6f20df1c329316cbd0c077b.
> Therefore the util.c variant does not use the environment variable and
> instead uses the BPF_OBJECT_PATH which is set at compile time.
[...]

PR #508, opened 2025-05-19
xdpdump: Adapt the util.c find_bpf_file() to the libxdp.c variant

lschrecken:
In contrast to the lib/libxdp/libxdp.c variant, the function find_bpf_file() in the lib/util/util.c file does not check and use the environment variable LIBXDP_OBJECT_PATH to overwrite the default path, but always uses the default variable BPF_OBJECT_PATH, which is set at compile time. 

To unify the behavior, the behavior of the util.c variant is adapted to that of the libxdp.c variant.

Concerning issue #507 

Signed-off-by: Lukas Schreckenberg <lukas.schreckenberg@iosb-ina.fraunhofer.de>

### DIFF
--- a/lib/util/util.c
+++ b/lib/util/util.c
@@ -1,5 +1,7 @@
 /* SPDX-License-Identifier: GPL-2.0 */
 
+#define _GNU_SOURCE
+
 #include <errno.h>
 #include <unistd.h>
 #include <stdlib.h>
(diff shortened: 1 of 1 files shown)


tohojo:
Looks good, but please add the signed-off-by tag to the commit message as well (to keep the DCO bot happy) :)

PR #518, opened 2025-07-24
xdp-trafficgen: fix probe_options missing END_OPTIONS

liangxiao1:
Probe does not work w/o "-i" option.
Fix the error by adding END_OPTIONS to probe_options.

$ sudo xdp-trafficgen probe -i ens5
Missing required option '--interface'

Usage: xdp-trafficgen [options] <hostname>
Use --help (or -h) to see full option list.


tohojo:
Oops - thank you for the fix!

PR #519, opened 2025-07-27
Doc: update libxdp README.org architecture support

vincentmli:
XDP multiprog on a single interface generally is supported by architectures supporting BPF trampoline.


vincentmli:
oh bad PR, close it

PR #520, opened 2025-07-27
Doc: update libxdp README.org and man page architecture support

vincentmli:
XDP multiprog on a single interface generally is supported by architectures supporting BPF trampoline.

### DIFF
--- a/lib/libxdp/README.org
+++ b/lib/libxdp/README.org
@@ -423,12 +423,13 @@ kernel and first appeared in kernel release 5.6. To *incrementally* attach
 multiple programs, a further refinement added by commit 4a1e7c0c63e0 ("bpf:
 Support attaching freplace programs to multiple attach points") is needed; this
 first appeared in the upstream kernel version 5.10. The functionality relies on
-the "BPF trampolines" feature which is unfortunately only available on the
-x86_64 architecture. In other words, kernels before 5.6 can only attach a single
-XDP program to each interface, kernels 5.6+ can attach multiple programs if they
-are all attached at the same time, and kernels 5.10 have full support for XDP
-multiprog on x86_64. On other architectures, only a single program can be
-attached to each interface.
+the "BPF trampolines" feature which is available on architectures that support it.
+
+The support matrix is as follows:
+- Kernels before 5.6 can only attach a single XDP program to each interface
+- Kernels 5.6+ can attach multiple programs if they are all attached at the same time
+- Kernels 5.10+ have full support for XDP multiprog on architectures supporting BPF trampolines
+- On architectures without BPF trampoline support, only a single program can be attached to each interface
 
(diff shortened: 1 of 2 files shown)


vincentmli:
got it,  lib/libxdp/libxdp.3 need to be updated

tohojo:
Erm, you just need to make sure you have emacs installed on the system, then run 'make' to re-generate the man pages, and add the man page update to the same commit as the README.org change :)

vincentmli:
I am not familiar with emacs editor, do I need to install emacs and use emacs to edit the libxdp.3 man page file? or just install emacs and run `make`? I just installed emacs and run `make`, nothing changed after `make`. sorry I am not familiar with man page generation.

vincentmli:
ah, I see, I run ./configure and make, it indeed make changes, but there is unrelated format changes to libxdp.3 like below when I run `git diff`, is that ok to commit the diff ?


```
diff --git a/lib/libxdp/libxdp.3 b/lib/libxdp/libxdp.3
index 77cde47..ce07e60 100644
--- a/lib/libxdp/libxdp.3
+++ b/lib/libxdp/libxdp.3
@@ -1,4 +1,5 @@
-.TH "libxdp" "3" "January 14, 2025" "v1.5.6" "libxdp - library for loading XDP programs" 
+.TH "libxdp" "3" "September 14, 2025" "v1.5.6" "libxdp - library for loading XDP programs" 
+
 .SH "NAME"
 libxdp \- library for attaching XDP programs and using AF_XDP sockets
 .SH "SYNOPSIS"
@@ -21,6 +22,7 @@ well as reading and writing packets from these sockets.
 .PP
 Some of the functionality provided by libxdp depends on particular kernel
 features; see the "Kernel feature compatibility" section below for details.
+
 .SS "Using libxdp from an application"
 .PP
 Basic usage of libxdp from an application is quite straight forward. The
@@ -53,13 +55,13 @@ program that is already loaded into the kernel:
 .RS
 .nf
 \fCstruct xdp_program *xdp_program__from_bpf_obj(struct bpf_object *obj,
-                                              const char *section_name);
+                                             const char *section_name);
 struct xdp_program *xdp_program__find_file(const char *filename,
-                                           const char *section_name,
-                                           struct bpf_object_open_opts *opts);
+                                          const char *section_name,
+                                          struct bpf_object_open_opts *opts);
...

```

tohojo:
Hmm, yeah, the unrelated format changes happen sometimes because of variations in the output of different emacs versions. They're not a huge issue, so it's fine if you just include them in the libxdp man page; don't include any of the other man page updates in the commit though (only `git add` the libxdp.3 file) :)

tohojo:
Merged as 89640b133653df88044d369e9d2012f48895c362 - thanks! :)

PR #562, opened 2026-02-26
xdp-forward/xdp_flowtable: Fix backward compatibility verifier issue

LorenzoBianconi:
Starting from commit 'ab427db17885 ("netfilter: flowtable: Add IPIP rx sw acceleration"), flow_offload_tuple layout has changed in order to support IP tunnel offload. Rely on bpf-core APIs to maintain backward compatibility on impacted fields.

### DIFF
--- a/xdp-forward/xdp_flowtable.bpf.c
+++ b/xdp-forward/xdp_flowtable.bpf.c
@@ -125,7 +125,6 @@ xdp_flowtable_update_port_csum(struct flow_ports *ports, void *data_end,
 		break;
 	}
 }
-
 static __always_inline void
 xdp_flowtable_snat_port(const struct flow_offload *flow,
 			struct flow_ports *ports, void *data_end,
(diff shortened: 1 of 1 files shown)


xdp-forward/xdp_flowtable.bpf.c
```diff
@@ -576,32 +619,16 @@ static __always_inline int xdp_flowtable_flags(struct xdp_md *ctx,
...
-
 	if (tuple.family == AF_INET6)
 		xdp_flowtable_forward_ipv6(flow, data, data_end, ports, dir,
 					   flags);
 	else
 		xdp_flowtable_forward_ip(flow, data, data_end, ports, dir,
 					 flags);
 
+	if (xdp_flowtable_route_lookup(ctx, &tuple, fib_flags))
+		return XDP_PASS;
```
tohojo: Hmm, this could potentially pass a packet up the stack that has already been modified, no? Why move the lookup to after the packet is modified?

LorenzoBianconi: @tohojo uhm, right. We need to update the touple  because the lookup must be done after performing dnat, but I guess we can postpone the packet update after the route lookup

tohojo: Yeah, splitting the two does seem like the right approach...

tohojo (approved):

PR #563, opened 2026-02-26
xdpdump: include <limits.h> removed from linux/ethtool.h

nuqleo:
Removing include <limits.h> from linux/ethtool.h resulted in a undeclared UINT_MAX and ULONG_MAX compilation errors

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=a8a11e5237aed71b7f5f9d33c554ef06fe974311

Added #include <limits.h> in xdp-dump/xdpdump.c


tohojo (approved):
Thank you for the fix!

PR #567, opened 2026-04-10
xdpsock: Fix partial batch handling

legezywzh:
If l2fwd or rx_drop stops partway through a peeked batch, make sure the untouched descriptors are cancelled before we loop around again.


tohojo (changes requested):
OK, so I think you're correct that this is needed, but it's by no means obvious why. So please update the commit message with an explanation of why this change is needed (and by extension, what buggy behaviour it fixes).

legezywzh:
> OK, so I think you're correct that this is needed, but it's by no means obvious why. So please update the commit message with an explanation of why this change is needed (and by extension, what buggy behaviour it fixes).

OK, sure. This issue was found while testing AF_XDP with the xdp-bench tool. I'll update the commit message to explain why this change is needed and what bug it fixes, and send a revised version soon.

tohojo:
Great, thank you! :)

tohojo (approved):
Excellent, thank you! :)

PR #569, opened 2026-04-11
default CC to cc instead of gcc

### LINKED ISSUE #568: [clang] ERROR: Cannot find tool gcc
building on a gcc-less setup fails:

```bash
$ make
sh configure
*** ERROR: Cannot find tool gcc
Makefile:17: config.mk: No such file or directory
make: *** [Makefile:55: config.mk] Error 1
```

peppergrayxyz:
building on a system without gcc fails with error:

sh configure
*** ERROR: Cannot find tool gcc

use system default cc to enhance portability

fixes: #568

### DIFF
--- a/configure
+++ b/configure
@@ -58,7 +58,7 @@ check_toolchain()
     local bpftool_version
 
     : ${PKG_CONFIG:=pkg-config}
-    : ${CC=gcc}
+    : ${CC=cc}
     : ${OBJCOPY=objcopy}
     : ${CLANG=clang}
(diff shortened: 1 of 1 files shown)


tohojo (changes requested):
Please add the description of what the patch is fixing to the commit message, instead of just referring to the issue number. The git history should be stand-alone and not rely on references to external resources such as github to make sense.

peppergrayxyz:
done!

tohojo (approved):

PR #574, opened 2026-05-08
configure: fix syntax error in bpftool detection

1715173329:
`&>` is bashism and not recognized by posix sh, it will cause the test condition to be always false.


tohojo (approved):
LGTM - thank you for the fix! :)

PR #575, opened 2026-05-08
libxdp: Return last action from multiple programs

matijaGP:
When loading multiple programs which have actions set like this:

 struct {
	__uint(priority, 40);
	__uint(XDP_PASS, 0);
	__uint(XDP_REDIRECT, 1);
 } XDP_RUN_CONFIG(xdp_prog);

i.e. action other to XDP_PASS invokes next program, dispatcher will not return action from the last program, but XDP_PASS which is set at the end of the dispatcher.

Fix issue by making sure verdict from the last program is returned.

### DIFF
--- a/lib/libxdp/xdp-dispatcher.c.in
+++ b/lib/libxdp/xdp-dispatcher.c.in
@@ -59,7 +59,7 @@ SEC("xdp")
 int xdp_dispatcher(struct xdp_md *ctx)
 {
         __u8 num_progs_enabled = conf.num_progs_enabled;
-        int ret;
+        int ret = XDP_PASS;
 forloop(`i', `0', NUM_PROGS,
 `
         if (num_progs_enabled < incr(i))
@@ -75,7 +75,7 @@ forloop(`i', `0', NUM_PROGS,
                 goto out;
         ret = compat_test(ctx);
 out:
-        return XDP_PASS;
+        return ret;
 }
 
 SEC("xdp")


tohojo:
Hmm, I don't think this is a straight-forward bug fix. If an XDP program indicates that execution should continue for a given return code, it gives up control of what happens after it exits. So it's not clear that the default semantics should be to pass on the last return code.

Can you share more details of the program configuration you are running here, and how you expect things to work?

matijaGP:
Here is simplified scenario: I have AF_XDP socket and two xdp programs loaded on the interface. These two programs have such config:

struct {
__uint(priority, 40);
__uint(XDP_PASS, 0);
__uint(XDP_REDIRECT, 1);
} XDP_RUN_CONFIG(xdp_prog);

In both of them, I'm classifying traffic. If it matches condition, I need that frame injected to the network stack immediately (XDP_PASS action). If it doesnt match condition in 1st, repeat for 2nd program which works in the same way, it just has different conditions. At the end of each program I'm returning XDP_REDIRECT - meaning if conditions in both programs were not satisfied, I need that frame sent to AF_XDP socket. My expectation is that XDP_REDIRECT is returned, while I get XDP_PASS.

Proposed patch gives user control on the final verdict - it is value from user's program.

tohojo:
Right, so my concern is that this changes a long-standing behaviour, and we have no way of knowing what this will break.

From your example, it seems the behaviour you desire could be achieved by adding a third program with a run config like:
```c
struct {
__uint(priority, 50);
__uint(XDP_PASS, 1);
__uint(XDP_REDIRECT, 0);
} XDP_RUN_CONFIG(xdp_prog);
```

and have that program unconditionally `return XDP_REDIRECT;`? Any reason why that wouldn't work?

matijaGP:
Right, that would work, but these being hot paths, it is unnecessary performance penalty of having 3rd program loaded for the sake of making sure return value is last one.

I understand your concerns about changing long standing behavior, but I have couple of points where I believe this is right way to proceed and I'd kindly ask you to give it some thought.

Earlier you mentioned that we give up control to dispatcher when we are using callchains, but I wouldnt observe it so. My understanding is that it is interface to the dispatcher on what to do in case of different actions being returned. But we are really not specifying what would be return value with that, merely what will be invoked. Having that, returning XDP_PASS is decision of the dispatcher, and sort of arbitrary one. I'm sure we could find good arguments to return XDP_PASS (e.g. if nothing dropped frame, maybe network stack wants it), or XDP_DROP (e.g. if it passed all filters, lets defensively drop the frame so we dont risk machine overload), or something else, but in the end I think it should be up to the user to decide, and having there last action returned feels pretty much natural.

Do you think there is appropriate forum to have feedback from more potentially interested parties, like doing some sort of review via xdp-project itself?

tohojo:
Matija Glavinic Pecotic ***@***.***> writes:

> returning XDP_PASS is decision of the dispatcher, and sort of
> arbitrary one. I'm sure we could find good arguments to return
> XDP_PASS (e.g. if nothing dropped frame, maybe network stack wants
> it), or XDP_DROP (e.g. if it passed all filters, lets defensively drop
> the frame so we dont risk machine overload), or something else, but in
> the end I think it should be up to the user to decide, and having
> there last action returned feels pretty much natural.

Well, who is the user? You're assuming a model where the same user
installs all the programs. Whereas the underlying assumption of the
dispatcher design is that the individual XDP programs are essentially
being installed by different users, and the dispatcher serves only as
the coordination point. So in that model it's not obvious that "the user
decides" means "use the last return code as fall-through".

But in the end, you're right that this is essentially an arbitrary
decision. If we were starting from scratch, this would be much easier to
accommodate. But we're not, so we're kinda stuck with the path
dependency here :/

The issue is also that the dispatcher is not just an internal
implementation detail of libxdp, it's also a protocol that other
libraries can implement to interoperate - see:

https://github.com/xdp-project/xdp-tools/blob/main/lib/libxdp/protocol.org

So changing the behaviour at this point is not just about changing the
code, it's also about interoperability with other implementations. At
the very least it would require bumping the dispatcher version field to
avoid mixing behaviour.

If we are going to change this, I think a better model would be to make
the return code configurable independently of the return value of the
last program. I.e., for xdp-loader this would be something like
`xdp-loader load eth0 prog.o --fallthrough-action XDP_REDIRECT`
(possibly with an option for `--fallthrough-action LAST_RETURN` or
something to that effect). This can then be stored as part of the
dispatcher config, and reused for subsequent attachments, and we can
bail out if someone tries to specify a conflicting value. WDYT about
that model?

matijaGP:
> If we are going to change this, I think a better model would be to make the return code configurable independently of the return value of the last program. I.e., for xdp-loader this would be something like `xdp-loader load eth0 prog.o --fallthrough-action XDP_REDIRECT` (possibly with an option for `--fallthrough-action LAST_RETURN` or something to that effect). This can then be stored as part of the dispatcher config, and reused for subsequent attachments, and we can bail out if someone tries to specify a conflicting value. WDYT about that model?

Sounds reasonable, thanks. How would you like to proceed, would you like to take it further, or let me propose something? Either works for me.

tohojo:
Matija Glavinic Pecotic ***@***.***> writes:

> Sounds reasonable, thanks. How would you like to proceed, would you like to take it further, or let me propose something? Either works for me.

Please go ahead and take a stab at it :)

matijaGP:
It will take me a while before I will get into this, do you want to keep this PR open for that time? As far as I'm concerned, we can close it until I have something to review.

tohojo:
Sure, let's close it :)

PR #576, opened 2026-05-13
libxdp: Close refcnt_map_fd

matijaGP:
refcnt_map_fd leaks in scenario where AF_XDP socket is configured and tore down in loops. Reproduce by creating and closing AF_XDP socket within the same process and observing bpf map xsk_def_.data not to be cleaned.

Sample program is attached. To reproduce, start mini-sample with interface name, e.g.

./mini-sample eth0

Program will create and tear down AF_XDP socket. In separate terminal execute 

bpftool map

and observe that with each initialization of AF_XDP socket within the same process "xsk_def_.data" leaks.

[mini-sample.c](https://github.com/user-attachments/files/27704700/mini-sample.c)

### DIFF
--- a/lib/libxdp/xsk.c
+++ b/lib/libxdp/xsk.c
@@ -1336,6 +1336,10 @@ static void xsk_release_xdp_prog(struct xsk_socket *xsk)
 	if (value < 0)
 		pr_warn("Error occurred when decrementing xsk XDP prog refcount: %s, please detach program yourself\n",
 			strerror(-value));
+
+	close(ctx->refcnt_map_fd);
+	ctx->refcnt_map_fd = -ENOENT;
+
 	if (value)
 		goto out;
 
--- a/lib/libxdp/tests/Makefile
+++ b/lib/libxdp/tests/Makefile
@@ -1,6 +1,6 @@
 # SPDX-License-Identifier: (GPL-2.0 OR BSD-2-Clause)
 
-USER_TARGETS := test_xsk_refcnt test_xdp_frags test_link_detach test_xsk_umem_flags test_xdp_devbound
+USER_TARGETS := test_xsk_refcnt test_xdp_frags test_link_detach test_xsk_umem_flags test_xdp_devbound test_xsk_map_leak
 STATIC_LINK_TARGETS := check_kern_compat test_dispatcher_versions
 BPF_TARGETS := xdp_dispatcher_v1 xdp_dispatcher_v2 xdp_pass
 USER_LIBS := -lpthread
--- a/lib/libxdp/tests/test-libxdp.sh
+++ b/lib/libxdp/tests/test-libxdp.sh
@@ -1,6 +1,6 @@
 # SPDX-License-Identifier: (GPL-2.0 OR BSD-2-Clause)
 
-ALL_TESTS="test_link_so test_link_a test_old_dispatcher test_xdp_devbound test_xdp_frags test_xsk_prog_refcnt_bpffs test_xsk_prog_refcnt_legacy test_xsk_non_privileged test_link_detach test_xsk_umem_flags"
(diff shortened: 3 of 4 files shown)


lib/libxdp/xsk.c
```diff
@@ -1339,6 +1339,11 @@ static void xsk_release_xdp_prog(struct xsk_socket *xsk)
 	if (value)
 		goto out;
 
+	if (ctx->refcnt_map_fd >= 0) {
```
tohojo: The if statement seems redundant since the code already bails out above if the fd is negative?

tohojo (changes requested):
One comment; also, any chance you could turn that sample into a selftest and add that in a separate commit?

lib/libxdp/xsk.c
matijaGP: Right, I'll fix that.

matijaGP:
> One comment; also, any chance you could turn that sample into a selftest and add that in a separate commit?

Sure

matijaGP:
I've updated patch. When I tested for multiqueue sockets i.e. few sockets on the same interface, I noticed original logic to close fd only after ref goes down was wrong - it was still leaking in that case as ref is for program, not actual map for that socket, so closing refcnt_fd was moved up.

tohojo:
Hmm, tests seem to be a bit flaky - PTAL :)

matijaGP:
> Hmm, tests seem to be a bit flaky - PTAL :)

Sorry for that - I was missing mount of the bpfs, now it's added and this is fixed:

>      [test_xsk_map_leak]           START:
>           ./test_xsk_map_leak xsk_veth0
>           libxdp: No bpffs found at /sys/fs/bpf
>           libxdp: Can't use dispatcher without a working bpffs
>           libxdp: Falling back to loading single prog without dispatcher
>           libxdp: No bpffs found at /sys/fs/bpf
>           libxdp: No XDP dispatcher found on ifindex 23
>           Maps leaked, before 1, after 3
>           Map 2, iterator.rodata
>           Map 2, iterator.rodata
>           Map 247, xsks_map
>           Map 248, xsk_def_.data
>           Test test_xsk_map_leak exited with return code: 1
>      [test_xsk_map_leak]           FAIL

matijaGP:
Seems like it still fails and it is not problem in bpfs mount as I thought earlier, but some maps really do linger after test ends.

I will do some repeated tests on my side and try to get into this state.

matijaGP:
I left that on my machine overnight, and it passed 600 iterations so far, but no failure. Would it make sense that I try to run this test enviroment of yours? Could you please give me few pointers in that direction?

tohojo:
Well, the environment is just Github actions, and the definitions are in the repo. Unfortunately Github blocks automatic test runners on pull requests from people who don't already have commits in the repo, so pushing new commits here requires me to press the "run tests" button each time. But I think you should be able to run them in your own fork?

Anyway, what usually leads to these flaky tests is simply that the Github runners are incredibly slow. So it may well be that the code is actually correct, but the test trips up simply because the maps have not been cleared from the kernel yet by the time the test code polls them.

The kernel BPF selftests contain this helper:
```c
/*
 * Trigger synchronize_rcu() in kernel.
 */
int kern_sync_rcu(void)
{
	return syscall(__NR_membarrier, MEMBARRIER_CMD_SHARED, 0, 0);
}
```

which is used to avoid this problem without adding `usleep()` calls. I'd suggest trying that before doing the second `count_bpf_maps()` call.

matijaGP:
Thanks a lot, I managed to run tests on my fork, and got only one failure, test_capt_pcapng failed (after I added sync, before sync it looked similar to here).

Lets see how it will look now.

tohojo (approved):
Yeah, that's better - thanks!

PR #582, opened 2026-06-09
xdp-filter: note --keep-maps does not preserve the policy

Strykar:
`xdp-filter`'s persisted maps under `--keep-maps` don't record the
policy (`allow` or `deny`) they were created under. Reloading
`xdp-filter` afterwards under the opposite policy reuses the same
entries with the new policy's semantics, so an allowlist is silently
turned into a denylist and vice versa.

Document this in the `-k, --keep-maps` section of the man page and
point users at the unload-fresh path for switching policies safely.
The added paragraph uses "policy" (matching the existing `-p,
--policy` option) rather than "mode", to avoid collision with `-m,
--mode` (native/skb/hw).

Docs-only.

### DIFF
--- a/xdp-filter/README.org
+++ b/xdp-filter/README.org
@@ -109,6 +109,13 @@ However, this will also remove the contents of the maps (the filtering rules),
 so this option can be used to keep the maps around so the rules persist until
 =xdp-filter= is loaded again.
 
+Caution: the persisted maps don't record the policy they were created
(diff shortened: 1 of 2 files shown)


tohojo (changes requested):
Text looks good, but please also rebuild the man page and include that update into the same commit. Should be enough to run `make` after making the change to the readme file (just make sure you have Emacs available on the system, and picked up by the configure script).

tohojo (approved):

PR #583, opened 2026-07-09
xdp-trafficgen: Reduce accounting cost

kkdwvd:
When the tracepoint is active for xdp_devmap_xmit, we will constantly hammer bucket spin lock even when the entry already exists. This scales poorly with multiple threads generating traffic. Defer update to create the entry to first lookup failure, so that contention can be avoided.

Before this change, up to 30% of time on CPUs would be spent in the slow path, and after this the contention drops to 0. In a test scenario xmit/s went from 15 million to 20 million with 16 threads.


tohojo (approved):
Looks good, thanks!

PR #588, opened 2026-09-06
add make uninstall target removes files installed by xdp-tools

### LINKED ISSUE #397: How to uninstall xdp-tools after installing it
Does anyone know how to uninstall all installed binaries and libraries after running `sudo make install`? It seems there's no `uninstall` command from the Makefile. Any suggestion or experience is appreciated, thank you!

eraycep:
Add a make uninstall target that removes files installed by xdp-tools, including binaries, libraries, headers, BPF objects, man pages, scripts, and tests.

The target respects existing installation variables such as DESTDIR and leaves shared directories intact.

Fixes #397.


lib/libxdp/tests/Makefile
```diff
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: (GPL-2.0 OR BSD-2-Clause)
 
 USER_TARGETS := test_xsk_refcnt test_xdp_frags test_link_detach test_xsk_umem_flags test_xdp_devbound test_xsk_map_leak
+OPTIONAL_USER_TARGETS := test_xsk_non_privileged
```
tohojo: I don't think we should be doing this - `install` and `uninstall` should be targeting the same files, so if something is not built (and hence not installed) we should not attempt to remove it either.

This goes for all the places where this "include optional" pattern is included, of course :)

lib/libxdp/tests/Makefile
```diff
@@ -56,13 +57,16 @@ all: $(USER_TARGETS) $(STATIC_LINK_TARGETS) $(BPF_OBJS)
...
-.PHONY: install
+.PHONY: install uninstall
 install: all
 	install -m 0755 -d $(DESTDIR)$(SCRIPTSDIR)/tests/libxdp
 	install -m 0644 $(TEST_FILE) $(DESTDIR)$(SCRIPTSDIR)/tests/libxdp
 	install -m 0755 $(USER_TARGETS) $(STATIC_LINK_TARGETS) $(DESTDIR)$(SCRIPTSDIR)/tests/libxdp
 	install -m 0644 $(BPF_OBJS) $(DESTDIR)$(SCRIPTSDIR)/tests/libxdp
 
+uninstall:
+	rm -f $(addprefix $(DESTDIR)$(SCRIPTSDIR)/tests/libxdp/,$(notdir $(TEST_FILE) $(sort $(USER_TARGETS) $(OPTIONAL_USER_TARGETS)) $(STATIC_LINK_TARGETS) $(BPF_OBJS)))
```
tohojo: I don't think we need to bother with the `sort` :)

tohojo (changes requested):
Thank you for the contribution! A few comments below. Also, I did an install/uninstall sequence like:

```
$ make DESTDIR=$(pwd)/dest install
$ make DESTDIR=$(pwd)/dest uninstall
$ find dest
dest
dest/usr
dest/usr/local
dest/usr/local/include
dest/usr/local/include/xdp
dest/usr/local/lib
dest/usr/local/lib/pkgconfig
dest/usr/local/lib/bpf
dest/usr/local/share
dest/usr/local/share/man
dest/usr/local/share/man/man3
dest/usr/local/share/man/man8
dest/usr/local/share/xdp-tools
dest/usr/local/share/xdp-tools/tests
dest/usr/local/share/xdp-tools/tests/libxdp
dest/usr/local/share/xdp-tools/tests/xdp-filter
dest/usr/local/share/xdp-tools/tests/xdp-loader
dest/usr/local/share/xdp-tools/tests/xdpdump
dest/usr/local/share/xdp-tools/tests/xdp-bench
dest/usr/local/share/xdp-tools/tests/xdp-forward
dest/usr/local/share/xdp-tools/tests/xdp-monitor
dest/usr/local/share/xdp-tools/tests/xdp-trafficgen
dest/usr/local/sbin
```

I would expect the xdp-tools specific directories to also be removed (i.e., `dest/usr/local/share/xdp-tools/` and `dest/usr/local/include/xdp`). Probably some rules with `rmdir --ignore-fail-on-non-empty` should be able to achieve this (with a bit of care for ordering recursive directories correctly).

Finally, please add a proper commit message explaining the change. I'd expect at least mention of the choices mentioned here (and in the comments below). I.e., "we only remove the files we install", "we remove xdp-tools specific directories, but leave shared ones in place" :)

lib/libxdp/tests/Makefile
eraycep: Got it, removed. Thank you :)

lib/libxdp/tests/Makefile
eraycep: That's right, I did a bit much here. Reverted back to old pattern. Thank you :)

lib/libxdp/Makefile
```diff
@@ -62,6 +62,17 @@ install: all
...
+uninstall:
+	$(Q)$(MAKE) -C $(TEST_DIR) uninstall
+	$(Q)rm -f $(addprefix $(DESTDIR)$(HDRDIR)/,$(notdir $(LIB_HEADERS)))
+	$(Q)rm -f $(DESTDIR)$(LIBDIR)/pkgconfig/$(notdir $(PC_FILE))
+	$(Q)rm -f $(addprefix $(DESTDIR)$(LIBDIR)/,$(notdir $(SHARED_LIBS) $(STATIC_LIBS)))
+	$(Q)rm -f $(addprefix $(DESTDIR)$(BPF_OBJECT_DIR)/,$(notdir $(XDP_OBJS)))
+	$(if $(MAN_FILES),$(Q)rm -f $(addprefix $(DESTDIR)$(MANDIR)/man3/,$(notdir $(MAN_FILES))))
+	$(Q)rmdir --ignore-fail-on-non-empty $(DESTDIR)$(HDRDIR)
+	$(Q)rmdir --ignore-fail-on-non-empty $(DESTDIR)$(SCRIPTSDIR)/tests
+	$(Q)rmdir --ignore-fail-on-non-empty $(DESTDIR)$(SCRIPTSDIR)
```
tohojo: Let's try to keep the removal close to the install rules that create the directories; makes it easier to keep things in sync in the future.

These last two should probably go into both `lib/libxdp/tests/Makefile` and `common.mk`.

tohojo: Also, let's not prefix the removal rules with `$(Q)` - we're a bit inconsistent around this with install already, but since we don't have nice output for any of this, it's better to make the commands visible, I think.

lib/testing/Makefile
```diff
@@ -13,3 +13,7 @@ install_local::
 	install -m 0755 -d $(DESTDIR)$(SCRIPTSDIR)
 	install -m 0644 test_config.install.sh $(DESTDIR)$(SCRIPTSDIR)/test_config.sh
 	install -m 0644 $(XDP_OBJ) $(DESTDIR)$(SCRIPTSDIR)/
+
+uninstall_local::
+	rm -f $(DESTDIR)$(SCRIPTSDIR)/test_config.sh
+	rm -f $(addprefix $(DESTDIR)$(SCRIPTSDIR)/,$(notdir $(XDP_OBJ)))
```
tohojo, line 19: The install rule here is creating `$(DESTDIR)/$(SCRIPTSDIR)`, so let's remove it here as well.

tohojo (changes requested):
A few more nits. Also, even with the `--ignore-fail-on-non-empty` switch, `rmdir` can still fail if the directory doesn't exist. Which in particular means that running `make uninstall` twice will produce errors. We can get `make` to ignore these errors by prefixing the `rmdir` commands with a `-`, so let's do that throughout (only for `rmdir`, `rm -f` is fine as-is).

lib/libxdp/Makefile
eraycep: Got it, removed `$(Q)` prefix, also moved last two into `lib/libxdp/tests/Makefile` and `common.mk`. I also tested running uninstall twice and it ignores the errors.

lib/testing/Makefile
eraycep: Got it, added it.

tohojo (approved):
Alright, I ran out of things to complain about - nice work!

I'll let CI run before merging, but otherwise all good.

eraycep:
Thank you so much for the help! I hope to do more contributions :)

